PR #418, opened 2014-04-14
rhbz1077439 Refactor copyTrans to use lucene search for matching text flow target

alex-sl-eng:
https://bugzilla.redhat.com/show_bug.cgi?id=1077439


seanf:
This pull request (commit 46e9575fe9d142b22a25e892977c00b50ace0cf1) changed the sorting of all TM results, which seems to be the cause of https://zanata.atlassian.net/browse/ZNTA-721

@aeng In my fix (https://github.com/zanata/zanata-server/pull/1006) Hibernate Search will now sort by Lucene score, then by date (if the score is identical).  Will that work for copy trans?

PR #531, opened 2014-07-09
Tests for account activation and login

djansen-redhat:


djansen-redhat:
Closed, stale. Replaced by #705

PR #636, opened 2014-11-18
Warn on startup if EAP/WildFly too old

seanf:


seanf:
@carlosmunoz Do you have any other review comments?

carlosmunoz:
No more comments. :+1:

huangp:
Should we also check mysql version? I remember seeing people asking on freenode when they have newer mysql running and getting nasty exceptions but don't know why

seanf:
@huangp Yes, it would be good to check the mysql version (or at least log it), but I can't think of an easy way to do that in `ZanataInit`. Feel free to put in an RFE or a PR.

PR #642, opened 2014-12-01
rhbz1124630 - convert server REST tests after client jersey change

huangp:
REST Test in client is done through fake server. Server REST tests are RESTEasy dependent(proxy based plus RESTEasy ClientResponse) which makes it too hard to rewrite. Server REST tests are mainly for testing server implementation therefore copy over all the I*Resources and ZanataProxyFactory from old client and make it easier.


huangp:

seanf:
So, we add a massive amount of single-use test code to avoid some other massive change in test code?  From a maintenance point of view, this doesn't seem good on the face of it.  How can we cut it down?

huangp:
Two reasons for doing this (as I found out half way converting the tests):
1. Unlike client, server has to and already depend on RESTEasy 2. Making massive changes to use the jersey version of client we will hit nasty classpath problem (functional test has been converted but had problem in zanata-war REST tests, because test and prod code are running in the same JVM). That's a even worse maintenance problem to solve.
2. Some REST endpoints don't have client usage e.g. TMX. There is no client command for it. It was using RESTEasy proxy object and the proxy object was only used by this test. I think moving the interface here it's easier than creating a single-use client command here.

huangp:

seanf:
After discussion with @carlosmunoz it looks like the best option is to merge the (interface-based) "REST Tests" into the (HTTP-based) "raw REST tests", which should eliminate the need to use `org/zanata/rest/client/I*Resource` or anything from `zanata-rest-client` in Zanata server's test code.

huangp:

huangp:

huangp:
@seanf finally the build is green again. Do you mind reviewing it again?

seanf:
This still looks like an unreviewable mess, but I'll see what I can do.

zanata-war/pom.xml
```diff
@@ -163,6 +161,7 @@
         <groupId>com.ning.maven.plugins</groupId>
         <artifactId>maven-duplicate-finder-plugin</artifactId>
         <configuration>
+          <checkTestClasspath>false</checkTestClasspath>
```
seanf: Why?  Ignoring problems here could lead to trouble.

zanata-war/src/main/java/org/zanata/action/validator/DuplicateEmailValidator.java
```diff
@@ -34,15 +34,20 @@
...
+        if (string == null) {
             return true;
-        if (string.length() == 0)
+        }
+        if (string.length() == 0) {
             return true;
-        PersonDAO personDAO =
-                ServiceLocator.instance().getInstance(PersonDAO.class);
+        }
+        PersonDAO personDAO = getPersonDAO();
```
seanf: If you want to refactor little things like this, that's good, but they really need to be in separate commits for reviewability.

seanf: > If you want to refactor little things like this, that's good, but they really need to be in separate commits for reviewability.

Separate from the substantial changes I mean.  Feel free to put multiple minor refactorings in one commit.

zanata-war/src/main/java/org/zanata/notification/MessageUnwrapper.java
```diff
@@ -46,7 +46,8 @@ static MessageUnwrapper unwrap(final Message message) {
         ImmutableMap.Builder<String, ? super Object> builder =
                 ImmutableMap.builder();
         for (String propName : propNames) {
-            builder.put(propName, getPropertyValue(message, propName));
+            Object propertyValue = getPropertyValue(message, propName);
+            builder.put(propName, propertyValue == null ? "" : propertyValue);
```
seanf: Why is this changing in a test-related PR?

zanata-war/pom.xml
huangp: Because jersey and resteasy both have jaxrs-api. But one is in prod one is in test code.

zanata-war/src/main/java/org/zanata/action/validator/DuplicateEmailValidator.java
huangp: This is in the increase test coverage commit. Not that substantial.

zanata-war/src/main/java/org/zanata/notification/MessageUnwrapper.java
huangp: This is in the increase test coverage commit as well. After I added the unit test I discovered a bug and fixed it. I probably should separate this one out.

zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java
```diff
@@ -162,6 +162,8 @@ boolean transferFromTextFlows(List<TextFlow> from, HDocument to,
                         Response.status(Status.BAD_REQUEST)
                                 .entity("encountered TextFlow with duplicate ID "
                                         + tf.getId()).build();
+                log.error("encountered TextFlow with duplicate ID "
+                        + tf.getId());
```
seanf: If this is just telling us about an error in user input, it should be WARN at most.

zanata-war/src/main/java/org/zanata/rest/service/SourceDocResourceService.java
```diff
@@ -421,7 +421,7 @@ private HLocale validateTargetLocale(LocaleId locale, String projectSlug,
         } catch (ZanataServiceException e) {
             log.warn(
                     "Exception validating target locale {} in proj {} iter {}",
-                    e, locale, projectSlug, iterationSlug);
+                    locale, projectSlug, iterationSlug);
```
seanf: Whether we should such an exception stacktrace could be debated, but I don't think it should change in this pull request.

zanata-war/pom.xml
seanf: The danger of tests not testing what we think is too high a price to pay for a (possibly) smaller download on the client.

zanata-war/src/main/java/org/zanata/action/validator/DuplicateEmailValidator.java
seanf: I know it's not substantial; that's why it should be separated from the substantial changes.

zanata-war/src/main/java/org/zanata/rest/service/SourceDocResourceService.java
huangp: This is simply wrong. You can't mix template with exception stacktrace. This was treating the exception as first variable and log its toString as locale.

seanf: Actually, this can work, to fix it we just need to move the exception to
the end of the varargs list, and slf4j will include the stack trace as
intended.

Or, if we just want the exception's `toString()`, we can add another `{}`
to the template.  But as it's only WARN, I would be inclined to keep the
stack trace.

zanata-war/pom.xml
seanf: Do we have two copies of API interfaces (in which case, we should fix it by blocking one of the transitive dependencies), or do we actually have two different versions of implementation code?  If it's the latter, perhaps you could move something out of zanata-war to make it work.

In any case, `<checkTestClasspath>false</checkTestClasspath>` is not something we should do. :-1:

zanata-war/src/main/java/org/zanata/rest/service/SourceDocResourceService.java
huangp: No you can not have exception mixed with varargs. There is no such method. It's either varargs (which is used as variables to fill in template) or simple message plus exception. I guess the reason is varargs has to be the last parameter of a method.

seanf: Try it.  It works.  http://slf4j.org/faq.html#paramException

huangp: It does work. Although IntelliJ still gives warning on using it this way. 

----- Original Message -----

seanf: > http://slf4j.org/faq.html#paramException
> If the exception is not the last argument, it will be treated as a plain object and its stack trace will NOT be printed. **However, such situations should not occur in practice.**

Ha! Obviously that wasn't true in this case.

seanf:
- [ ] The exception still seems to be missing from the log message in validateTargetLocale.

seanf:
@huangp Do the raw rest tests cover JSON encoding much?  (See https://bugzilla.redhat.com/show_bug.cgi?id=1205473)

huangp:
Looks like only compatibility tests has json tests. The converted raw tests (previously REST test but using zanata-rest-client) are not doing json encoding.

huangp:
It's possible and easier to add JSON test now that we switched to raw tests.

seanf:
I've reviewed this as best I can, but not as thoroughly as I would like because of the mixture of different changes.  Please make sure you run all the functional tests.

seanf:
Looks good.  I've kicked off all functional tests.

huangp:
squash merged in commit 4c49354

PR #647, opened 2014-12-05
Fix formatting

seanf:


seanf:
Squash-merged as https://github.com/zanata/zanata-server/commit/f75a66b6146d06a46728bc234db8675c0e6a75b8 .

PR #651, opened 2014-12-15
Add locale aliases to project and version locales.

davidmason:
- change UI for locale settings in project
 - change UI for locale settings in version
 - add locale aliases to project and version
 - add REST endpoint */locales under standard project and version endpoints

Note: this will fail the PR build until we merge https://github.com/zanata/zanata-api/pull/25 (merged++)


zanata-war/src/main/java/org/zanata/action/HasLanguageSettings.java
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Enable any disabled locale that is selected.
+     *
+     * A disabled locale is selected if its localeId maps to true in
+     * selectedDisabledLocales.
+     */
+    void enableSelectedLocales();
+
+    void enableLocale(HLocale locale);
+}
```
aeng: No new line.

zanata-war/src/main/webapp/resources/zanata/ajax-command-link.xhtml
```diff
@@ -0,0 +1,45 @@
...
+    </span>
+    <span class="js-in-progress is-hidden">
+      <composite:renderFacet name="loading" />
+    </span>
+    <span class="js-default">
+      <composite:insertChildren/>
+    </span>
+  </a4j:commandLink>
+</composite:implementation>
+</html>
```
aeng: Isn't this similar to loader-button-label we had in assets?

zanata-war/src/main/webapp/resources/zanata/ajax-command-button.xhtml
```diff
@@ -0,0 +1,50 @@
...
+    </span>
+    <span class="js-in-progress is-hidden">
+      <composite:renderFacet name="loading" />
+    </span>
+    <span class="js-default">
+      <composite:insertChildren/>
+    </span>
+  </zui:commandButton>
+</composite:implementation>
+</html>
```
aeng: Isn't this similar to loader-button-label we had in assets? Maybe combining both?

zanata-war/src/main/webapp/WEB-INF/layout/version/settings-tab-languages.xhtml
```diff
@@ -0,0 +1,753 @@
...
+  }
+
+  function localeAliasInput(localeId) {
+    return localeAliasForm(localeId).find('input:text');
+  }
+
+  function onLocaleAliasInputKeyDown(localeId, event) {
+
+    // Enter
+    if (event.keyCode === 13) {
```
aeng: Can use isEnterKey() in components-script.js

zanata-war/src/main/webapp/WEB-INF/layout/version/settings-tab-languages.xhtml
```diff
@@ -0,0 +1,753 @@
...
+      event.preventDefault();
+      localeAliasForm(localeId).find('input:submit').click();
+      return false;
+    }
+    // Esc
+    if (event.keyCode === 27) {
```
aeng: Maybe should extract this to components-script.js (same as isEnterKey)?

zanata-war/src/main/webapp/WEB-INF/layout/version/settings-tab-languages.xhtml
```diff
@@ -0,0 +1,753 @@
...
+      return isSupported;
+    }
+  })();
+
+  /**
+   * Attach to keypress event to prevent submission of form when enter is
+   * pressed in a textbox.
+   */
+  function doNotSubmit(event) {
+    if (event.keyCode === 13) {
```
aeng: See comment above on isEnterKey()

zanata-war/src/main/java/org/zanata/rest/service/ProjectIterationLocalesService.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+import java.util.Map;
+
+@Name("projectIterationLocalesService")
+@Path(ProjectIterationLocalesService.SERVICE_PATH)
+public class ProjectIterationLocalesService extends LocalesService implements ProjectIterationLocalesResource {
+    @PathParam("projectSlug")
+    String projectSlug;
+
+    @PathParam("iterationSlug")
```
aeng: I'm using `versionSlug` for all the new endpoint. Personal preferences.

zanata-war/src/main/java/org/zanata/ui/AbstractTextSearch.java
```diff
@@ -0,0 +1,57 @@
...
+    private String query = "";
+
+    private String searchedQuery = "";
+
+    @Getter
+    private List<E> results = new ArrayList<>();
+
+    public synchronized void search() {
+        // only search if it's necessary
+        if( !query.equals(searchedQuery) ) {
```
aeng: spacing after `if` before `(`. And don't think spacing is needed before and  after `(` `)`

zanata-war/src/main/java/org/zanata/ui/input/TextInputRenderer.java
```diff
@@ -0,0 +1,85 @@
...
+
+            @Override
+            public void startElement(String name, UIComponent component)
+                    throws IOException {
+                super.startElement(name, component);
+                if("input".equals(name)) {
+                    // Add the extra attributes
+                    for(ExtraAttributes att : ExtraAttributes.values()) {
+                        Object value = component.getAttributes().get(att.name());
+                        if( value != null ) {
```
aeng: Again spacing after `if`, and no spacing in between `(` and `)`

zanata-war/src/main/resources/db/changelogs/db.changelog-3.6.xml
```diff
@@ -30,4 +30,53 @@
       referencedTableName = "HProject" referencedColumnNames = "id"/>
   </changeSet>
 
+  <changeSet id="1" author="damason@redhat.com">
+    <comment>Create table HProject_LocaleAlias.</comment>
```
aeng: Do we still want to keep the naming prefix with `H`? All the new tables has no prefix in it now.

zanata-war/src/main/resources/db/changelogs/db.changelog-3.6.xml
```diff
@@ -30,4 +30,53 @@
...
+      baseColumnNames = "projectId"
+      constraintName = "FK_HProject_LocaleAlias_HProject"
+      referencedTableName = "HProject"
+      referencedColumnNames = "id"/>
+  </changeSet>
+
+  <changeSet id="2" author="damason@redhat.com">
+    <comment>Create table HProjectIteration_LocaleAlias.</comment>
+
+    <createTable tableName="HProjectIteration_LocaleAlias">
```
aeng: Same for this table naming. See comment above.

zanata-war/src/main/webapp/WEB-INF/layout/project/settings-tab-languages.xhtml
```diff
@@ -0,0 +1,752 @@
...
+      return isSupported;
+    }
+  })();
+
+  /**
+   * Attach to keypress event to prevent submission of form when enter is
+   * pressed in a textbox.
+   */
+  function doNotSubmit(event) {
+    if (event.keyCode === 13) {
```
aeng: See isEnterKey() in components-script.js

zanata-war/src/main/webapp/WEB-INF/layout/project/settings-tab.xhtml
```diff
@@ -22,7 +23,7 @@
   }
 
   function onPayloadInputKeyDown(input, event) {
-      if(event.keyCode == 13) {
+      if(event.keyCode === 13) {
```
aeng: See isEnterKey() in components-script.js

zanata-war/src/main/webapp/resources/zanata/ajax-command-link.xhtml
davidmason: @aeng does loader-label wait for a response from an ajax request?

zanata-war/src/main/java/org/zanata/rest/service/ProjectIterationLocalesService.java
davidmason: I agree that we should change from `"iteration"` to `"version"` everywhere, but we should really do it all at once - doing it piecemeal will just cause confusion.

In this case it is a sub-endpoint of the iteration resource, so it uses the path from that. If we change the path in the API package to use `"versionSlug"` in the path template, then we can change this here. I think we should do that when we define the alternative endpoint for that resource, which we can do by using a more complex path template (so that all the code that uses it will just work, except for the change of the slug name).

I think such changes are out of scope for this pull request.

zanata-war/src/main/resources/db/changelogs/db.changelog-3.6.xml
davidmason: It is mapping from class `HProject` to class `LocaleAlias`, so the `H` is just part of the class name that is represented by the id in the first column.

zanata-war/src/main/resources/db/changelogs/db.changelog-3.6.xml
davidmason: Same response. See response above :arrow_up:

davidmason:

davidmason:

zanata-war/src/main/webapp/resources/zanata/ajax-command-link.xhtml
davidmason: @carlosmunoz this is your addition, what are your thoughts on using loader__label in this?

zanata-war/src/main/java/org/zanata/action/VersionHome.java
```diff
@@ -404,35 +400,31 @@ public void copyValidationFromProject() {
...
-            this.conversationScopeMessages = conversationScopeMessages;
-        }
-        return update();
+    @Transactional
+    public String updateFromSuper() {
+        this.joinTransaction();
+        this.getEntityManager().flush();
+        this.raiseAfterTransactionSuccessEvent();
+        return "updated";
     }
```
davidmason: @carlosmunoz does this look reasonable to you? The code is copied directly from `super.update()` but leaves out the line that adds the "successfully updated" message.

zanata-war/src/main/java/org/zanata/service/impl/ConfigurationServiceImpl.java
```diff
@@ -105,7 +105,6 @@ public String getConfig() {
                             .append("\n")
                             .append(makeProjectTypeSection(getProjectType()))
                             .append("\n");
-            appendLocalesIfPresent();
```
davidmason: @seanf all generated zanata.xml files will no longer have a `<locales>` section. That includes the zanata.xml in zip downloads. It should not change anything about po files.

seanf: But what about the names of the PO files (and the locale directories they
live in, inside zips)?  Will they still be consistent with the behaviour of
the cli client?  Will the user still be able to download an offline
translation zip, extract it, edit some po files and run zanata-cli push,
and have it all work?

davidmason: I think that is a piece of scope that we missed.

zanata-war/src/main/java/org/zanata/action/ProjectHome.java
```diff
@@ -122,11 +134,82 @@
...
+    private Map<LocaleId, String> enteredLocaleAliases = Maps.newHashMap();
+
+    @Getter
+    @Setter
+    private Map<LocaleId, Boolean> selectedEnabledLocales = Maps.newHashMap();
+
+    // Not sure if this is necessary, seems to work ok on selected disabled
+    // locales without this.
+    public Map<LocaleId, Boolean> getSelectedEnabledLocales() {
+        if (selectedEnabledLocales == null) {
```
aeng: selectedEnabledLocales is already instantiate on create.

zanata-war/src/main/java/org/zanata/action/VersionHome.java
```diff
@@ -521,43 +502,373 @@ private void ensureMutualExclusivity(
...
-        protected Collection<HLocale> getLocales() {
-            if (StringUtils.isNotEmpty(projectSlug)
-                    && StringUtils.isNotEmpty(slug)) {
-                return localeServiceImpl
-                        .getSupportedLanguageByProjectIteration(projectSlug,
-                                slug);
+    @Restrict("#{s:hasPermission(versionHome.instance, 'update')}")
+    public void removeAllLocaleAliases() {
+        List<LocaleId> removed = new ArrayList<>();
+        List<LocaleId> aliasedLocales =
```
aeng: I think Lists.newArrayList(getLocaleAliases().keySet()) is neater.

seanf: Yes, slightly, but only with a static import.

zanata-war/src/main/java/org/zanata/action/ProjectHome.java
seanf: Might be worth using this: http://projectlombok.org/features/GetterLazy.html

davidmason: @aeng I actually think I need to use the opposite approach - I should set `selectedEnabledLocales` to `null` when the list of enabled locales changes, otherwise it will not contain the correct set of locales. I will sort that out.

davidmason: I have already modified this on my cleanup branch, but I will put the branch through QA since it is too late to merge the cleanup branch, and this check is not destructive.

davidmason: @seanf in this case the field has to be reinitialized any time the list of enabled locales changes, so `GetterLazy` is not a good fit. I am glad I know about `GetterLazy` now though, looks very handy.

djansen-redhat:

djansen-redhat:
:small_red_triangle: Language entries receive click even if just clicking the menu, Cancel
:small_red_triangle_down: Language lists are not sorted
:small_red_triangle_down: Language alias maxlength 255?

davidmason:
@lukebrooker was there something in the server code to do about the click events being passed through, or is that to be fixed in assets?

davidmason:
@djansen-redhat 

> Language lists are not sorted
They seemed to have a consistent order whenever I have messed with them, but maybe sorting by locale code rather than display name (which in itself may be considered a bug - ideally it would be sortable with the most commonly needed as default).

> Language alias maxlength 255?
I matched the maxlength to that of locale codes. Other than being consistent in that way, I have no reason for that particular limit.

davidmason:
> The same alias can be mapped to several languages, but only the first will work. Are these supposed to be unique?

There was no mention of uniqueness in the specification, and nothing in the implementation checks for uniqueness (or mutual exclusivity with locale ID). Enforcing uniqueness would disallow some operations that may be useful, but we would need some careful thought to decide which approach is best for our system.

zanata-war/src/main/java/org/zanata/rest/service/ProjectIterationLocalesService.java
djansen-redhat: [RFE: Rename iterationSlug to versionSlug everywhere](https://bugzilla.redhat.com/show_bug.cgi?id=1198379)

davidmason:
Squash-merged in https://github.com/zanata/zanata-server/commit/bcb2e36ef8cd3847b3f54836addcd07ec28d7d75

PR #658, opened 2014-12-22
Use Graphene and Drone (Arquillian) for functional tests

carlosmunoz:
Replace our own Suite extension with a more evolved one from Maven Central.
Create a very basic test that showcases Graphene and Drone.


carlosmunoz:
Closing this as it's been stale for too long.

PR #661, opened 2015-01-14
Fix admin and related tests after UI update

djansen-redhat:


functional-test/src/main/java/org/zanata/page/administration/ManageLanguagePage.java
```diff
@@ -40,14 +41,24 @@
     public static final int LOCALE_COLUMN = 0;
     public static final int ENABLED_COLUMN = 3;
 
-    private By languageTable = By.id("languageForm:threads");
+    private By languageTable = By.id("languageForm");
     private By addLanguageButton = By.linkText("Add New Language");
+    private By moreActions = By.id("more-actions");
+    private By enableByDefault = By.linkText("Enable by default");
+    private By disableByDefault = By.linkText("Disable by default");
```
carlosmunoz: Are these a good idea? I mean, the moment we change the message these will start failing. Wouldn't an id be better?
(I know these are all over the place, but at least we can stop adding more)

carlosmunoz:
Review finished.

functional-test/src/main/java/org/zanata/page/administration/ManageLanguagePage.java
djansen-redhat: Ideally, yes - but the issue is that these are a repeated item (for every language in the list) and giving them an id would result in duplicate ids - unless of course we're very clever about it and assign them contextually accurate ids (e.g. id = "enable" + $language)

carlosmunoz: Got it. :+1:

PR #662, opened 2015-01-15
check database compatibility at start up

huangp:
If running zanata against a whole new database schema which will execute all liquibase change logs, we are known to only compatible with mysql up to 5.5. Not quite sure about mariaDB as I am on RHEL and I don't have access to it.

This change will check for database and its version. If it's incompatible, it will log the message and throw exception. It will be easier for people to figure out what's wrong with their setup.

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/662)

### DIFF
--- a/zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java
+++ b/zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java
@@ -266,10 +266,10 @@ public class ApplicationConfiguration implements Serializable {
     public List<String> getAdminEmail() {
         String s = databaseBackedConfig.getAdminEmailAddress();
         if (s == null || s.trim().length() == 0) {
-            return new ArrayList<String>();
+            return new ArrayList<>();
         }
         String[] ss = s.trim().split("\\s*,\\s*");
-        return new ArrayList<String>(Arrays.asList(ss));
+        return new ArrayList<>(Arrays.asList(ss));
     }
 
     public String getFromEmailAddr() {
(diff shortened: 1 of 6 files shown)


djansen-redhat:
> 11:22:50 Code coverage enforcement failed for the following metrics:
> 11:22:50     Methods's stability is 44.93 and set mininum stability is 45.0

djansen-redhat:
A preliminary test of this shows the error, but the server continues to load up and start. Is this *expected*?  A user might miss the error and believe the server has started correctly but doesn't behave correctly.

huangp:
The JBoss server will always continue and start. But what about zanata itself? I thought an exception will stop it from loading up. If not I have to find some other way.

I will fix the test coverage thingy after I finish converting tests. I am afraid it will drop further since they will become integration tests.

seanf:
@huangp you might want to use the new annotation `@CoverageIgnore` on the methods which can't be tested via unit tests.  Just don't overdo it!

seanf:
This seems like an *incredible* amount code to write and maintain for a simple version check.  Let's see if we can avoid dependending on the liquibase datasource, JNDI and all that.  There should be an easier and safer way to inject a datasource and ask for its `DataBaseMetaData`.  (Perhaps a small factory bean could inject the datasource, and outject/produce a DatabaseVersionInfo.)  Even the version check itself seems to be using three different classes.

Also, I don't think it's a good idea to completely prevent deployment on newer databases.  Who's to say MariaDB 11.0 won't work, until we try it?  And Zanata does work on MySQL 5.6 now.

I think we should (a) have a simple warning only, or (b) throw an exception, but with a helpful error message: one which mentions the system property to override the check (eg ignore.db.version).

Perhaps we should put this on hold until after the CDI migration.  It might help to reduce the amount of code required.

PR #666, opened 2015-01-19
Fix intermittent smtp tests

seanf:


seanf:
@huangp Could you please look at this when you get a chance?  We want to tighten up the functional tests before making any big changes for CDI (and before releasing 3.6).

huangp:
:+1:

PR #667, opened 2015-01-19
Implement plural support for zanata-spa

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1172437


aeng:
Reviewed in Zanata-SPA https://github.com/zanata/zanata-spa/pull/89

djansen-redhat:
:x: The Esc keyboard shortcut doesn't seem to work
:x: Saving seems to be flaky, sometimes the spinner doesn't stop
:question: Is it expected that, when translating the first target in a set of plurals, pressing copy on the *second* source copies it to the first target?
:x: Pressing copy seems to add a newline to the target after the text

aeng:
@djansen-redhat :question: Is it expected that, when translating the first target in a set of plurals, pressing copy on the second source copies it to the first target? 

Yes. Copy source will copy to the focused text area. But if you're using key shortcut for copy source, it will try to copy the corresponding source[focused target index] to focused target. If there's more target than source, it will copy the last source to the focused target.

aeng:
@djansen-redhat :question: Pressing copy seems to add a newline to the target after the text

Confirmed there's new line character from source. Zanata-spa doesn't show it, check with our editor.

aeng:
Transfered to new PR for release branch: https://github.com/zanata/zanata-server/pull/674

PR #668, opened 2015-01-20
Fix test element focus issues

djansen-redhat:
In some cases, element.blur works more reliably than a page click

### DIFF
--- a/functional-test/src/main/java/org/zanata/page/CorePage.java
+++ b/functional-test/src/main/java/org/zanata/page/CorePage.java
@@ -193,6 +193,15 @@ public class CorePage extends AbstractPage {
         }
     }
 
+    /**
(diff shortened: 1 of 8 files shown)


seanf:
@djansen-redhat What user action is `defocus()` supposed to simulate?

djansen-redhat:
@seanf It's supposed to be a click on the background - sending a "tab" press to WebDriver seems unreliable - but it appears that interacting with elements from a WebDriver is not *exactly* the same as what a user would do.
e.g. sendKeys on the field differs slightly to a user clicking and entering text, causing a lack of "direct" focus.
blur() is effectively the standard browser "unhook" process from a focused field.

seanf:
:+1:

PR #672, opened 2015-01-27
Replace CSS file with JS for loading Google Web Fonts

lukebrooker:
Uses the JS service instead of CSS file so it doesn't block rendering.

Should fix: [Bug: WebUI is very slow if users cannot access Google](https://bugzilla.redhat.com/show_bug.cgi?id=1186084)


djansen-redhat:
Wrong target branch, bro (integration/master)

PR #675, opened 2015-01-27
Fix input field not save when using same text as placeholder: https://bu...

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1185134

### DIFF
--- a/zanata-war/src/main/java/org/zanata/action/ServerConfigurationBean.java
+++ b/zanata-war/src/main/java/org/zanata/action/ServerConfigurationBean.java
@@ -248,13 +248,12 @@ public class ServerConfigurationBean implements Serializable {
     }
 
     private void persistPropertyToDatabase(PropertyWithKey<String> property) {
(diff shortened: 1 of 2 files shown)


huangp:
:+1:

djansen-redhat:
:white_check_mark: Tested.  Build has also passed but not reflected here correctly.

aeng:
Merged into release. https://github.com/zanata/zanata-server/commit/4be04f5e536047adfa7f5434b0916804b7bcf92d

davidmason:
@aeng it looks like the only code changes are layout changes and a variable name. Does this really do what the title says?

djansen-redhat:
@aeng Not *exactly*. The major issue is using the actual Zanata ToU url in a placeholder will cause our current system to break.
In future, we should fix the field so that it can tell if the text entered is placeholder or user entered.

davidmason:
There is a fix on the branch I am working in that will separate placeholder text into an attribute properly, so we should be able to fix the field once that is merged.

PR #682, opened 2015-02-04
Improve unit test coverage

seanf:


zanata-war/src/main/java/org/zanata/webtrans/client/util/JavascriptUtil.java
```diff
@@ -33,6 +35,7 @@
      * @param varName Variable name.
      * @return The value (as a string) assigned to varName.
      */
+    @CoverageIgnore
```
carlosmunoz: No reason here?

zanata-war/src/main/java/org/zanata/webtrans/client/util/DateUtil.java
```diff
@@ -53,8 +44,7 @@ public static String formatTime(Date date) {
      */
     public static String formatLongDateTime(Date date) {
         if (date != null) {
-            return DateTimeFormat.getFormat(DATE_TIME_LONG_PATTERN)
-                    .format(date);
+            return dtfLong.format(date);
```
carlosmunoz: Why did we change it to "not thread safe"?
From what I can see, it was creating its own formatter before which made it thread safe.

zanata-war/src/test/java/org/zanata/action/validator/EmailListValidatorTest.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Test
+    public void DIsValid() throws Exception {
+        assertTrue(validator.isValid("d@example", context));
+    }
+
+    @Test
+    public void aBCDIsValid() throws Exception {
+        assertFalse(validator
```
carlosmunoz: I thought we were sticking to Assertj for assertions from now on.

zanata-war/src/test/java/org/zanata/webtrans/client/util/DateUtilTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @BeforeMethod
+    public void before() {
+        aDate = new GregorianCalendar(2013, 11, 25, 1, 2, 3).getTime();
+    }
+
+    @Test
+    public void testFormatShortDate() throws Exception {
+        String s = DateUtil.formatShortDate(aDate);
+        assertEquals(s, "25/12/13 01:02");
```
carlosmunoz: Same comment about assertions here.

zanata-war/src/main/java/org/zanata/webtrans/client/util/JavascriptUtil.java
seanf: That was the first one I put in, before I thought of adding a reason to the annotation.  I'll put it in.

zanata-war/src/main/java/org/zanata/webtrans/client/util/DateUtil.java
seanf: The old version couldn't be tested without using GWTTestCase, which we won't work with TestNG.  GWT code all runs in a single Javascript thread anyway.

zanata-war/src/test/java/org/zanata/action/validator/EmailListValidatorTest.java
seanf: AssertJ > Hamcrest, but if all you need is a simple boolean check, I think it's overkill.

zanata-war/src/test/java/org/zanata/webtrans/client/util/DateUtilTest.java
seanf: Again, it's just equality.  I don't think AssertJ would make this more readable, or produce better failure messages.

carlosmunoz:
:+1: Reviewed

PR #687, opened 2015-02-11
disable rendering in filter list if empty string

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1190557


huangp:
:+1:

djansen-redhat:
- [ ] Spaces actually affect the behaviour (i.e. will filter for documents with a space) so using trim() breaks the Cancel if the user entered a space.  I suggest Cancel clears the field if there is anything in there.

djansen-redhat:
:white_check_mark: Tested

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/51fd4b3e3c756b831026f9e01682edebd9fd40b1

PR #690, opened 2015-02-12
Fix inactive account redirection with openid account

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1188457


aeng:
See https://github.com/zanata/zanata-server/pull/691

PR #692, opened 2015-02-12
Implement customise language display name

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1169662

### DIFF
--- a/zanata-war/src/main/resources/db/changelogs/db.changelog-3.7.xml
+++ b/zanata-war/src/main/resources/db/changelogs/db.changelog-3.7.xml
@@ -13,4 +13,18 @@
     </addColumn>
   </changeSet>
 
+  <changeSet id="2" author="aeng@redhat.com">
+    <comment>Add displayName, nativeName column to HLocale</comment>
+    <addColumn tableName="HLocale">
+      <column name="displayName" type="varchar(100)">
+        <constraints nullable="true" />
+      </column>
+    </addColumn>
+    <addColumn tableName="HLocale">
+      <column name="nativeName" type="varchar(100)">
+        <constraints nullable="true" />
+      </column>
+    </addColumn>
+  </changeSet>
+
 </databaseChangeLog>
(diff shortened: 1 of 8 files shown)


huangp:
:+1:

aeng:

aeng:

zanata-war/src/main/resources/db/changelogs/db.changelog-3.7.xml
```diff
@@ -13,4 +13,18 @@
     </addColumn>
   </changeSet>
 
+  <changeSet id="2" author="aeng@redhat.com">
+    <comment>Add displayName, nativeName column to HLocale</comment>
+    <addColumn tableName="HLocale">
+      <column name="displayName" type="varchar(100)">
+        <constraints nullable="true" />
+      </column>H
```
djansen-redhat: ?

djansen-redhat:
Do we *need* a tooltip, or something, to indicate the clearing the fields reverts to the default value?

aeng:
Placeholder text indicated the default value if field leave empty. Its not showed on the first load due to seam issue.

djansen-redhat:
It appears that when the name has been changed, then removed, the item shows up without a label on the Languages list page.

djansen-redhat:
- [ ] Changing the name, but not the native name, sometimes causes an exception

Also, do you think this change effectively invalidates [1035068 Improve the Admin Add language feature](https://bugzilla.redhat.com/show_bug.cgi?id=1035068)?  I'd go with "no" as the fields are not available at the time of create, and the behaviour is subtly different - but being able to customise the display name at all gives the user an option.

djansen-redhat:
:white_check_mark: Tested

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/81054b713c3d9b1e1962915188c8d40c6513c0b5

PR #694, opened 2015-02-16
Rhbz1183994 user contribution matrix

huangp:
https://bugzilla.redhat.com/show_bug.cgi?id=1183994


zanata-war/src/main/java/org/zanata/action/ProfileHome.java
```diff
@@ -0,0 +1,138 @@
+/*
+ * Copyright 2010, Red Hat, Inc. and individual contributors
```
davidmason: 2010...

djansen-redhat: ^ This comment wins the Github.

huangp:
@carlosmunoz do you want to review this one? Just the server side implementation. The javascript part is reviewed by Luke.

carlosmunoz:
Sure. Build failed though...

zanata-war/src/main/java/org/zanata/dao/DatabaseSpecific.java
```diff
@@ -0,0 +1,18 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Marker annotation to mark methods that will only work in certain database.
+ */
+@Target({ElementType.METHOD, ElementType.LOCAL_VARIABLE, ElementType.TYPE, ElementType.PARAMETER})
+@Retention(RetentionPolicy.SOURCE)
+public @interface DatabaseSpecific {
```
carlosmunoz: Do we need an extra annotation? Or could we make it part of the `@NativeQuery` one?

zanata-war/src/main/java/org/zanata/rest/dto/matrix/DetailMatrix.java
```diff
@@ -0,0 +1,20 @@
+package org.zanata.rest.dto.matrix;
+
+import org.zanata.common.ContentState;
+import org.zanata.common.LocaleId;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+/**
+ * @author Patrick Huang
```
carlosmunoz: Some comments explaining this class would be appreciated as its name is a bit generic.

zanata-war/src/main/java/org/zanata/dao/TextFlowTargetHistoryDAO.java
```diff
@@ -159,4 +174,136 @@ public boolean findConflictInHistory(HTextFlowTarget target,
...
+            prefix = "-";
+            standardOffset = -standardOffset;
+        }
+        return String.format("%s%02d:00", prefix,
+                TimeUnit.MILLISECONDS.toHours(standardOffset));
+    }
+
+    @Getter
+    @RequiredArgsConstructor
+    public static class UserTranslationMatrix {
```
carlosmunoz: If this class is going to be used as a DTO, I'd rather not have it be an inner class.

zanata-war/src/main/java/org/zanata/dao/DatabaseSpecific.java
huangp: This is mainly to indicate the query uses some database specific functions. NativeQuery alone can still be portable between different databases.

zanata-war/src/main/java/org/zanata/dao/TextFlowTargetHistoryDAO.java
huangp: oops. I forgot to push the last commit. It's gone now.

zanata-war/src/main/java/org/zanata/rest/service/DateRange.java
```diff
@@ -0,0 +1,66 @@
...
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+/**
+* @author Patrick Huang
+*         <a href="mailto:pahuang@redhat.com">pahuang@redhat.com</a>
+*/
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+class DateRange {
```
carlosmunoz: Nice :+1: 
I wonder if there's a way to make Resteasy do the mapping automatically for us.
Maybe in V 3.0

zanata-war/src/main/webapp/WEB-INF/urlrewrite.xml
```diff
@@ -196,6 +196,18 @@
     <to last="true">/profile/merge_account.seam$1</to>
   </rule>
 
+  <rule>
+    <from casesensitive="true">^/profile/view/([^/]+)(/([^?]*)\??(.*))?$</from>
```
carlosmunoz: How is this working? Are we allowing sub-paths for this view?

huangp: it's like project page. /profile/view/someusername will go to someone's profile page with username 'someusername'.

zanata-war/src/main/java/org/zanata/dao/DatabaseSpecific.java
carlosmunoz: That's what I mean, maybe we can just add an extra param to that annotation. Something like:
```java
@NativeQuery( description="This is a native", specificTo="MySQL" )
```
It's just a marker annotation with some info, so no need to create more classes for it IMO.

zanata-war/src/main/webapp/WEB-INF/urlrewrite.xml
carlosmunoz: Ah ok... it's just that the project page in particular has a more complicated regexp because it basically allows any urls of the form `/project/view/myproject/../..` where you can add as many parameters as you want. These parameters will then get parsed by javascript to show specific portions of the UI.
If we don't need this functionality in this page, we can use a simpler regex. That's all I was suggesting.

huangp: Right now we don't need it. But from Luke's design eventually we will have a similar view with tabs or different portions of the page configured to show different matrices.

carlosmunoz: :+1: ok

zanata-war/src/main/webapp/WEB-INF/urlrewrite.xml
```diff
@@ -196,6 +196,18 @@
...
+    <to
+      last="true">/profile/home.seam\?username=$1&amp;$4
```
carlosmunoz: ... the insertion of $4 at the end might look weird.

huangp:
@carlosmunoz all done. Can you please review this again?

carlosmunoz:
:+1:

djansen-redhat:
More work incoming, removed ON_QA

djansen-redhat:
- [ ] :small_red_triangle_down: Search bar should say Search Projects, People

djansen-redhat:
The last change didn't seem to work. I tried the jenkins war and a locally built one.

djansen-redhat:
The invalid username error now goes to an exception page - was this intended?

huangp:
fixed. Please also test it on production size database and see if the server can boot up correctly(adding indices to large tables).

djansen-redhat:
:white_check_mark: Tested.
Execution of liquibase table alteration is around 4 minutes.

PR #695, opened 2015-02-16
Invalidate doc statistic cache when uploading source and translation

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1192271

### DIFF
--- a/zanata-war/src/main/java/org/zanata/service/impl/TranslationStateCacheImpl.java
+++ b/zanata-war/src/main/java/org/zanata/service/impl/TranslationStateCacheImpl.java
@@ -181,18 +181,8 @@ public class TranslationStateCacheImpl implements TranslationStateCache {
                 new DocumentLocaleKey(event.getDocumentId(),
                         event.getLocaleId());
 
-        WordStatistic stats = documentStatisticCache.get(key);
-
-        if (stats != null) {
-            HTextFlow textFlow = getTextFlowDAO().findById(
-                    event.getTextFlowId());
-
-            stats.decrement(event.getPreviousState(),
-                    textFlow.getWordCount().intValue());
-            stats.increment(event.getNewState(),
-                    textFlow.getWordCount().intValue());
-            documentStatisticCache.put(key, stats);
-        }
+        //invalidate document statistic cache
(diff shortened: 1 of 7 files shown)


zanata-war/src/main/java/org/zanata/service/impl/TranslationStateCacheImpl.java
```diff
@@ -181,18 +181,8 @@ public void textFlowStateUpdated(TextFlowTargetStateEvent event) {
...
-                    event.getTextFlowId());
-
-            stats.decrement(event.getPreviousState(),
-                    textFlow.getWordCount().intValue());
-            stats.increment(event.getNewState(),
-                    textFlow.getWordCount().intValue());
-            documentStatisticCache.put(key, stats);
-        }
+        //invalidate document statistic
+        documentStatisticCache.remove(key);
```
carlosmunoz: We are currently doing something very efficient by altering the stats instead of re-querying everything. Is there something we can do to keep that mechanism?

seanf:
As per the discussion with @carlosmunoz , we should synchronise access to the methods of `AbstractStatistic` and subclasses, synchronise around the mutations in `TranslationStateCacheImpl.textFlowStateUpdated`, and, for now, just remove the cache entry whenever a batch of translations is uploaded in `TranslationServiceImpl.SaveBatchWork`.

And longer term, we should make the *Statistic objects immutable and generally look at overhauling the whole statistics design to minimise the need for synchronisation.

aeng:

zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java
```diff
@@ -688,6 +692,11 @@ protected Void work() throws Exception {
                     }
                 }.workInTransaction();
 
+                //invalidate document statistic
```
seanf: This should be inside the batch, and thus inside the transaction.

seanf:
:+1: Looks safer.  I just hope the cache invalidation doesn't kill performance.

aeng:
Merged into release. https://github.com/zanata/zanata-server/commit/4a2019594da3cb16180ddf59a8a304137000028d

PR #700, opened 2015-02-24
Fix password reset with null pointer exception

aeng:
fix null pointer exception when request for password reset.

### DIFF
--- a/zanata-war/src/main/java/org/zanata/action/PasswordResetRequestAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/PasswordResetRequestAction.java
@@ -71,18 +71,30 @@ public class PasswordResetRequestAction implements Serializable {
 
     public String requestReset() {
         if(getAccount() == null) {
-            FacesMessages.instance().add(msgs.get("jsf.account.notFound"));
-            return null;
-        } else if(isAccountWaitingForActivation()) {
+            return getAccountNoFoundMessage();
+        }
+
+        HAccountResetPasswordKey key =
+            userAccountServiceImpl.requestPasswordReset(getAccount());
+
+        if(key == null) {
+            return getAccountNoFoundMessage();
+        }
+
+        if(isAccountWaitingForActivation()) {
             FacesMessages.instance().add(msgs.get("jsf.account.notActivated"));
             return null;
         }
-        String message =
-                emailServiceImpl.sendPasswordResetEmail(
-                    getAccount().getPerson(),
-                    account.getAccountActivationKey().getKeyHash());
+
+        String message = emailServiceImpl.sendPasswordResetEmail(
+            getAccount().getPerson(), key.getKeyHash());
         FacesMessages.instance().add(message);
         return "home";
+    }
+
+    private String getAccountNoFoundMessage() {
+        FacesMessages.instance().add(msgs.get("jsf.account.notFound"));
+        return null;
     }
 
     @End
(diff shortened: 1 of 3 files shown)


zanata-war/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java
```diff
@@ -58,24 +58,22 @@
...
         HAccountResetPasswordKey key =
-                (HAccountResetPasswordKey) session
-                        .createCriteria(HAccountResetPasswordKey.class)
-                        .add(Restrictions.eq("account", account))
-                        .uniqueResult();
-        if (key != null) {
-            session.delete(key);
-            session.flush();
-        }
+                accountResetPasswordKeyDAO.findByAccount(account.getId());
```
huangp: It used to check whether the key is null. Is now safe to ignore the check?

aeng: Good one. it should check for null.

zanata-war/src/main/java/org/zanata/action/PasswordResetRequestAction.java
```diff
@@ -70,17 +70,19 @@
 
 
     public String requestReset() {
-        if(getAccount() == null) {
+        HAccountResetPasswordKey key =
+            userAccountServiceImpl.requestPasswordReset(getAccount());
+
+        if(getAccount() == null || key == null) {
```
seanf: You're using `getAccount()` (above) before checking if it's null.

zanata-war/src/main/java/org/zanata/action/PasswordResetRequestAction.java
```diff
@@ -112,10 +114,10 @@ public boolean isAccountWaitingForActivation() {
         if (account == null) {
             return false;
         }
-        if (account.getAccountActivationKey() == null) {
-            return false;
+        if (account.getAccountActivationKey() != null) {
+            return true;
         }
-        return true;
+        return false;
```
seanf: How about `return account.getAccountActivationKey() != null` ?

zanata-war/src/main/java/org/zanata/action/PasswordResetRequestAction.java
aeng: Good one.

zanata-war/src/main/java/org/zanata/service/impl/UserAccountServiceImpl.java
```diff
@@ -58,24 +58,24 @@
...
-        // TODO This should be done in a DAO
         HAccountResetPasswordKey key =
-                (HAccountResetPasswordKey) session
-                        .createCriteria(HAccountResetPasswordKey.class)
-                        .add(Restrictions.eq("account", account))
-                        .uniqueResult();
-        if (key != null) {
-            session.delete(key);
-            session.flush();
+                accountResetPasswordKeyDAO.findByAccount(account.getId());
```
seanf: +1 on the dao extraction.

huangp:
:+1:

djansen-redhat:
:white_check_mark: Tested.

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/3b0d55a8d4113479be77d0e9c667e9c4ef017066

PR #701, opened 2015-02-24
Fix security functional test error

djansen-redhat:


functional-test/src/test/java/org/zanata/feature/security/SecurityTest.java
```diff
@@ -141,12 +141,13 @@ public void invalidResetPasswordFieldEntries() {
...
-        assertThat(error.equals("size must be between 3 and 20") ||
-                error.equals("must match ^[a-z\\d_]{3,20}$"))
-                .isTrue()
-                .as("Invalid email error is displayed");
+        assertThat(resetPasswordPage.getErrors().get(0))
+                .isIn("Between 3 and 20 lowercase letters, numbers and " +
+                            "underscores only",
+                        "size must be between 3 and 20",
+                        "must match ^[a-z\\d_]{3,20}$")
+                .as("Invalid username error is displayed");
```
seanf: Strictly speaking, we should require only "Between 3 and 20 lowercase letters, numbers and underscores only", but that's a problem for another day.

seanf:
:+1:

PR #702, opened 2015-02-24
Fix failed functional test

aeng:


zanata-test-war/src/main/java/org/zanata/util/SampleProjectProfile.java
```diff
@@ -125,28 +126,41 @@ private void purgeLuceneIndexes() {
...
+            String displayName, String nativeName) {
+        return forLocale(enabledByDefault, localeId, displayName, nativeName)
+                .makeAndPersist(entityManager,
+                    HLocale.class);
     }
 
-    public void makeLanguage(LocaleId localeId) {
-        forLocale(true, localeId).makeAndPersist(entityManager,
-                HLocale.class);
+    public HLocale makeLanguage(boolean enabledByDefault, LocaleId localeId) {
```
seanf: It's generally good to add new parameters at the end of the list, rather than inserting them before the old ones.  Doesn't matter too much here though.

seanf:
All functional tests passed on EAP 6.  Running against WF8.1 now.

djansen-redhat:
:white_check_mark: Tested. This looks fine to me.

seanf:
Green on WF8.1 as well, by the way.

PR #706, opened 2015-03-02
Implement email reasons for admin

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1197572

### DIFF
--- a/zanata-war/src/main/java/org/zanata/service/impl/EmailServiceImpl.java
+++ b/zanata-war/src/main/java/org/zanata/service/impl/EmailServiceImpl.java
@@ -49,7 +49,10 @@ import org.zanata.model.HLocaleMember;
 import org.zanata.model.HPerson;
 import org.zanata.service.EmailService;
 
+import javax.annotation.Nullable;
 import javax.mail.internet.InternetAddress;
+
+import com.google.common.collect.Lists;
 
 import static org.zanata.email.Addresses.getAddresses;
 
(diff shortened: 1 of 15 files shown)


zanata-war/src/main/java/org/zanata/service/impl/EmailServiceImpl.java
```diff
@@ -136,16 +139,21 @@ public String sendPasswordResetEmail(HPerson person, String key) {
     }
 
     @Override
-    public String sendToAdmins(EmailStrategy strategy) {
+    public String sendToAdmins(EmailStrategy strategy,
```
carlosmunoz: Lets comment what received reasons are and how they will be used.

carlosmunoz:
:+1: with the documentation additions.
(I think the wording for those reasons captures what @davidmason suggested, but feel free to add more details.)

zanata-war/src/main/resources/messages.properties
```diff
@@ -1168,6 +1168,8 @@ jsf.email.admin.UserMessageIntro=Zanata user "{0}" with id "{1}" has sent the fo
 jsf.email.ReplyInstructions=You can reply to {0} at {1}
 jsf.email.admin.ReceivedReason=You are an administrator in the system configuration
 jsf.email.admin.user.ReceivedReason=You are an administrator
+jsf.email.admin.ReceivedReason.language.noCoordinator=There is no coordinator for language "{0} ({1})"
+jsf.email.admin.ReceivedReason.versionGroup.noMaintainer=There is no maintainer for version group "{0}"
```
davidmason: Suggest:

 - This message is intended for the language coordinators of {0} ({1}). No coordinators are assigned to {0} so this will be sent to server administrators instead.
- This message is intended for the maintainers of version group "{0}". No maintainers are assigned to version group "{0}" so this will be sent to server administrators instead.

aeng: I doubt anyone will actually read message this long. Personally I prefer short and simple.

definite:
- [ ] Coordinator
  - [ ] Mail to admin if no coordinator
  - [ ] NOT mail to admin if coordinator exists, and admin is not a coordinator.
   -[ ] Still mail to admin if admin also a coordinator .
 - [ ] Group maintainer
  - [ ] Mail to admin if no group maintainer
  - [ ] NOT mail to admin if group maintainer exists, and admin is not a group maintainer.
   -[ ] Still mail to admin if admin also a group maintainer .

definite:
:+1:

PR #707, opened 2015-03-03
Remove reference of helpContent

aeng:
Remove reference of helpContent from application.
Post cleanup from https://bugzilla.redhat.com/show_bug.cgi?id=1195384


zanata-model/src/main/java/org/zanata/model/HApplicationConfiguration.java
```diff
@@ -44,13 +44,18 @@
 @AllArgsConstructor
 public class HApplicationConfiguration extends ModelEntityBase {
 
+    /**
+     * obsolete data key. Help content was removed from UI
+     */
+    @Deprecated
+    public static String KEY_HELP_CONTENT = "pages.help.content";
```
seanf: ... "but the content was left in the database for reference purposes."

PR #712, opened 2015-03-04
Readthedocs

carlosmunoz:
Add markdown documentation. Not everything has been migrated from the different sources, but that will be easier to do so once we have it integrated so that everyone can chip in.


docs/configuration/email.md
```diff
@@ -0,0 +1,5 @@
+As of Zanata 3.6? (https://github.com/zanata/zanata-server/pull/633) email configuration is taken directly from JBoss/WildFly's standalone.xml configuration, using the mail session configured with the JNDI name "java:jboss/mail/Default".  In the default configuration of JBoss/WildFly, this expects an SMTP server on localhost port 25.
```
davidmason: Should this have a heading, or does it get a heading automatically from somewhere else and does not need it because it has only one section?

docs/configuration/jms-messaging.md
```diff
@@ -0,0 +1,106 @@
+Make sure your jboss standalone.xml has management native socket binding enabled.
```
davidmason: Heading needed?

docs/configuration/wildfly.md
```diff
@@ -0,0 +1 @@
+**Note:** _This is a work in progress and there may be issues when deploying Zanata to Wildfly_
```
davidmason: Needs more content :dog:

docs/user-guide/Opening-a-Translation-Project.md
```diff
@@ -0,0 +1,2 @@
+# Instructions on opening a translation project
+
```
davidmason: Needs more words.

For these sections that don't have their content yet, we could quickly add the URL for the current content. So that:

 - We have an easy reference when we are ready to add the content.
 - If we miss adding content for some reason, the user gets more than an empty page.

docs/configuration/email.md
carlosmunoz: Good catch as this hasn't even been integrated into the main branch yet. I will leave the content but I will comment it out of the index page until such time as it's ready to go back in.

davidmason:
May be out of scope for this, but we need a good way to version the screenshots. Ideally we would update any that would change in the same or an adjacent commit as the UI changes, but at very least before making a release.

docs/configuration/wildfly.md
carlosmunoz: Definitely, just moving our existing content from the other sources.

seanf:
Is this branch set up to do builds in readthedocs.org?

carlosmunoz:
@seanf I set it up at http://zanata.readthedocs.org/en/latest/
It's still obviously incomplete, but that's where I'll need everyone's help to get it up to par.

carlosmunoz:
Release notes are now in a single page.

seanf:
:+1:

PR #713, opened 2015-03-05
Ensure that a success message is shown when project about page is updated

davidmason:
Used to use the default "Successfully updated" message, but that has been overridden.


zanata-war/src/main/resources/messages.properties
```diff
@@ -290,6 +290,8 @@ jsf.languageSettings.LanguagesDisabled=Disabled languages: {0}.
 jsf.languageSettings.LanguageEnabled=Language "{0}" has been enabled.
 jsf.languageSettings.LanguagesEnabled=Enabled languages: {0}.
 jsf.project.LanguageUpdateFromGlobal=Updated languages from global settings.
+jsf.project.AboutPageUpdated=About page updated.
+jsf.project.AboutPageUpdateFailed=There was a problem while updating the about page.
```
djansen-redhat: while?

davidmason: If there was an error, I don't have any way here to know whether it was updated or not.

In practice the case where this message would be displayed may never come up - I think it would redirect to the error page if there was an exception, and the library code looks like it is hard-coded to indicate success (so an exception is the only way that it would not indicate success to my code).

djansen-redhat: Ah, so it *could* have been updated?

carlosmunoz:
:+1:

zanata-war/src/main/resources/messages.properties
davidmason: @djansen-redhat yes - basically if that message is displayed, the state is unknown to the code.

PR #714, opened 2015-03-05
Fix language tests

djansen-redhat:
Significant changes to language add/remove behaviour in projects.
Moves the test page behaviour to a utility class, as it is shared in
project and version settings.

### DIFF
--- a/functional-test/src/main/java/org/zanata/page/projects/projectsettings/ProjectLanguagesTab.java
+++ b/functional-test/src/main/java/org/zanata/page/projects/projectsettings/ProjectLanguagesTab.java
@@ -20,18 +20,14 @@
  */
 package org.zanata.page.projects.projectsettings;
 
-import com.google.common.base.Function;
-import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
-import com.google.common.collect.Lists;
 import lombok.extern.slf4j.Slf4j;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.zanata.page.projects.ProjectBasePage;
-import org.zanata.util.WebElementUtil;
+import org.zanata.util.LanguageList;
 
-import javax.annotation.Nullable;
 import java.util.List;
 
 /**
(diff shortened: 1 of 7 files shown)


functional-test/src/main/java/org/zanata/page/projects/projectsettings/ProjectLanguagesTab.java
```diff
@@ -43,8 +39,10 @@
 @Slf4j
 public class ProjectLanguagesTab extends ProjectBasePage {
 
-    private By addNewLanguageField = By.id("languageAutocomplete-autocomplete__input");
-    private By settingsLanguagesForm = By.id("settings-languages-form");
+    private By activeLocales = By.id("active-locales-list");
+    private By inactiveLocales = By.id("available-locales-list");
```
davidmason: Would it be easier to follow if all the identifiers stuck to `enabled` and `disabled`?

davidmason: At this point I realize that the `id` of the components doesn't help. Part of cleaning up that code will be to make all those consistently use `enabled` and `disabled` as well.

functional-test/src/main/java/org/zanata/page/projects/projectsettings/ProjectLanguagesTab.java
```diff
@@ -111,37 +119,24 @@ public ProjectLanguagesTab enterSearchLanguage(String languageQuery) {
      * Add a language to the languages list.
      * Assumes that the search string has already been entered.
```
davidmason: Does this comment about the search string still count, now that this does not use an autocomplete component?

functional-test/src/main/java/org/zanata/page/projects/projectsettings/ProjectLanguagesTab.java
```diff
@@ -111,37 +119,24 @@ public ProjectLanguagesTab enterSearchLanguage(String languageQuery) {
...
-                }
-                return clickedLocale;
+                return LanguageList.toggleLanguageInList(
+                        getDriver().findElement(inactiveLocales),
+                        searchLocaleId);
             }
         });
-
-
         refreshPageUntil(this, new Predicate<WebDriver>() {
```
davidmason: This looks like the equivalent of pressing F5 repeatedly (going by the method name). Is that right?

davidmason: Or does it just refresh selenium's model of the page?

djansen-redhat: No. Yes.

functional-test/src/main/java/org/zanata/page/projects/projectsettings/ProjectLanguagesTab.java
djansen-redhat: Fixed.

davidmason:
:+1:   :sake:

PR #715, opened 2015-03-06
Add test for setting a project language alias

djansen-redhat:


functional-test/src/main/java/org/zanata/util/LanguageList.java
```diff
@@ -36,7 +37,12 @@
     private LanguageList(){}
 
     private static By localeId = By.className("js-locale-id");
+    private static By localeAlias = By.className("js-locale-alias");
     private static By languageActions = By.className("button--group");
+    private static By dropdown = By.className("i--arrow-down");
+    private static By aliasInput = By.className("form--inline__input");
+    private static By setAlias = By.className("form--inline__addon");
```
davidmason: These identifiers look a bit fragile, it might be worth adding a few classes that will be consistent if we change the styles. We could do that now, or if the tests break in the future. Either option seems fine to me.

davidmason:
:+1: All looks good to me. Let me know if you want those better identifiers today, but they wouldn't be hard to add later if needed.
:dragon:

PR #717, opened 2015-03-06
Reduce unnecessary log warnings

seanf:
We really need to generate these files from a template, or perhaps a template for EAP and one for WildFly.


carlosmunoz:
:+1:

PR #719, opened 2015-03-09
Merge translation from existing version to another

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1133989


zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
```diff
@@ -117,10 +118,25 @@
...
+
+    private boolean revisionCommentChanged = false;
+
     // Only for internal use (persistence transient)
     @Setter(AccessLevel.PRIVATE)
     private Integer oldVersionNum;
 
+    public void setRevisionComment(String revisionComment) {
+        this.revisionComment = revisionComment;
+        revisionCommentChanged = true;
```
davidmason: Maybe it is the same. Suggest wrapping both lines in a null-safe check that the old and new value are different.

zanata-war/src/main/java/org/zanata/action/CopyTransAction.java
```diff
@@ -52,7 +54,7 @@
  */
 @Name("copyTransAction")
 @Scope(ScopeType.CONVERSATION)
-public class CopyTransAction implements Serializable, ProgressBar {
```
davidmason: Why is `ProgressBar` gone? Was it not being used?

zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
aeng: revisionComment is nullable. There's no need for null safe check.

zanata-war/src/main/java/org/zanata/action/CopyTransAction.java
aeng: Its been replaced with interface CopyAction which extends ProgressBar

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
```diff
@@ -0,0 +1,230 @@
...
+import org.zanata.common.EntityStatus;
+import org.zanata.dao.ProjectDAO;
+import org.zanata.dao.ProjectIterationDAO;
+import org.zanata.i18n.Messages;
+import org.zanata.model.HProject;
+import org.zanata.model.HProjectIteration;
+import org.zanata.ui.CopyAction;
+
+/**
+ * Handles action from merge_trans_modal.xhtml
```
davidmason: It is helpful to mention the template file that calls this class's methods. On its own, it does not give developers a quick overview of what this is about though (reading through xhtml to see what it does is not quick or easy), so can you write a sentence or two about the responsibilities this class has?

I imagine it would be something about handling actions from that template to start translation merges between different project-versions, and providing methods to display their progress.

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
```diff
@@ -0,0 +1,230 @@
...
+    @Getter
+    private String sourceProjectSlug;
+
+    @Getter
+    @Setter
+    private String sourceVersionSlug;
+
+    @Getter
+    @Setter
+    private boolean keepExistingTranslation;
```
davidmason: :thought_balloon:  Lots of `@Getter` and `@Setter` - it might be a bit cleaner to have them at the class level, and suppress the few exceptions with `AccessLevel.NONE`. Either way is fine. ¯\_(ツ)_/¯

zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
davidmason: e.g. `if (!StringUtils.equals(this.revisionComment, revisionComment)) {`

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
aeng: I prefer the `Getter` and `Setter` in individual field for easy reading.

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
```diff
@@ -0,0 +1,230 @@
...
+    public void setTargetProjectSlug(String targetProjectSlug) {
+        this.targetProjectSlug = targetProjectSlug;
+
+        /**
+         * This should only true during first instantiation. See pages.xml
+         * for setter of mergeTransAction.targetProjectSlug
+         */
+        if(sourceProjectSlug == null) {
+            setSourceProjectSlug(targetProjectSlug);
+        }
```
davidmason: Is it because we want the default source project to be the target project? We should state that explicitly so that future developers know why this code is here.

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    public void setSourceProjectSlug(String sourceProjectSlug) {
+        if(!StringUtils.equals(this.sourceProjectSlug, sourceProjectSlug)) {
+            this.sourceProjectSlug = sourceProjectSlug;
+            sourceProject = null;
```
davidmason: It looks like this is to force the `getSourceProject`  code to re-fetch the project. I would warp this in a method `refreshSourceProject()` or something to make it clear what the purpose is.

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
```diff
@@ -0,0 +1,230 @@
...
+            if (!getSourceProject().getProjectIterations().isEmpty()) {
+                sourceVersionSlug =
+                        getSourceProject().getProjectIterations().get(0)
+                                .getSlug();
+            }
```
davidmason: Will this keep the old `sourceVersionSlug` if the new project has no versions? It probably does no harm at the moment because it will not display anything, but it would be good to set it to `null` to make sure future changes do not cause unexpected behaviour.

As a tangent, it would be best if people could not select a project that has no versions, since it is useful.

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
```diff
@@ -0,0 +1,230 @@
...
+            targetVersion =
+                    projectIterationDAO.getBySlug(targetProjectSlug,
+                            targetVersionSlug);
+        }
+        return targetVersion;
+    }
+
+    public HProject getSourceProject() {
+        if(sourceProject == null) {
+            sourceProject = projectDAO.getBySlug(sourceProjectSlug);
```
davidmason: I think this will throw an exception if `sourceProjectSlug` is `null`. I think it would be worth looking at all usages of the 4 slugs in here and make sure they will be ok if the value is null.

zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
aeng: It can be the same revisionComment, and that should flag `revisionCommentChanged` to true.
e.g, same target being copied by Merge-trans twice from different version.

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
```diff
@@ -0,0 +1,230 @@
...
+
+        if(versions.isEmpty()) {
+            return Collections.emptyList();
+        }
+
+        //remove target version if both are the same project or obsolete
+        if (StringUtils.equals(sourceProjectSlug, targetProjectSlug)) {
+            for (HProjectIteration version : versions) {
+                if (version.getSlug().equals(targetVersionSlug)
+                        || version.getStatus().equals(EntityStatus.OBSOLETE)) {
```
davidmason: I think this check for obsolete versions should happen regardless whether it is the same project. This looks like it would still show obsolete versions if they are in a different project.
Maybe remove the first `if` and move its condition in front of the second `if` so that it is

```
if (version.getStatus().equals(EntityStatus.OBSOLETE) ||
        StringUtils.equals(sourceProjectSlug, targetProjectSlug) &&
        version.getSlug().equals(targetVersionSlug)) {
```

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
```diff
@@ -0,0 +1,230 @@
...
+                }
+            }
+        }
+        return versions;
+    }
+
+    public List<HProject> getProjects() {
+        return projectDAO.getProjects(EntityStatus.ACTIVE,
+                EntityStatus.READONLY);
+    }
```
davidmason: Could filter the projects here to only show projects that have 1 or more non-obsolete versions. This might be too surprising to users though, so maybe it is better just to leave it as-is.

@lukebrooker any thoughts?

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
```diff
@@ -0,0 +1,230 @@
...
+
+    public void startMergeTranslations() {
+        if (StringUtils.isEmpty(sourceProjectSlug)
+                || StringUtils.isEmpty(sourceVersionSlug)
+                || StringUtils.isEmpty(targetProjectSlug)
+                || StringUtils.isEmpty(targetVersionSlug)) {
+            FacesMessages.instance().add(StatusMessage.Severity.WARN,
+                msgs.get("jsf.iteration.mergeTrans.noSourceAndTarget"));
```
davidmason: I think error severity would be better here, since it is a total failure of the process.

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
```diff
@@ -0,0 +1,230 @@
...
+                msgs.get("jsf.iteration.mergeTrans.hasCopyActionRunning"));
+            return;
+        }
+        mergeTranslationsManager.startMergeTranslations(sourceProjectSlug,
+                sourceVersionSlug, targetProjectSlug, targetVersionSlug,
+                !keepExistingTranslation);
+    }
+
+    // Check if copy-trans, copy version or merge-trans is running for given
+    // version
```
davidmason: "for given version" suggests that you can give a version as the argument. I suggest "for the target version" instead.

zanata-war/src/main/java/org/zanata/action/MergeTranslationsManager.java
```diff
@@ -0,0 +1,113 @@
...
+import org.jboss.seam.annotations.In;
+import org.jboss.seam.annotations.Name;
+import org.jboss.seam.annotations.Scope;
+import org.jboss.seam.security.Identity;
+import org.zanata.async.AsyncTaskHandleManager;
+import org.zanata.async.handle.MergeTranslationsTaskHandle;
+import org.zanata.service.MergeTranslationsService;
+
+/**
+ * Manages copy translations from existing version to another tasks.
```
davidmason: Slightly confusing to read, I suggest "Manages tasks to copy translations from one existing version to another."

zanata-war/src/main/java/org/zanata/action/MergeTranslationsManager.java
```diff
@@ -0,0 +1,113 @@
...
+    @In
+    private Identity identity;
+
+    /**
+     * Merge translations from an existing version to another.
+     *
+     * @param sourceProjectSlug - source project identifier
+     * @param sourceVersionSlug - source version identifier
+     * @param targetProjectSlug - target project identifier
+     * @param targetVersionSlug - target version identifier
```
davidmason: These sort of things seem like a waste of space to me, but I know others in the team think that we should include them even though they do not provide any more information than the parameter name.

zanata-war/src/main/java/org/zanata/action/MergeTranslationsManager.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Key used for copy version task
+     *
+     */
+    @EqualsAndHashCode
+    @Getter
+    @AllArgsConstructor
+    public static final class MergeTranslationsKey implements Serializable {
```
davidmason: All the methods and this class could all have shorter names since they are in `MergeTranslationsManager`. Could just call them `start`, `cancel`, `getProcessHandle`, `isRunning` and `Key`.

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
```diff
@@ -0,0 +1,230 @@
...
+                (double) handle.getCurrentProgress() / (double) handle
+                    .getMaxProgress() * 100;
+            if (Double.compare(completedPercent, 100) == 0) {
+                onComplete();
+            }
+            return PERCENT_FORMAT.format(completedPercent);
+        } else {
+            return "0";
+        }
+    }
```
davidmason: It looks like this code has been copied around several times. It should go in a utility class, or maybe pulled up into the task handle class.

zanata-war/src/main/java/org/zanata/service/MergeTranslationsService.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2010, Red Hat, Inc. and individual contributors
```
davidmason: 2015

zanata-war/src/main/java/org/zanata/action/VersionHomeAction.java
```diff
@@ -324,12 +326,27 @@ public boolean isInProgress() {
...
             if (handle != null) {
                 double completedPercent =
                         (double) handle.getCurrentProgress() / (double) handle
                                 .getMaxProgress() * 100;
+                if (Double.compare(completedPercent, 100) == 0) {
+                    onComplete();
+                }
                 return PERCENT_FORMAT.format(completedPercent);
             } else {
                 return "0";
```
davidmason: Another copy of the completed percentage code...

zanata-war/src/main/java/org/zanata/dao/ProjectDAO.java
```diff
@@ -433,4 +435,22 @@ public int getMaintainedProjectCount(HPerson maintainer, String filter) {
                         .setParameter("filter", "%" + sqlFilter + "%");
         return ((Long) q.uniqueResult()).intValue();
     }
+
+    public List<HProject> getProjects(EntityStatus... includeStatus) {
```
davidmason: This does almost the same thing as `getOffsetListOrderByName(int offset, int count, boolean filterActive, boolean filterReadOnly, boolean filterObsolete)`, just with offset 0 and no count. Can the two be combined, or just use the existing one with `0` and `Integer.MAX_VALUE`, rather than create another method that we have to maintain?

zanata-war/src/main/java/org/zanata/dao/TextFlowDAO.java
```diff
@@ -300,4 +302,73 @@ public long countActiveTextFlowsInProjectIteration(Long versionId) {
         return builder.build();
     }
 
+    /**
+     * Return text flows from HProjectIteration#id=sourceVersionId and
+     * HProjectIteration#id=targetVersionId which has matching content
+     * hash and document id.
```
davidmason: Suggest "Return text flows that have matching document id and content between the given source and target version".

zanata-war/src/main/java/org/zanata/dao/TextFlowDAO.java
```diff
@@ -300,4 +302,73 @@ public long countActiveTextFlowsInProjectIteration(Long versionId) {
...
+        for(Object result: query.list()) {
+            Object[] castResults = (Object[]) result;
+            results.add(new HTextFlow[] { (HTextFlow) castResults[0],
+                (HTextFlow) castResults[1] });
+        }
+        return results;
+    }
+
+    public int getSourceByMatchedContextCount(Long sourceVersionId,
+        Long targetVersionId) {
```
davidmason: This is mostly the same as `getSourceByMatchedContext`, just the first line of the query, a few params, and the way that results are collated look different. Could a method be added that takes the `StringBuilder` after the select clause is set, and add all the common lines to it?

zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    /**
+     * Check if sourceVersion or targetVersion has source document.
+     *
+     * @param sourceVersion
+     * @param targetVersion
+     */
+    private boolean isVersionsEmpty(HProjectIteration sourceVersion,
+            HProjectIteration targetVersion) {
```
davidmason: This seems like something we should really be checking when a user selects a project (showing "empty" next to versions in the list) or when they select an empty version (disable the button to start a merge if source or target project-version is empty).

zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java
```diff
@@ -0,0 +1,278 @@
...
+
+    // @formatter:off
+    /**
+     * Rule of which translation should merge
+     * |          from         |       to         |   copy?   |
+     * |-----------------------|------------------|-----------|
+     * |fuzzy/untranslated     |       any        |     no    |
+     * |-----------------------|------------------|-----------|
+     * |different source text/ |                  |           |
+     * |document id            |       any        |     no    |
```
davidmason: The original table also has different context. Was this just missed by mistake, or removed purposely? Is the code checking for context?

zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsWork.java
```diff
@@ -0,0 +1,137 @@
+package org.zanata.service.impl;
```
davidmason: I think this needs the copyright header.

zanata-war/src/main/java/org/zanata/ui/CopyAction.java
```diff
@@ -0,0 +1,36 @@
...
+ *  * This software is distributed in the hope that it will be useful, but WITHOUT
+ *  * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ *  * FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ *  * details.
+ *  *
+ *  * You should have received a copy of the GNU Lesser General Public License
+ *  * along with this software; if not, write to the Free Software Foundation,
+ *  * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA, or see the FSF
+ *  * site: http://www.fsf.org.
+ */
```
davidmason: There is an extra column of stars here for some reason.

zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsWork.java
```diff
@@ -0,0 +1,137 @@
...
+
+        List<HTextFlow[]> matches =
+                textFlowDAO.getSourceByMatchedContext(
+                        sourceVersionId, targetVersionId, batchStart,
+                        batchLength);
+
+        for (HTextFlow[] results : matches) {
+            HTextFlow sourceTf = results[0];
+            HTextFlow targetTf = results[1];
+            boolean hasChanged = false;
```
davidmason: I would call this `foundChange` so that it is always true. We know right here that we did not find a change yet so we can honestly set it to `false`, and we know that if we get to the end and we have not found a change, we can assume there are no changes.

zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsWork.java
```diff
@@ -0,0 +1,137 @@
...
+                            targetTft.setComment(hComment);
+                        }
+                        hComment
+                                .setComment(sourceTft.getComment()
+                                    .getComment());
+                    }
+                    targetTft.setRevisionComment(MessageGenerator
+                            .getMergeTranslationMessage(sourceTft));
+
+                    raiseSuccessEvent(targetTft, oldState);
```
davidmason: Line 84-107 could easily be extracted to a method, since this is way too big.

Could probably also extract the whole of the HTextFlow block and the whole of the HLocale block to separate methods to make this a lot easier to follow. I should be able to see the whole of each method on my screen at once is possible.

zanata-war/src/main/java/org/zanata/action/CopyTransAction.java
davidmason: Ah, I figured it out. Just didn't occur to me that `CopyAction` extends `ProgressBar`.

zanata-war/src/main/java/org/zanata/util/MessageGenerator.java
```diff
@@ -0,0 +1,115 @@
...
+ */
+
+package org.zanata.util;
+
+import org.apache.commons.lang.StringUtils;
+import org.zanata.model.HDocument;
+import org.zanata.model.HTextFlowTarget;
+
+/**
+ * Generate messages/comments of Zanata business actions
```
davidmason: Are these all messages that will be added as comments to say how the translation was copied there? If so it would be helpful to mention that in the description, so that we can get a better idea of the purpose of this class.

zanata-war/src/main/java/org/zanata/util/MessageGenerator.java
```diff
@@ -0,0 +1,115 @@
...
+            author = tft.getLastModifiedBy().getName();
+        }
+        return generateAutoCopiedMessage(PREFIX_COPY_TRANS,
+                document.getProjectIteration()
+                        .getProject().getName(), document.getProjectIteration()
+                        .getSlug(), document.getDocId(), author);
+    }
+
+    /**
+     * Create Zanata generated comment for copy version
```
davidmason: In all these, is "Zanata generated" meaning that these messages are being generated by Zanata, or does it refer to the fact that the messages are saying that Zanata has generated the translations somehow? I think it might be clearer just to leave those two words out.

zanata-war/src/main/resources/db/changelogs/db.changelog-3.7.xml
```diff
@@ -27,4 +27,18 @@
     </addColumn>
   </changeSet>
 
+  <changeSet id="3" author="aeng@redhat.com">
+    <comment>Add revisionComment HTextFlowTarget, HTextFlowTargetHistory</comment>
```
davidmason: Is it "Add revisionComment *to* HTextFlowTarget...."?

zanata-war/src/main/webapp/WEB-INF/layout/version/merge_trans_modal.xhtml
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    function displayProjectSelector(display) {
+      if(display) {
+        jQuery(getProjectSelector()).removeClass('is-hidden');
+        jQuery('#project-static').addClass('is-hidden');
+      } else {
+        jQuery(getProjectSelector()).addClass('is-hidden');
+        jQuery('#project-static').removeClass('is-hidden');
+      }
```
davidmason: I think you can do this instead:

```
jQuery(getProjectSelector()).toggleClass('is-hidden', !display);
jQuery('#project-static').toggleClass('is-hidden', display);
```

zanata-war/src/test/java/org/zanata/dao/TextFlowTargetDAOTest.java
```diff
@@ -48,26 +48,26 @@
...
-                DatabaseOperation.CLEAN_INSERT));
+            "org/zanata/test/model/LocalesData.dbunit.xml",
+            DatabaseOperation.CLEAN_INSERT));
     }
 
     @BeforeMethod(firstTimeOnly = true)
     public void setup() {
         textFlowTargetDAO =
-                new TextFlowTargetDAO((Session) getEm().getDelegate());
+            new TextFlowTargetDAO((Session) getEm().getDelegate());
```
davidmason: It looks like there are no changes in this file. Why has it been reformatted?

zanata-war/src/test/java/org/zanata/service/impl/MergeTranslationsServiceImplTest.java
```diff
@@ -0,0 +1,307 @@
...
+    @Test
+    public void testDocIdAndContentHash() {
+        /**
+         * this test is covered by @see org.zanata.dao.TextFlowTargetDAOTest#
+         * testGetTranslationsByMatchedContext
+         *
+         * check different docId won't copy check different tf.contentHash won't
+         * copy
+         */
+    }
```
davidmason: I think an empty test is not the way to communicate this. Probably better just to make this a comment, maybe in the doc comment in this class.

zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
davidmason: Maybe it would be better to call it `revisionCommentSet` then, since that could mean it is set to the same thing, that way there won't be danger of someone else also thinking that it should have a check.

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
davidmason: Fair enough.

davidmason:
@aeng finished initial review, see comments. Changes so far look good.

zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
aeng: I don't see any difference between changing the `set` word in the method name. The general Java `setter` method doesn't limit to just set to different value, it's just a setter method of a value. The rules or limit of allowing to set same value or not should be sitting in business layer or the service class. 

In this case, same revisionComment value are allowed. Because it changes the behaviour of EntityListener for HTextFlowTarget when persisting.

zanata-war/src/main/java/org/zanata/action/MergeTranslationsManager.java
aeng: This is like saying having a comment of `check if string is empty` in `StringUtils.isStringEmpty()` method if waste of space.

zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java
aeng: what context do we meant in here? locale id?

zanata-war/src/main/java/org/zanata/util/MessageGenerator.java
aeng: For now, its only that. But it can be extend to any message that is generated by Zanata. I've included comments on each method.

zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java
aeng: Preloading all these data (versions with filter out obsolete state) is not ideal and heavy especially when we load all projects available in Zanata in the drop down.

lukebrooker: Initially I was only going to load projects the user is currently the maintainer of, not everything.

I would think use a drop down for that. I also have to create a filterable drop down.

Thanks,

Luke Brooker
@lukebrooker
http://lukebrooker.com

aeng: How about user who wish to copy from other projects?

aeng:
@davidmason its ready to review again with changed made.

zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java
lukebrooker: I'm not saying we definitely shouldn't do it but we need to separate the lists and have a better selection method.

I was thinking a drop down with tabs and the initial activated tab is 'my projects' and the other tab is all projects.

I think initially this would work with just the ones you maintain. If you really want to copy another one they can add you as a maintainer.

Do we have any use cases for copying from a project you don't maintain?

aeng: @lukebrooker project maintainer can copy translation into their projects from any other projects. They don't need to be maintainer of the `copy from` project.

carlosmunoz:
@aeng :memo: Please don't forget to add documentation about this new feature (under the `docs` folder)

zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
davidmason: @aeng I don't mean change the method name, I mean change `private boolean revisionCommentChanged` to `private boolean revisionCommentSet`

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
```diff
@@ -0,0 +1,226 @@
...
+
+    private void refreshSourceProject() {
+        sourceProject = null;
+    }
+
+    public HProjectIteration getTargetVersion() {
+        if(targetVersion == null) {
+            targetVersion =
+                    projectIterationDAO.getBySlug(targetProjectSlug,
+                            targetVersionSlug);
```
davidmason: This will throw an exception when `targetVersionSlug` is `null`.

zanata-war/src/main/java/org/zanata/dao/ProjectDAO.java
davidmason: Looking at this now, I think it would be better for readability to replace the one that takes some booleans with one that takes the list of `EntityStatus`, but I think there is a significant amount of code using the boolean one so it would not be a good change to make right now. Better to just keep the existing method for now. :+1:

zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java
davidmason: > If you really want to copy another one they can add you as a maintainer.

I disagree with this for the same reason I would not want to give someone commit access on my github projects just so they can fork them.

davidmason: For our production servers, the dropdown would have a lot of projects so it would be nice to have it able to be filtered by typing the name at least, but it is not an everyday thing to do so I think we should only try to do something like that if the story would still be completed this sprint, otherwise we should just make a better project-version selection widget an enhancement for later.

Can we separate the project-version selection part of the ui into its own component with some params so that we can use it in other places and have them all get enhancements for it? It may not be worthwhile yet since it should be easy to separate out later.

zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java
davidmason: @aeng I assume it is meant to be the same as in copytrans, which says it is `resId` and `msgctxt`.
@carlosmunoz I believe you were the one to mention this, can you confirm it?

zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
aeng: ah.. alright

zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java
carlosmunoz: I think the table in the bugzilla means that all those things (source, resId, msgctxt, docId) need to match for a copy to be made.

aeng: At the moment, we merge translations that has matching `docid`, `textFlow.contentHash`, and in same language. Not sure about resId and msgctxt as this might limit the merge if its 2 different project.

aeng: Alright. I missed out that part. Will make the change now.

aeng: Confirmed check
- textFlow.contentHash, textFlow.docId, textFlow.resId, textFlowTarget.locale

zanata-war/src/main/resources/messages.properties
```diff
@@ -635,9 +635,20 @@ jsf.iteration.NoDocumentInVersion=No documents in this version
...
+jsf.iteration.mergeTrans=Merge Translations
+jsf.iteration.mergeTrans.description1=All <dfn title='Translation which has identical language, source string, document name.'>matching</dfn> <strong><span class='txt--success'>translated</span>/<span class='txt--highlight'>approved</span></strong> translations from the source version will be copied to the target version.
+jsf.iteration.mergeTrans.description2=If there is an <strong>existing <span class='txt--success'>translated</span>/<span class='txt--highlight'>approved</span></strong> translation, the <strong>newer</strong> translation will be used.
+jsf.iteration.mergeTrans.toVersion.label=To <span class="heading--secondary">(Target)</span>
+jsf.iteration.mergeTrans.fromVersion.label=From <span class="heading--secondary">(Source)</span>
+jsf.iteration.mergeTrans.keepExistingTranslated=Keep <strong>existing <span class='txt--success'>translated</span>/<span class='txt--highlight'>approved</span></strong> strings
+jsf.iteration.mergeTrans.keepExistingTranslated.Description=Don't replace existing translated/approved translations with newer translations from this source.
```
davidmason: I have heard from docs people that it is better not to use contractions - i.e. "Do not replace..."

docs/user-guide/merge-translations.md
```diff
@@ -0,0 +1,59 @@
+# Merge translations from another project version
+
+Merge translations allows project maintainers to copy translations across different project version based on matching context. (See below for context matching rules)
+
+## Restrictions
+
+- This feature only available to project maintainers.
```
davidmason: feature **is** only

docs/user-guide/merge-translations.md
```diff
@@ -0,0 +1,59 @@
...
+    </tr>
+     <tr>
+        <td>translated/approved</td><td>translated/approved</td><td>copy if `From` is newer and <br />`Keep existing translated/approved translations` is unchecked</td>
+    </tr>
+</table>
+     
+## Run Merge translations
+
+1. Login to Zanata.
+1. Select a project version of which you wish to copy translations to.
```
davidmason: "Select a project version you wish to copy translations to." (current line does not make grammatical sense)

docs/user-guide/merge-translations.md
```diff
@@ -0,0 +1,59 @@
...
+</figure>
+1. If you do not want to overwrite translated/approved translations, ensure that `Keep existing translated/approved translations` is checked. If this option is not checked, translated/approved translations will be replaced with newer translated/approved translations if they are available.
+1. Click `Merge Translations` button to start the process.
+1. The progress of the merge process is shown by a progress bar on the version page.
+<figure>
+    <img alt="Merge translation in progress" src="images/version-merge-trans-progress.png" />
+</figure>
+
+## Cancel Merge translation
+**_Note: This will only stop additional translations being merged. Any translations that have already been merged will remain merged. **
```
davidmason: I think the space before the final `**` may prevent it rendering as bold. Might need to check this.

davidmason:
@aeng it looks like there are a few formatting issues: see the note under "Cancel Merge translation" and everything in the table. I think that is the only remaining issue. This is where I am seeing the trouble with formatting: https://github.com/zanata/zanata-server/blob/rhbz1133989/docs/user-guide/merge-translations.md

aeng:
@davidmason looking at the page using `mkdocs serve`. The table looks fine.

davidmason:
:+1: I think that's everything.

docs/user-guide/merge-translations.md
aeng: Fixed that in the last commit with _**.

definite:
Test cases:
- [ ] File Check
  - [ ] Filename and Directory name are same:
  - [ ] Filename same, but Directory different: NOP
  - [ ] Filename different,  but Directory same: NOP
- [ ] To Untranslated
  - [ ] Untranslated -> Untranslate: NOP
  - [ ] Fuzzy -> Untranslate: NOP
  - [ ] Translate -> Untranslate 
      - [ ] Same msgctxt
      - [ ] Different msgctxt
  - [ ] Rejected -> Untranslate: NOP
  - [ ] Approved -> Untranslate:
      - [ ] Same msgctxt
      - [ ] Different msgctxt
- [ ] To Fuzzy
  - [ ] Untranslated -> Fuzzy: NOP
  - [ ] Fuzzy -> Fuzzy: NOP
  - [ ] Translate -> Fuzzy
      - [ ] Same msgctxt
      - [ ] Different msgctxt
  - [ ] Rejected -> Untranslate: NOP
  - [ ] Approved -> Untranslate:
      - [ ] Same msgctxt
      - [ ] Different msgctxt
- [ ] To Rejected
  - [ ] Untranslated -> Rejected: NOP
  - [ ] Fuzzy -> Translated: NOP
  - [ ] Translate -> Translated
      - [ ] Same msgctxt
      - [ ] Different msgctxt
  - [ ] Rejected -> Translate: NOP
  - [ ] Approved -> Translate:
      - [ ] Same msgctxt
      - [ ] Different msgctxt
- [ ] To Translated
  - [ ] Untranslated -> Translated: NOP
  - [ ] Fuzzy -> Translated: NOP
  - [ ] Translate -> Translated
      - [ ] Same msgctxt
        - [ ] TO is older
          - [ ] Keep existing translation is OFF
          - [ ] Keep existing translation is ON
        - [ ] TO is newer
      - [ ] Different msgctxt
  - [ ] Rejected -> Translate: NOP
  - [ ] Approved -> Translate:
      - [ ] Same msgctxt
      - [ ] Same msgctxt
        - [ ] TO is older
          - [ ] Keep existing translation is OFF
          - [ ] Keep existing translation is ON
        - [ ] TO is newer
      - [ ] Different msgctxt
- [ ] To Approved
  - [ ] Untranslated -> Approved: NOP
  - [ ] Fuzzy -> Approved: NOP
  - [ ] Translate -> Approved
      - [ ] Same msgctxt
        - [ ] TO is older
          - [ ] Keep existing translation is OFF
          - [ ] Keep existing translation is ON
        - [ ] TO is newer
      - [ ] Different msgctxt
  - [ ] Rejected -> Approved: NOP
  - [ ] Approved -> Approved:
      - [ ] Same msgctxt
      - [ ] Same msgctxt
        - [ ] TO is older
          - [ ] Keep existing translation is OFF
          - [ ] Keep existing translation is ON
        - [ ] TO is newer
      - [ ] Different msgctxt

definite:
:+1:

aeng:
@definite merge integration/master into this branch to resolve conflict and making jenkins built before doing the merge. Thanks

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/bacdfcdba3dafc7c20f66e504b2aa9458ead53ae

PR #721, opened 2015-03-10
fail fast if jms is not configured properly

huangp:


seanf:
:+1: but I think we should put this in `release` too.

zanata-war/src/main/java/org/zanata/notification/NotificationManager.java
```diff
@@ -59,6 +60,19 @@
...
 
+
+    @Create
+    public void onCreate() {
+        try {
+            mailQueueSender.getQueue();
+        } catch (JMSException e) {
+            // it will never reach this block. As long as you call getQueue()
+            // and if the queue is not defined, seam will terminate:
+            // org.jboss.seam.jms.ManagedQueueSender.getQueue(ManagedQueueSender.java:45
```
djansen-redhat: Is there a missing closing bracket here?

djansen-redhat:
I don't know if/when/how this is supposed to fail.  Is it supposed to occur at start-up, and halt start-up of the server?
A few more details would be nice, if there is no attached bug report.

seanf:
Yes, the app server should refuse to deploy the war file.  The server will
still be running, but with an ugly error message (and returning 404, I
think).

djansen-redhat:

PR #724, opened 2015-03-11
Change resId for plaintext and libreoffice to use content hash.

davidmason:
This includes migration of resId in all existing documents of the above types to be the content hash.

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1194543

### DIFF
--- a/zanata-war/src/main/java/org/zanata/adapter/OpenOfficeAdapter.java
+++ b/zanata-war/src/main/java/org/zanata/adapter/OpenOfficeAdapter.java
@@ -31,7 +31,7 @@ import net.sf.okapi.filters.openoffice.OpenOfficeFilter;
  */
 public class OpenOfficeAdapter extends OkapiFilterAdapter {
     public OpenOfficeAdapter() {
-        super(prepareFilter(), IdSource.subDocNameAndTextUnitId, true);
+        super(prepareFilter(), IdSource.contentHash, true);
     }
 
     private static OpenOfficeFilter prepareFilter() {
--- a/zanata-war/src/main/java/org/zanata/liquibase/custom/ChangePositionalResIdToContentHash.java
+++ b/zanata-war/src/main/java/org/zanata/liquibase/custom/ChangePositionalResIdToContentHash.java
@@ -106,6 +106,8 @@ public class ChangePositionalResIdToContentHash implements CustomTaskChange {
                 selectDocumentsOfType = sdot;
                 selectTextFlows = stf;
 
+                log.info("Changing resource id from positional to content hash for text flows in some document types.");
+
                 for (DocumentType type : documentTypes) {
                     migrateResId(type);
                 }
@@ -129,7 +131,7 @@ public class ChangePositionalResIdToContentHash implements CustomTaskChange {
         try (ResultSet countOfType = countDocumentsOfType.executeQuery()) {
             if (countOfType.next()) {
                 docsOfType = countOfType.getLong(1);
-                log.info("processing " + docsOfType + " documents of type " + type.name());
+                log.info("preparing to process " + docsOfType + " documents of type " + type.name());
             } else {
                 // unexpected error, but does not have to destroy everything.
                 throw new SQLException("Count query returned no rows.");
@@ -140,14 +142,12 @@ public class ChangePositionalResIdToContentHash implements CustomTaskChange {
         try (ResultSet documentIds = selectDocumentsOfType.executeQuery()) {
             while (documentIds.next()) {
                 Long documentId = documentIds.getLong(1);
-                migrateResIdForDocument(documentId);
+                int textFlowsInDocument = migrateResIdForDocument(documentId);
                 docsProcessed++;
-                if (docsProcessed % 100 == 0) {
-                    log.info("    processed " + docsProcessed + " of " + docsOfType);
-                }
+                log.info("        processed " + textFlowsInDocument + " text flows in document " + docsProcessed + " of " + docsOfType);
                 totalDocs++;
             }
-            log.info("    processed " + docsProcessed + " of " + docsOfType);
+            log.info("    finished processing " + docsProcessed + " documents of type " + type.name());
         }
     }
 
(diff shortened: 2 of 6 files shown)


aeng:
I'm not sure about ChangePositionalResIdToContentHash class. Personally I would not prefer to have a java class sitting around in WAR forever that only runs 1 time. Anyway to convert it to sql statement in changelog xml file?

davidmason:
@aeng since the content hash is already available in the database, I guess it would technically be possible to write this completely in SQL. The duplicate id problem would make it quite complicated though, I definitely don't know enough SQL right now to know how to approach that - basically I would need to keep a variable to know how much to adjust `pos`, and would need to keep a record of which ids I have already used - I think all that is possible, but may be messy.

zanata-war/src/main/java/org/zanata/liquibase/custom/ChangePositionalResIdToContentHash.java
```diff
@@ -0,0 +1,224 @@
...
+        try {
+            prepareStatements((JdbcConnection) database.getConnection());
+            // could use try-with-resources to define the PreparedStatements here
+            // but making them fields allows for cleaner code.
+            try {
+                for (DocumentType type : documentTypes) {
+                    migrateResId(type);
+                }
+            } finally {
+                closeStatements();
```
seanf: This won't be called if there is an exception preparing the second or third statements.  I think you should use try with resources, but you can still leave them as fields.

zanata-war/src/main/java/org/zanata/liquibase/custom/ChangePositionalResIdToContentHash.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+        selectDocumentsOfType.setString(1, type.name());
+        try (ResultSet documentIds = selectDocumentsOfType.executeQuery()) {
+            while (documentIds.next()) {
+                Long documentId = documentIds.getLong(1);
+                migrateResIdForDocument(documentId);
+                docsProcessed++;
+                if (docsProcessed % 100 == 0) {
+                    log.info("    processed " + docsProcessed + " of " + docsOfType);
```
seanf: :+1: on the progress logging, but it could be a little confusing if it says:

    processed 0 of 250
    processed 100 of 250
    processed 200 of 250

but never says

    processed 250 of 250

seanf:
Looks pretty good.  Have you tested the migration against a production-sized database?

Does this leave any document types still using position-based `resId`s?

zanata-war/src/main/java/org/zanata/adapter/OpenOfficeAdapter.java
```diff
@@ -31,7 +31,7 @@
  */
 public class OpenOfficeAdapter extends OkapiFilterAdapter {
     public OpenOfficeAdapter() {
-        super(prepareFilter(), IdSource.subDocNameAndTextUnitId, true);
```
seanf: To avoid this problem happening again, I think we should put a warning on `IdSource.subDocNameAndTextUnitId` and `IdSource.textUnitId`, perhaps deprecating if not removing them.

seanf:
On 2015-03-12 07:18, Alex Eng wrote:
> I'm not sure about ChangePositionalResIdToContentHash class. Personally
> I would not prefer to have a java class sitting around in WAR forever
> that only runs 1 time. Anyway to convert it to sql statement in
> changelog xml file?

That SQL statement would also have to hang around in the war file just
as long as the class.

Well, we may be able eliminate the class from zanata.war if and when we
rebase the database, at which point we could move it (and the old
changesets) into a zanata DB migration utility.

But I think using a class for this is okay, in view of the duplicate id
problem David mentioned.


-- 
Sean Flanigan

Principal Software Engineer
Red Hat

zanata-war/src/main/java/org/zanata/liquibase/custom/ChangePositionalResIdToContentHash.java
davidmason: @seanf Do you know a way to use try-with-resources to set fields? I couldn't find a way. I guess I can always just bind them to fields at the top of the try block.

zanata-war/src/main/java/org/zanata/liquibase/custom/ChangePositionalResIdToContentHash.java
davidmason: @seanf I wanted to avoid it logging progress for types that have few documents since that would just be a lot of noise with no value. I guess I could add after the `while` loop:

```
if (docsProcessed > 100) {
    log.info(" processed " + docsProcessed + " of " + docsOfType);
}
```

(or just log that unconditionally).

davidmason:
@seanf 

> Does this leave any document types still using position-based resIds?

That depends what the filters return. I don't think any of the current types that use the id produced by the filter are getting position-based identifiers from the filter.

zanata-war/src/main/java/org/zanata/adapter/OpenOfficeAdapter.java
davidmason: If we really want to avoid this problem, we should be fixing it at the merge algorithm in `ResourceUtils.transferFromTextFlows(List<TextFlow>, HDocument)`. As long as we have the REST API we can never guarantee that id will always be sensible or consistent, so we're better off working well regardless of id, and let clients and adapters use whatever id makes sense for the relevant format or environment.

davidmason: @seanf deprecating them would be excessive, since the validity of `textUnitId` as a value for `resId` is dependent on the filter.

seanf:
:+1:

davidmason:
@zanata-jenkins retest this please.

djansen-redhat:
:-1: No space between number of documents processed and the word "documents"

djansen-redhat:
Entered
```
<system-properties>
    <property name="jboss.as.management.blocking.timeout" value="1000"/>
```
in standalone.xml, otherwise it times out at 300

djansen-redhat:
:white_check_mark: Tested.
Requires entry in release notes re startup timeout, but otherwise good.

seanf:
The timeout value should also be changed in all six copies of standalone.xml we have in the source tree.  (Feel free to implement a templating solution if that bothers you...)

PR #725, opened 2015-03-12
Fix servr url bug

huangp:
https://bugzilla.redhat.com/show_bug.cgi?id=1201080

### DIFF
--- a/zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java
+++ b/zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java
@@ -122,7 +122,7 @@ public class ApplicationConfiguration implements Serializable {
 
     private Optional<String> openIdProvider; // Cache the OpenId provider
 
-    private String serverPath;
+    private String defaultServerPath;
 
     @Create
     public void load() {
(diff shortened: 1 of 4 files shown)


zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java
```diff
@@ -239,20 +239,26 @@ public String getRegisterPath() {
     public String getServerPath() {
         String configuredValue = databaseBackedConfig.getServerHost();
         if (configuredValue != null) {
-            serverPath = configuredValue;
+            return configuredValue;
+        } else if (defaultServerPath != null) {
+            return defaultServerPath;
+        } else {
+            createDefaultServerPath();
+            return defaultServerPath;
```
davidmason: Don't need the `else` clauses since every code path has a return statement. Doesn't really make the code any more or less maintainable though.

davidmason: I find it a little more readable as follows, but is fine either way.

```
if (configuredValue != null) {
    return configuredValue;
}
if (defaultServerPath == null) {
    createDefaultServerPath();
}
return defaultServerPath;
```

davidmason:
:+1: looks good to me.

djansen-redhat:
:white_check_mark: Tested :+1:

PR #730, opened 2015-03-13
Move back to Ehcache for hibernate 2nd level cache.

carlosmunoz:
At least until a solution is found in terms of compatibility between Wildfly (hibernate 4.3, infinispan 6) and EAP (hibernate 4.2, infinispan 5).
Other data caches will still use infinispan.
The main reverted commit is: c513552187f78dcf7db96af66c6f2b6236c85a71
Parts of these commits were also reverted: 
0fa0a8ea34fe1a5c6463daa29f70f081748df0e4
f1a16afa548e0337514d5a093cf8d6938f08c234

### DIFF
--- a/docs/configuration/infinispan.md
+++ b/docs/configuration/infinispan.md
@@ -1,9 +1,6 @@
 # Infinispan for caching
 
-Zanata uses Infinispan to manage its internal data caches and search indexes. Configuration for these caches happens in JBoss' `standalone/configuration/standalone.xml`. There are two different caches that need to be configured for Zanata:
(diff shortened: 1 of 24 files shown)


seanf:
You should list exactly which commits were reverted.

seanf:
Looks okay, to me.  So is ehcache not a transactional cache, or is it just that we don't use that feature?

carlosmunoz:
I'm sure EhCache has a transactional mode, I'm just reverting to what we had before for simplicity.

PR #732, opened 2015-03-17
Remove Seam Text

seanf:
https://bugzilla.redhat.com/show_bug.cgi?id=1198433

- [x] Implement CommonMark rendering for home/project about pages
- [x] Implement CommonMark preview for edit home/project content
- [x] Test conversion from Seam Text to CommonMark by comparing rendered HTML
- [x] Implement Liquibase database migration from Seam Text to CommonMark
- [x] Test Liquibase database migration


zanata-war/src/main/java/org/zanata/action/HomePage.java
```diff
@@ -0,0 +1,67 @@
...
+@AutoCreate
+@Name("homePage")
+@Scope(ScopeType.APPLICATION)
+@Slf4j
+public class HomePage {
+
+    @In
+    private ApplicationConfiguration applicationConfiguration;
+
+    @In("commonMarkRenderer")
```
carlosmunoz: Any reason for using explicit names?

zanata-war/src/main/java/org/zanata/action/HomePage.java
```diff
@@ -0,0 +1,67 @@
...
+    private CommonMarkRenderer renderer;
+
+    private String html;
+
+    public String getHtml() {
```
carlosmunoz: Can we call this something more descriptive? Perhaps `mainContent` or `homeContent` (since that's exactly what it's fetching)

zanata-war/src/main/java/org/zanata/action/HomePage.java
seanf: I wanted the variable name to be shorter, without shortening the component
name.  The names aren't checked at compile time anyway, so I don't think we
lose anything.

The explicit name reference will go away with CDI anyway.

zanata-war/src/main/java/org/zanata/action/HomePage.java
seanf: Is `homePage.html` not enough?  That's how it will be used.

carlosmunoz:
Unit or funtional tests for this?
Also, documentation? (although it might need an overal project settings configuration page)

seanf:
I have now updated the previous functional tests for these pages to work
with the modifications I made, and modified the user guide's documentation
for the About Project page.

carlosmunoz:
Changes reviewed :+1:

seanf:
@carlosmunoz Do you want to look at the Liquibase code?

carlosmunoz:
:+1:

definite:

definite:

definite:
Tested with text:
    My code doesn't work:

    `for (int i=0; i<100; i--)
    {
        doSomething();
    }`

   Any ideas?

And it converted to 

    

    My code doesn't work:

    ```
    for (int i=0; i&lt;100; i--)
    {
         doSomething();
    }
    ```


    Any ideas?

While the preview (Setting->about) render correctly, but in the actual About page, the code squeeze into one line.

seanf:
@definite thanks.  Fortunately, I think it's just a problem with the rendering, not with the conversion.  That should make it easier for me to fix and for you to test the fix.

definite:
Also table rendered good in preview, but it squeezed as one line in About tab

    <table>
        <tr><td>First name:</td><td>Gavin</td></tr>
        <tr><td>Last name:</td><td>King</td></tr>
    </table>

seanf:
@definite I have whitelisted some more HTML tags, so please grab the latest build.

seanf:
Squash-merged as c970d50b3bd884203446bd44214ee3ffdfb4b33f.

seanf:
Release notes for this are in https://github.com/zanata/zanata-server/commit/ee7b454f4ed829cfbd48e90eda142e49434f4d9b .

PR #733, opened 2015-03-17
rhbz1197902 - clear hibernate session after each translation batch save

huangp:
https://bugzilla.redhat.com/show_bug.cgi?id=1197902

- [ ] Need to backport to release branch

### DIFF
--- a/zanata-war/src/main/java/org/zanata/async/AsyncTaskManager.java
+++ b/zanata-war/src/main/java/org/zanata/async/AsyncTaskManager.java
@@ -101,7 +101,7 @@ public class AsyncTaskManager {
                     taskFuture.set(returnValue);
                 } catch (Throwable t) {
                     taskFuture.setException(t);
-                    log.warn(
+                    log.error(
                             "Exception when executing an asynchronous task.", t);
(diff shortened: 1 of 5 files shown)


huangp:

zanata-war/src/main/java/org/zanata/async/AsyncTaskManager.java
```diff
@@ -101,7 +101,7 @@ public void execute() {
                     taskFuture.set(returnValue);
                 } catch (Throwable t) {
                     taskFuture.setException(t);
-                    log.warn(
+                    log.error(
                             "Exception when executing an asynchronous task.", t);
```
seanf: This change is okay, but it really should have had its own commit with its own commit message.

seanf: Perhaps we should log the stacktrace as WARN, and log the exception.toString as ERROR.

djansen-redhat:
:white_check_mark: Tested.
The difference in time is quite significant, generally between ten and twenty times faster. :+1:

PR #734, opened 2015-03-17
Hash on release

davidmason:
This is a back-port of https://github.com/zanata/zanata-server/pull/724 to allow a point-release with just this change, as noted in https://bugzilla.redhat.com/show_bug.cgi?id=1194543#c7


djansen-redhat:
:white_check_mark: Tested, smoke and sanity testing.  Already verified in master.

PR #735, opened 2015-03-17
Increase timeout on server startup in all standalone.xml files.

davidmason:
Changes the timeout to 1000 seconds (16min 40s) from default 300s (5min)
to prevent timeout when liquibase changesets take longer than 5 minutes
to run.

### DIFF
--- a/functional-test/src/test/resources/conf/standalone.xml
+++ b/functional-test/src/test/resources/conf/standalone.xml
@@ -28,6 +28,7 @@
   </extensions>
   <system-properties>
     <property name="zanata.countAjax" value="true" />
+    <property name="jboss.as.management.blocking.timeout" value="1000"/>
   </system-properties>
   <management>
     <security-realms>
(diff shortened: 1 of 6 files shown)


seanf:
:+1:

djansen-redhat:
:see_no_evil: I've already tested that this works, and is a configuration change.

djansen-redhat:
Due to #734 this will need to be in release branch also

carlosmunoz:
:memo: Lets add a note about this in the docs.
It isn't a mandatory thing, so we should mention what it's for and how to set it up.

davidmason:
@carlosmunoz it is mentioned in the release notes, but I agree it would be worth a mention in the server config documentation. I will add that in a separate branch since this change needs to be applied to release.

davidmason:
Rebased onto master, see https://github.com/zanata/zanata-server/pull/739

PR #736, opened 2015-03-19
Indicate active, readonly, and archived versions

davidmason:
https://bugzilla.redhat.com/show_bug.cgi?id=1202670

Moved most changes to https://github.com/zanata/zanata-server/pull/738
Closing until https://github.com/zanata/zanata-server/pull/738 is merged to release and master, then I will re-open this with the remaining change that could not be applied to the release branch (because the file does not exist in that branch).

### DIFF
--- a/zanata-war/src/main/java/org/zanata/webtrans/client/resources/WebTransMessages.java
+++ b/zanata-war/src/main/java/org/zanata/webtrans/client/resources/WebTransMessages.java
@@ -381,8 +381,11 @@ public interface WebTransMessages extends Messages {
     @DefaultMessage("Page size")
     String pageSize();
 
-    @DefaultMessage("Read only")
-    String readOnly();
+    @DefaultMessage("This project-version is readonly. It cannot be edited.")
+    String readOnlyTooltip();
+
+    @DefaultMessage("This project-version is archived. It cannot be edited.")
+    String obsoleteTooltip();
 
     @DefaultMessage("Advanced user configuration")
     String otherConfiguration();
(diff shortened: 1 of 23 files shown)


zanata-war/src/main/java/org/zanata/webtrans/client/resources/WebTransMessages.java
```diff
@@ -381,8 +381,11 @@ String undoUnsuccessful(@PluralCount int unsuccessfulCount,
     @DefaultMessage("Page size")
     String pageSize();
 
-    @DefaultMessage("Read only")
-    String readOnly();
+    @DefaultMessage("This project-version is readonly. It cannot be edited.")
+    String readOnlyTooltip();
+
+    @DefaultMessage("This project-version is obsolete. It cannot be edited.")
```
carlosmunoz: Should we unify terms and call this "Archived" instead? Not entirely sure which one we are going by these days.

carlosmunoz:
:+1: (Aside from the question on terminology)

zanata-war/src/main/java/org/zanata/webtrans/client/resources/WebTransMessages.java
davidmason: Yes. It is `obsolete` in all the code, so I had that in mind. I'll change the string to "archived"

seanf: Except that Luke's been talking about renaming "read-only" as archived
(which makes a lot of sense if you don't worry about confusing users who
know our previous meaning for archived).

davidmason:
Closing for now since most changes are now in https://github.com/zanata/zanata-server/pull/738 to target release branch.

PR #737, opened 2015-03-19
Allow anonymous REST request on READ action

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1172618

See https://github.com/zanata/zanata-client/pull/54 for client side changes.


zanata-war/src/main/java/org/zanata/limits/RateLimitManager.java
```diff
@@ -125,7 +125,7 @@ public String getFieldValueAsString(String fieldName) {
                 });
     }
 
-    public RestCallLimiter getLimiter(final String apiKey) {
+    public RestCallLimiter getLimiter(final String key) {
```
seanf: Any reason for the rename?

seanf: Oh, it's because it could be an api key, or it could be an ip address?  Lumping them both into a String could be a bit error-prone.  Perhaps `Either` could help? http://stackoverflow.com/a/16579332/14379  And the JavaDoc should say what's going on.

zanata-war/src/main/java/org/zanata/limits/RateLimitingProcessor.java
```diff
@@ -36,6 +36,16 @@ public void processApiKey(String apiKey, HttpResponse response,
...
+        Runnable taskToRun) throws IOException {
+        process(username, response, taskToRun);
+    }
+
+    public void processClientIp(String ip, HttpResponse response,
+        Runnable taskToRun) throws IOException {
+        process(ip, response, taskToRun);
+    }
+
     private void process(String key, HttpResponse response, Runnable taskToRun)
```
seanf: "process" is something we do to the `Runnable` task, so processUsername and processClientIp don't quite make sense.  (I know, processUsername already existed.)

What about `processForUser` and `processForAnonymousIP`?

zanata-war/src/main/java/org/zanata/rest/RestLimitingSynchronousDispatcher.java
```diff
@@ -45,18 +71,33 @@ public RestLimitingSynchronousDispatcher(
...
 
+    HttpServletRequest getServletRequest() {
+        return ServletContexts.instance().getRequest();
+    }
+
     @Override
     public void invoke(final HttpRequest request, final HttpResponse response) {
 
+        //This is for REST request from browser where user is logged in.
         HAccount authenticatedUser = getAuthenticatedUser();
```
seanf: > This is for REST request from browser where user is logged in.
What is?  Could `authenticatedUser` have been authenticated by api key at this point?

seanf: So at this point `authenticatedUser` is only non-null if the request came from a browser?  Could we say that?

zanata-war/src/main/java/org/zanata/rest/RestLimitingSynchronousDispatcher.java
```diff
@@ -45,18 +71,33 @@ public RestLimitingSynchronousDispatcher(
...
-            // we are not validating api key but will rate limit any api key
-            if (authenticatedUser == null && Strings.isNullOrEmpty(apiKey)) {
-                response.sendError(
-                        Response.Status.UNAUTHORIZED.getStatusCode(),
-                        API_KEY_ABSENCE_WARNING);
+            if (authenticatedUser == null
+                    && Strings.isNullOrEmpty(apiKey)
+                    && !SecurityFunctions.canAccessRestPath(authenticatedUser,
+                            request.getHttpMethod(),
+                            request.getPreprocessedPath())) {
```
seanf: Every incoming request should trigger a call to `SecurityFunctions.canAccessRestPath`, not just the anonymous ones.

zanata-war/src/main/java/org/zanata/rest/RestLimitingSynchronousDispatcher.java
```diff
@@ -45,18 +71,33 @@ public RestLimitingSynchronousDispatcher(
...
+
+                /**
+                 * Not using response.sendError because that will allow JBoss to
+                 * intercept and generate html page with the message. We want to
+                 * return a message string.
```
seanf: What about `response.sendError(int status, String message)`?

zanata-war/src/main/java/org/zanata/security/SecurityFunctions.java
```diff
@@ -389,4 +390,32 @@ private static final ZanataIdentity getIdentity() {
...
+
+    /*****************************************************************************************
+     * HTTP request rules
+     ******************************************************************************************/
+
+    /**
+     * Check if user can access to REST URL with httpMethod.
+     * If request is from anonymous user(account == null),
+     * only 'Read' action are allowed. Otherwise, role base check will should be
+     * perform in related RestService class.
```
seanf: That should be "**Additionally**, role-based check will be performed in the REST service class".  At least that's the way it should be working.

First we should check whether you can talk to our API at all, then we should check whether you can perform the specific API action you requested.

zanata-war/src/main/java/org/zanata/security/SecurityFunctions.java
```diff
@@ -389,4 +390,32 @@ private static final ZanataIdentity getIdentity() {
...
+     *
+     * This rule apply to all REST endpoint.
+     *
+     * @param account - Authenticated account
+     * @param httpMethod - {@link javax.ws.rs.HttpMethod}
+     * @param restServicePath - service path of rest request.
+     *                        See annotation @Path in REST service class.
+     */
+    public static final boolean canAccessRestPath(HAccount account,
```
seanf: `@Nullable HAccount account` would be nice.

zanata-war/src/main/java/org/zanata/util/HttpUtil.java
```diff
@@ -0,0 +1,106 @@
...
+     *
+     * This will also check for the possibility of client behind proxy
+     * before returning default remote address in request.
+     *
+     * NOTE: Not all proxy server include client ip information in http header
+     * and different proxy MIGHT use different http header for such information.
+     * Default remote address in request will be returned if client information
+     * is not found in header.
+     *
+     * see http://stackoverflow.com.80bola.com/questions/4678797/how-do-i-get-the-remote-address-of-a-client-in-servlet
```
seanf: What the heck is `80bola.com` doing in that stackoverflow URL?  A link to stackoverflow.com might stay valid longer...

zanata-war/src/main/java/org/zanata/util/HttpUtil.java
```diff
@@ -0,0 +1,106 @@
...
+            if(!isIpUnknown(ip)) {
+                return ip;
+            }
+        }
+        return request.getRemoteAddr();
+    }
+
+    private static boolean isIpUnknown(String ip) {
+        return StringUtils.isEmpty(ip) || StringUtils.equalsIgnoreCase(ip,
+                "unknown");
```
seanf: I think perhaps we should include `localhost` and `127.0.0.1` here too.

seanf: > I think perhaps we should include `localhost` and `127.0.0.1` here too.

In case we messed up the proxy header handling, that is.  Of course, during development and testing the connections often *are* from `localhost`.

zanata-war/src/main/java/org/zanata/util/HttpUtil.java
```diff
@@ -0,0 +1,106 @@
...
+
+    public static final String X_FORWARDED_FOR = "X-Forwarded-For";
+    public static final String PROXY_CLIENT_IP = "Proxy-Client-IP";
+    public static final String WL_Proxy_Client_IP = "WL-Proxy-Client-IP";
+    public static final String HTTP_CLIENT_IP = "HTTP_CLIENT_IP";
+    public static final String HTTP_X_FORWARDED_FOR = "HTTP_X_FORWARDED_FOR";
+
+    private static final LinkedList<String> PROXY_HEADERS = Lists
+            .newLinkedList((Arrays.asList(X_FORWARDED_FOR, PROXY_CLIENT_IP,
+                    WL_Proxy_Client_IP, HTTP_CLIENT_IP, HTTP_X_FORWARDED_FOR)));
```
seanf: If you only use these strings once, I would just put them in here as literals.  Making them public constants increases the size of this class's API.  It's a bit of a random collection of header names, any attribution other than stackoverflow?

http://distinctplace.com/infrastructure/2014/04/23/story-behind-x-forwarded-for-and-x-real-ip-headers/

I think X-Forwarded-For can actually be a list of IP addresses, starting with the client's purported IP address (possibly forged, or inside a LAN with NAT) and hopefully ending with the real IP address which actually connected to our HTTP proxy.  Unless of course Zanata is running without Apache/nginx/haproxy in front of it, in which case we shouldn't be looking at any proxy headers.

Since we're using this for rate limiting, we need to be careful about trusting the headers.  What's to stop someone putting a false header into their requests?  Or perhaps they have a genuine HTTP proxy on their office LAN, the IP addresses of which are irrelevant to us.

Perhaps we should only accept a single optional header name, which should be specified as a system property by the system admin.  `X-Forwarded-For` seems to be pretty standard, so we should make sure we can extract the last IP in the list, as mentioned in the linked page above.

zanata-war/src/main/java/org/zanata/exception/InvalidApiKeyException.java
```diff
@@ -0,0 +1,43 @@
+package org.zanata.exception;
+
+/**
+ * Exception for invalid API key.
+ *
+ * @author Alex Eng <a href="mailto:aeng@redhat.com">aeng@redhat.com</a>
+ */
+public class InvalidApiKeyException extends Exception {
```
seanf: Since we never create an instance of this exception, it might be better to call this class `InvalidApiKeyUtil` and put it in the `org.zanata.rest` package.  

I'm sure there's a better way of handing it than yet another util class (which is what this exception really is), but I can't think of it right now.  (It probably involves code generation to enable type-safe formatting for messages.)

zanata-war/src/main/java/org/zanata/rest/RestLimitingSynchronousDispatcher.java
aeng: /**
                 * Not using response.sendError because that will allow JBoss to
                 * intercept and generate html page with the message. We want to
                 * return a message string.
                 */

zanata-war/src/main/java/org/zanata/rest/RestLimitingSynchronousDispatcher.java
aeng: @seanf , YES. `authenticatedUser` is only non-null if request came from browser.

zanata-war/src/main/java/org/zanata/limits/RateLimitManager.java
aeng: Can be username(if no api key generate with user), apiKey or anonymous IP address.

zanata-war/src/main/java/org/zanata/rest/RestLimitingSynchronousDispatcher.java
seanf: @aeng are you saying that when you call sendError with a message (the overload with 2 params), it ignores the message?

seanf: Oh.  It just includes the specified message string in its HTML template?

seanf: To be clearer (assuming I've understood correctly), how about this?

	/**
	* Not using response.sendError because the app server will generate
	* an HTML page which includes the message. We want to return
	* the message string as is.
	*/

aeng: That sounds better.

zanata-war/src/main/java/org/zanata/limits/RateLimitManager.java
seanf: > Can be 
1) username(if no api key generate for that user)
2) api key
3) anonymous IP address.

Of course 1) case will not happen as the request will already be rejected in ZanataRestSecurityInterceptor

I do think it's important to distinguish between these cases.  What about a simple immutable value class with factory methods which take either username, API key or IP address?

Also, won't case 1 happen when we invoke REST services from the browser, if the user has no api key?

djansen-redhat:
- [ ] :no_entry: Release note

zanata-war/src/main/java/org/zanata/limits/RateLimitManager.java
seanf: > not sure what you meant by "immutable" value class for the key

I was thinking of something like this:

```java
    @Getter
    class RateLimiterKey {
    	private final String username;
    	private final String apiKey;
    	private final String ipAddress;
    	// and three different static factory methods...
    }
```
Or perhaps an enum would be better:
```java
    enum RateLimiterKeyType {
    	USERNAME, API_KEY, IP_ADDRESS
    }

    @Getter
    class RateLimiterKey {
    	private final RateLimiterKeyType keyType;
    	private final String username;
    	private final String apiKey;
    	private final String ipAddress;
    	// toString()
    	// and three different static factory methods...
    }
```
However, that's all getting to be a bit much for such a simple thing.  We don't store these values long term, so we can always change the implementation later if need be.

- [ ] Perhaps we should just call `key` something like `token` instead, just to make it clear that it may not be an API key.

seanf:
- [ ] The release notes ought to say something about the proxy header, and the fact that Zanata interprets the header as as a comma-separated list (X-Forwarded-For style), using the last value.

aeng:
Yup, I will create the release note as separate PR targeting integration/master branch.

seanf:
- [ ] When the rate limiter starts rejecting requests, we need to modify the error message to make it clear that it could be caused by too many requests from the same login, API key, or IP address.  In the past, there was only one way of hitting that error (too many requests with same API key), so the message probably needs updating.

seanf:
Everything else looks fine; just need to find and tweak that error message.

seanf:
It looks like the error message (in `RateLimitingProcessor`) is "Too many concurrent requests for this user (maximum is %d)".  

I would suggest something like "Too many concurrent requests for client '%s' (maximum is %d)" where %s is the username, or the IP address, or the literal string "API key" (because we shouldn't put the user's API key in the message).

seanf:
If you merge the current version of `release`, there is now a `docs/release-notes.md` file you could edit.

seanf:
@aeng so does this currently make it possible for anonymous REST users to fetch stats, to pull source/target documents, but not to export TMX?

aeng:
@seanf need to check the TMX export method.. if its `get` type, then it allows anonymous.

aeng:
@seanf checked TranslationMemoryResource. Export TM (GET) is opened for anonymous

seanf:
@aeng now that we're opening up GET to everybody, we need to update the security rules so that "get TMX" still requires authentication.

seanf:
@aeng Looking good, but don't forget about the release notes and the improved error message for the rate limiter.

aeng:
@seanf added the release note and updated the error message.

zanata-war/src/main/java/org/zanata/limits/RateLimitingProcessor.java
```diff
@@ -48,16 +58,20 @@ private void process(String key, HttpResponse response, Runnable taskToRun)
...
+            if(key.getType().equals(RateLimiterToken.TYPE.API_KEY)) {
+                clientIdentifier = "API key";
+            } else  {
+                clientIdentifier = key.getValue();
+            }
+
             String errorMessage =
                     String.format(
-                            "Too many concurrent requests for this user (maximum is %d)",
+                            "Too many concurrent requests for client '%s' (maximum is %d)",
```
seanf: Probably best not to put "API key" in quotes here, or it will look like a username: `Too many concurrent requests for client 'API key'`.  Since you have access to the actual Token type here, you could customise the message properly: `username '%s`", `IP address '%s'` or `API key`.

seanf:
@aeng :+1:  All looks good now.  But just for future reference, please note that my previous :+1: was for bd0d3e8, not necessarily the whole PR.  (Not sure if github makes that clear when sending email notifications.)

definite:
Is it normal that  
    curl -v -X GET -H "Content-Type:application/json" ${ZANATA_URL}rest/projects/p/${PRJ}/iterations/i/${VERSION}/r/${DOC}/translations/${LANG} 
returns XML instead of JSON?

definite:
The failed to return JSON bug should be address in https://bugzilla.redhat.com/show_bug.cgi?id=1205473

definite:
Test cases:
-[ ] REST api
  -[ ] AccountResource
    -[ ] /accounts/u/{username}
      -[ ] 200 user authenticated 
      -[ ] 401 failed authenticated
      -[ ] 401 anonymous
  -[ ] AsynchronousProcessResource
    -[ ] /async/projects/p/{projectSlug}/iterations/i/{iterationSlug}/r/accounts/u/{username}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 401 anonymous
  -[ ] CopyTransResource
    -[ ] /copytrans/proj/{projectSlug}/iter/{iterationSlug}/doc/{docId}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 401 No authenticated
  -[ ] FileResource
    -[ ] /file/accepted_type
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
    -[ ] /file/download/{downloadId}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
    -[ ] /file/source/{projectSlug}/{iterationSlug}/{fileType}  
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
  -[ ] GlossaryResource
    -[ ] /glossary
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 401 No authenticated
    -[ ] /glossary/{locale}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 401 No authenticated
  -[ ] ProjectIterationLocalesResource
    -[ ] /projects/p/{projectSlug}/iterations/i/{iterationSlug}/locales
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
  -[ ] ProjectIterationResource
    -[ ] /projects/p/{projectSlug}/iterations/i/{iterationSlug}/locales
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
    -[ ] /projects/p/{projectSlug}/iterations/i/{iterationSlug}/config
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
  -[ ] ProjectLocalesResource
    -[ ] /projects/p/{projectSlug}/locales
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
  -[ ] ProjectResource
    -[ ] /projects/p/{projectSlug}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
  -[ ] ProjectsResource
    -[ ] /projects
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
  -[ ] SourceDocResource
    -[ ] /projects/p/{projectSlug}/iterations/i/{iterationSlug}/r
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
    -[ ] /projects/p/{projectSlug}/iterations/i/{iterationSlug}/r/{id}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
    -[ ] /projects/p/{projectSlug}/iterations/i/{iterationSlug}/r/{id}/meta
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
  -[ ] StatisticsResource
    -[ ] /stats/proj/{projectSlug}/iter/{iterationSlug}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
    -[ ] /stats/proj/{projectSlug}/iter/{iterationSlug}/doc/{docId}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
    -[ ] /stats/project/{projectSlug}/version/{versionSlug}/contributor/{username}/{dateRange}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated  
  -[ ] TranslatedDocResource
    -[ ] /projects/p/{projectSlug}/iterations/i/{iterationSlug}/r/{id}/translations/{locale}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated  
  -[ ] TranslationMemoryResource
    -[ ] /tm/{slug}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 401 No authenticated  
      -[ ] TranslationMemoryResource
    -[ ] /tm/projects/{projectSlug}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 401 No authenticated
    -[ ] /tm/projects/{projectSlug}/iterations/{iterationSlug}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 401 No authenticated
  -[ ] VersionResource
    -[ ] /tm/projects/{projectSlug}/iterations/{iterationSlug}
      -[ ] 200 user authenticated
      -[ ] 401 failed authenticated
      -[ ] 200 No authenticated
-[ ] Rate limiter
Note that most test is test against GET methods, and the number before the test items are expected HTTP status code.

definite:
with /tm/projects/{projectSlug}

HTTP 500 is returned

server.log:

16:26:28,173 ERROR [org.zanata.rest.RestLimitingSynchronousDispatcher] (http-/0.0.0.0:8080-1) Failed to process REST request: javax.el.ELException: Function 's:hasPermission' not found

definite:
:+1:

aeng:
Merged into release. https://github.com/zanata/zanata-server/commit/8ccbfa173e78e44d0ebc3d066c0ddbc7d1c6e894

PR #744, opened 2015-03-24
Update docs

aeng:
Moved command hook documentation into readthedocs.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,31 +1,22 @@
 Zanata Server
 =============
-
-TBD
-
-Features
---------
-
-- TBD
+Please visit [About|Zanata](http://zanata.org/about/)
 
 Installation
 ------------
 
-<!-- For installation instructions please see the [Installation Section](installation) -->
+[Installation section](configuration/installation.md)
 
 Contribute
 ----------
 
-- Issue Tracker: http://bugzilla.redhat.com
-- Source Code: github.com/zanata/zanata-server
+- [Issue Tracker](http://bugzilla.redhat.com/buglist.cgi?product=Zanata)
+- [Source Code on GitHub](http://github.com/zanata)
 
 Support
 -------
-
-If you are having issues, please let us know.
-We have a mailing list located at: zanata-users@redhat.com
+For help and support options, see [Help|Zanata](http://zanata.org/help/)
 
 License
 -------
-
-<!-- The project is licensed under the BSD license. -->
+Zanata is Free software, licensed under the [LGPL](http://www.gnu.org/licenses/lgpl-2.1.html).
--- a/docs/user-guide/client-configuration.md
+++ b/docs/user-guide/client-configuration.md
@@ -1,5 +1,3 @@
-# Common configuration files for Flies clients
-
 # Introduction
 
 In general, Zanata clients should get their configuration from user config, project config and command line args (or similar).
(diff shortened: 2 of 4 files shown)


docs/index.md
```diff
@@ -1,31 +1,23 @@
 Zanata Server
 =============
+Please visit [our about page](http://zanata.org/about/)
```
davidmason: ... [About | Zanata] ...

docs/index.md
```diff
@@ -1,31 +1,23 @@
...
-Features
---------
-
-- TBD
 
 Installation
 ------------
 
-<!-- For installation instructions please see the [Installation Section](installation) -->
+[Installation Section](installation)
```
davidmason: Does this actually link to something?

docs/index.md
```diff
@@ -1,31 +1,23 @@
...
 
 Contribute
 ----------
 
-- Issue Tracker: http://bugzilla.redhat.com
-- Source Code: github.com/zanata/zanata-server
+- Issue Tracker: [Bugzilla](http://bugzilla.redhat.com/buglist.cgi?product=Zanata)
+- Source Code: [GitHub](http://github.com/zanata)
```
davidmason:
```
- [Issue Tracker](...
- [Source Code on GitHub](...
```

docs/index.md
```diff
@@ -1,31 +1,23 @@
...
 
 Support
 -------
-
-If you are having issues, please let us know.
-We have a mailing list located at: zanata-users@redhat.com
+Click [here](http://zanata.org/help/) for help and support
```
davidmason: We should avoid having link text like "here" since it causes big problems for accessibility. We can do something like:

```
For help and support options, see [Help | Zanata](...).
```

docs/index.md
```diff
@@ -1,31 +1,23 @@
...
 
 License
 -------
-
-<!-- The project is licensed under the BSD license. -->
\ No newline at end of file
+Click [here](https://github.com/zanata/zanata-server/blob/master/README.md) for our license information.
```
davidmason: That is the readme, not the license file. We should at least copy the section from the readme so people don't have to click and search for the info:

```
Zanata is Free software, licensed under the [LGPL](http://www.gnu.org/licenses/lgpl-2.1.html).
```

docs/user-guide/client-configuration.md
```diff
@@ -86,4 +84,43 @@ URLs in project configuration should be matched against the [servers] defined in
 
 The client should also provide the ability to override user/project configuration values, perhaps with command line arguments or GUI options, etc.
 
-See ZanataMavenIntegration for the Maven configuration properties and example pom.xml.
\ No newline at end of file
+## Client Command Hook
```
davidmason: This section is big enough to be its own page.

docs/index.md
```diff
@@ -1,31 +1,23 @@
 Zanata Server
 =============
+Please visit [our about page](http://zanata.org/about/)
 
-TBD
-
-Features
---------
```
davidmason: This is probably still a good section to have. We should leave an xml comment reminding us to add it, at least.

aeng: The feature section should be in http://zanata.org/features/. 
I think its best if we can focus on zanata-admin(self host) documentation in readthedocs and maintainer/users guide in zanata.org/help.

carlosmunoz: Maintainer's and users' guide should also be on readthedocs IMO. Anything that we may want to keep versioned should really go there. That way we can show different documentation for the most recent versions of Zanata.

davidmason:
:+1:

PR #746, opened 2015-03-25
Fix language sorting in members size

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1205468

### DIFF
--- a/zanata-war/src/main/java/org/zanata/action/SupportedLanguageAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/SupportedLanguageAction.java
@@ -4,14 +4,17 @@ import java.io.Serializable;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
 
 import com.google.common.collect.Lists;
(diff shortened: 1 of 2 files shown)


djansen-redhat:
:white_check_mark: Tested, looks good

zanata-war/src/main/java/org/zanata/action/SupportedLanguageAction.java
```diff
@@ -41,10 +46,19 @@
     protected List<HLocale> fetchAll() {
         if (supportedLanguages == null) {
             supportedLanguages = localeServiceImpl.getSupportedLocales();
```
davidmason: I have been seeing this pattern a lot lately. We should probably wrap it up in something (not in this PR though).

```
class LazyField<T> {
    private T value = null;

    /**
     * Override this with code to look up the value.
     */
    protected abstract T fetchValue();

    public T get() {
        if (value == null) {
            value = fetchValue();
        }
        return value;
    }
}
```

davidmason:
Code looks fine :+1:

PR #747, opened 2015-03-25
upgrade zanata-common to 3.6.0

huangp:
https://bugzilla.redhat.com/show_bug.cgi?id=1029734


djansen-redhat:
- [ ] :x: Release note

huangp:
Added release note

aeng:
:+1:

djansen-redhat:
Just to clarify - this change makes it "not break", rather than successfully parse, when the header is invalid?

huangp:
Yes and no. Before this change if you upload a po file with "\n" or "abc" in header entry, it will throw exception and terminate. Now it will log a warning and ignore the offending entry but continue parsing. To offer better feedback to user is out of scope for this one.

djansen-redhat:
:white_check_mark: Tested

PR #753, opened 2015-03-27
Log as DEBUG when permission is denied (not WARN)

seanf:
We get `hasPermission` denials all the time, so they are just noise.


djansen-redhat:
Do we do this "if(loglevel) log.level(msg)" *everywhere* we log!?

djansen-redhat:
:+1:

seanf:
No, I generally only include the preliminary check if the log parameters
require method calls *and* the method with logging is high volume.

djansen-redhat:
:white_check_mark: Tested

PR #754, opened 2015-03-30
Zanata assets jar

aeng:
Use packaged zanata-assets with zanata-server. See https://github.com/zanata/zanata-assets/blob/master/pom.xml

https://bugzilla.redhat.com/show_bug.cgi?id=1207423


zanata-war/src/main/webapp/WEB-INF/urlrewrite.xml
```diff
@@ -293,6 +293,15 @@
     <to last="true">/webtrans/Application.seam$1</to>
   </rule>
 
+  <!-- zanata-assets resource -->
+  <!-- JSF serves zanata-assets with suffix of .seam,
+      this is to make sure any reference to zanata-assets without .seam can access to the resource.
+      e.g. /assets/style.css are redirect to /assets/style.css.seam -->
```
seanf: How about this?

    eg a request for /assets/style.css will be redirected to /assets/style.css.seam

zanata-war/src/main/webapp/WEB-INF/urlrewrite.xml
```diff
@@ -293,6 +293,15 @@
...
+  <rule>
+    <from casesensitive="true">^/javax.faces.resource/assets/(.*)(?&lt;!.seam)$</from>
+    <to last="true">/javax.faces.resource/assets/$1.seam</to>
```
seanf: If we make this a little more general (not just `.../assets/`) we could use it for the pre-packaged WebJars from webjars.org too, like the ones for jQuery and Angular (or commonmark.js).  

I think we could just map from /webjars/* to `/javax.faces.resource/webjars/$1.seam` and we would also need to modify the zanata-assets webjar to be consistent with webjars.org, ie `META-INF/resources/webjars/zanata-assets/{version}`

aeng: @seanf our zanata-assets.jar is not generated from webjar lib. It using pom.xml I created in zanata-assets repo.
Should we even have the `webjar` reference in our url and package structure?

aeng: And also with the `/{version}` as part of the structure, I think it will be harder to be referenced from html page given that we need to pass the version number around.

seanf: > @seanf our zanata-assets.jar is not generated from webjar lib. It using pom.xml I created in zanata-assets repo.
Should we even have the webjar reference in our url and package structure?

Well, perhaps it is a bad idea to use `/webjar` since the zanata-assets artifact isn't currently in the maven group `org.webjar`.  I was hoping we could make it transparent, in case we switch to bower packaging and let webjars.org package zanata-assets for us.  But perhaps that isn't a good enough reason to borrow the webjar path ahead of time.

Okay then, don't use `/webjar` for it, but I do think we need something generic so that we don't need multiple urlrewrite rules when we package other things this way.

> And also with the /{version} as part of the structure, I think it will be harder to be referenced from html page given that we need to pass the version number around.

We could use some code to help us build the correct URLs.  If we don't put the version into the URL, it may be difficult to get the caching strategy right.

But we can tackle the versioning problem another time, so feel free to leave it out for now.

seanf: Okay, how about we simply map from `/jars/(.*)` (no need for the negative lookahead in this case) to `/javax.faces.resource/jars/$1.seam` and modify the zanata-assets webjar to be  `META-INF/resources/jars/zanata-assets/*`.  (With the versioning as an optional future extension.)

zanata-war/src/main/webapp/WEB-INF/urlrewrite.xml
seanf: or (updated)

> eg a request for /jars/assets/style.css will be redirected to /javax.faces.resource/assets/style.css.seam

seanf:
@aeng Or perhaps `/zanatajars/` would be better than `/jars/`?  Or even `/zanatawebjars/`, but I think that's getting a bit too long.

zanata-war/src/main/webapp/WEB-INF/urlrewrite.xml
aeng: So for the jar structure, it would be META-INF/resources/zanatajars/zanata-assets/

And the url mapping would be /zanatajars/(*) to /javax.faces.resource/zanatajars/$1.seam

Am I correct?

seanf: > So for the jar structure, it would be META-INF/resources/zanatajars/zanata-assets/

Yes, though `META-INF/resources/zanatajars/assets/` might do.

> And the url mapping would be /zanatajars/(*) to /javax.faces.resource/zanatajars/$1.seam

Well, `/zanatajars/(.*)` to be precise, but yes.

zanata-war/src/main/java/org/zanata/WebAssetsConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    private final static String DEFAULT_WEB_ASSETS_URL = String.format("%s%s/%s/%s",
+            FacesContext.getCurrentInstance().getExternalContext()
+                    .getRequestContextPath(),
+            ResourceHandler.RESOURCE_IDENTIFIER, "jars", "assets");
+
+    private String webAssetsUrlBase;
+
+    private String getWebAssetsUrl(String resource) {
+        return String.format("%s/%s", getWebAssetsUrlBase(), resource);
```
seanf: Why not simply concatenate the strings?

zanata-war/src/main/java/org/zanata/WebAssetsConfiguration.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    private String getWebAssetsUrlBase() {
+        if (StringUtils.isBlank(webAssetsUrlBase)) {
+            String assetsProperty = System.getProperty(ASSETS_PROPERTY_KEY);
+
+            webAssetsUrlBase =
+                    StringUtils.isEmpty(assetsProperty) ? DEFAULT_WEB_ASSETS_URL
+                            : assetsProperty;
+        }
+        return webAssetsUrlBase;
```
seanf: We should calculate this value just once, when the bean is initialised.

aeng: I'm sure @Create annotation cannot be place in constructor.

seanf:
:+1: Looks good, but it might be a good idea to run all the functional tests, just in case of a typo in the XHTML.

aeng:
full test: http://zanata-build.lab.eng.bne.redhat.com/job/zanata-server-github-pull-requests/3260/

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/bae3c77b6fa4f12e5233ff3190e76a2865c5be36

PR #755, opened 2015-03-30
Restore constraints on glossary tables for H2

seanf:
The constraints which were dropped by changeset db.changelog-1.6a.xml:1:aeng@redhat.com were never reinstated, so this changeset puts them back.  (Discovered while investigating https://bugzilla.redhat.com/show_bug.cgi?id=1048293)

Note: this change is untested, and to test it will require running Zanata server against a persistent H2 database.  That isn't something we normally do these days, but I want to record this for completeness, and in case we ever want to run Zanata against H2 in future (eg for demos or quickstarts).

TODO (some time): Add a basic test which deploys Zanata on H2, thus validating that our liquibase changesets at least work on an empty H2 database.  Trying to be database-agnostic might help with future migration, eg https://bugzilla.redhat.com/show_bug.cgi?id=1205945

### DIFF
--- a/zanata-war/src/main/resources/db/changelogs/db.changelog-3.7.xml
+++ b/zanata-war/src/main/resources/db/changelogs/db.changelog-3.7.xml
@@ -27,6 +27,16 @@
     </addColumn>
   </changeSet>
 
+  <!-- Restore H2 constraints removed by changeset db.changelog-1.6a.xml:1:aeng@redhat.com -->
+  <changeSet author="sflaniga@redhat.com" id="1.6a:1:aeng@redhat.com:fix" dbms="h2">
+    <addForeignKeyConstraint baseTableName="HGlossaryTerm"
+      baseColumnNames="glossaryEntryId" constraintName="UKglossaryterm_glossary_entry_id"
+      referencedTableName="HGlossaryEntry" referencedColumnNames="id" />
+    <addPrimaryKey tableName="HGlossaryEntry" columnNames="id" />
+    <addPrimaryKey tableName="HGlossaryTerm" columnNames="id" />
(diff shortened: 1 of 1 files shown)


seanf:

davidmason:
:+1:

seanf:
Rather than wait until I find time to set up sanity tests for liquibase on H2, I'd like to merge this change now.  Without testing, it's possible that it won't quite work on H2, but the same goes for most of the other changesets.  I still think we should set it up before trying to switch to PostgreSQL or similar.

Any feelings on this, @davidmason @aeng @huangp @djansen-redhat  ?

aeng:
I think don't merge in if there's no assurance of this changeset will work. If its part of preparation for PostgreSQL, then it should be sitting in that branch.

seanf:
@aeng the thing is, I don't want us to forget that H2 is missing some constraints, and I think this PR is better than a bugzilla to keep track of it.

We do have some assurance - it's the same way we created those constraints in other changesets, and this branch wouldn't pass the functional tests unless liquibase was at least capable of parsing the changeset (even if it didn't run it against H2).  

If we leave it too long, someone will change the structure of the glossary tables, bitrot will set in and invalidate this changeset.

I don't want to create a long-running branch for PostgreSQL if I can avoid it.  Long-running branches always lead to pain.  Best avoided where possible.

aeng:
@seanf fair point. :+1:

davidmason:
:+1:  I have no problem with merging this if all our tests pass. At worst we are breaking something that we never use, but for all we know it could be fixing something broken that we never use - it is pretty much moot.

seanf:
Okay, thanks guys.

PR #756, opened 2015-03-30
rhbz1048293 Fix liquibase changeset to be compatible with mysql/mariadb version > 5.5

seanf:
https://bugzilla.redhat.com/show_bug.cgi?id=1048293

### DIFF
--- a/functional-test/pom.xml
+++ b/functional-test/pom.xml
@@ -36,6 +36,12 @@
     <ds.connection.url>jdbc:mysql://localhost:${mysql.port}/${ds.database}?characterEncoding=UTF-8</ds.connection.url>
     <ds.driver.class>com.mysql.jdbc.Driver</ds.driver.class>
 
+    <!-- mysql-dist 5.5.34 is the default version, but 5.6.14 may also work.
+         mysql-dist 5.6.21 won't work due to
+         https://github.com/jcabi/jcabi-mysql-maven-plugin/issues/47
+     -->
+    <mysql.dist.version>5.5.34</mysql.dist.version>
+
     <!-- target zanata instance will be built by war overlay -->
     <zanata.test.war.name>zanata-test-${project.version}</zanata.test.war.name>
     <zanata.instance.url>http://${cargo.host}:${cargo.servlet.port}/${context.path}/</zanata.instance.url>
@@ -477,7 +483,7 @@
                     <artifactItem>
                       <groupId>com.jcabi</groupId>
                       <artifactId>mysql-dist</artifactId>
-                      <version>5.5.34</version>
+                      <version>${mysql.dist.version}</version>
                       <classifier>${mysql.classifier}</classifier>
                       <type>zip</type>
                       <overWrite>false</overWrite>
--- a/zanata-war/src/main/resources/db/changelogs/db.changelog-1.5.xml
+++ b/zanata-war/src/main/resources/db/changelogs/db.changelog-1.5.xml
@@ -132,22 +132,38 @@
             <column name="sourceRef" type="longtext"/>
         </addColumn>   
 	</changeSet>
-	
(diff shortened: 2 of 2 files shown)


zanata-war/src/main/resources/db/changelogs/db.changelog-1.5.xml
```diff
@@ -134,13 +134,24 @@
...
+        <dropPrimaryKey tableName="HTermComment"/>
         <addAutoIncrement tableName="HGlossaryEntry" columnName="id" columnDataType="bigint"/>
         <addAutoIncrement tableName="HGlossaryTerm" columnName="id" columnDataType="bigint"/>
         <addAutoIncrement tableName="HTermComment" columnName="id" columnDataType="bigint"/>
+        <addPrimaryKey tableName="HGlossaryEntry" columnNames="id" />
+        <addPrimaryKey tableName="HGlossaryTerm" columnNames="id" />
+        <addPrimaryKey tableName="HTermComment" columnNames="id" />
+        <addForeignKeyConstraint baseTableName="HGlossaryTerm"
+                baseColumnNames="glossaryEntryId" constraintName="UKglossaryterm_glossary_entry_id"
+                referencedTableName="HGlossaryEntry" referencedColumnNames="id" />
```
davidmason: Can you add an xml comment with the reason for removing then adding the keys and constraints? I assume that the `addAutoIncrement` lines cannot be applied while the keys and constraints are in place, but we should not have to make such an assumption.

seanf: Good point.  Done.

functional-test/pom.xml
```diff
@@ -36,6 +36,9 @@
     <ds.connection.url>jdbc:mysql://localhost:${mysql.port}/${ds.database}?characterEncoding=UTF-8</ds.connection.url>
     <ds.driver.class>com.mysql.jdbc.Driver</ds.driver.class>
 
+    <mysql.dist.version>5.5.34</mysql.dist.version>
+    <!-- <mysql.dist.version>5.6.21</mysql.dist.version> -->
```
davidmason: Did you purposely just modify this?

davidmason: :+1:

seanf: I'm trying to test against both versions, and for the moment it's easier than switching the build params in jenkins.

davidmason: Fair enough. Ping me if anything else changes and it needs another review.

seanf:
@davidmason I have modified the changeset again.  It now works on 5.6.14, just trying 5.5 again now.

seanf:
@davidmason sorry, modified the changeset again.  I just had to add another `validCheckSum` to make it work on existing databases.  Looks like this is a reasonable procedure: https://wiki.openmrs.org/display/docs/How+to+Modify+a+Bad+Changeset

davidmason:
:+1:

seanf:
@djansen-redhat Are you taking this one over?

djansen-redhat:
:white_check_mark: Tested. (Fedora 21, mariadb 10.0.17-1, Wildfly 8.2)
Ran into an issue with installing mariadb as a service, as well as an issue with duplicate mysql-connector-java packages which was resolved by removing the one in JBoss deployment, yum installing it and symlinking it to the deployments dir.

seanf:
> Ran into an issue with installing mariadb as a service, as well as an issue with duplicate mysql-connector-java packages which was resolved by removing the one in JBoss deployment, yum installing it and symlinking it to the deployments dir.


@djansen-redhat that sounds like a problem with zanata-3.6.0-wildfly-8.1.zip, is that right?

Are you saying you had duplicate mysql-connector-java packages, even before you put in the symlink?

PR #757, opened 2015-03-31
Remove log message that was unintentionally committed

davidmason:
Left this in by mistake, and @carlosmunoz noticed that it is generating a bunch of noise in the log.


djansen-redhat:
:see_no_evil:

PR #758, opened 2015-03-31
Rename Obsolete to Archived in UI and tests

djansen-redhat:


functional-test/src/main/java/org/zanata/page/groups/VersionGroupsPage.java
```diff
@@ -68,8 +68,8 @@ public VersionGroupPage goToGroup(String groupName) {
         return new VersionGroupPage(getDriver());
     }
 
-    public VersionGroupsPage toggleObsolete(final boolean show) {
-        WebElement showObsolete = waitForWebElement(toggleObsolete);
+    public VersionGroupsPage toggleArchived(final boolean show) {
+        WebElement showObsolete = waitForWebElement(toggleArchived);
         if (show != showObsolete.isSelected()) {
             showObsolete.click();
```
davidmason: This could be `showArchived` too.

davidmason:
:+1: looks good, has one more easy place to replace. It would be good to get all the keys in `messages.properties` updated too, but we don't have decent refactor support for that so I wouldn't block this change on it.

djansen-redhat:
@davidmason thanks, updated

davidmason:
:+1:

PR #759, opened 2015-03-31
Update join language page

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1207426

Changes on all "contact" dialog
- contact admin
- contact language coordinator

Update on request to join dialog
- request to join group
- request to join language


zanata-war/src/main/java/org/zanata/action/LanguageJoinAction.java
```diff
@@ -42,119 +42,86 @@
...
+        return msgs.format("jsf.language.email.joinrequest.Subject",
+            getLoginName(), getLocaleId().getId());
+    }
+
+    private String getLoginName() {
+        if(authenticatedAccount != null) {
+            return authenticatedAccount.getUsername();
         }
-        return title;
+        return "";
```
davidmason: This looks like someone could send a request to join a language team, even though we don't know who they are. I think if we can't get a username then it is pointless to send this email, and it should abort with an error.

aeng: Agree.

zanata-war/src/main/java/org/zanata/action/VersionGroupJoinAction.java
```diff
@@ -149,14 +152,34 @@ public String send() {
         }
     }
 
-    @AllArgsConstructor
-    public final class SelectableProject {
+    public void resetData() {
```
davidmason: What is the point of this? Please add documentation on this, and add a comment where it is used that indicates why it is used there.

zanata-war/src/main/java/org/zanata/action/VersionGroupJoinAction.java
```diff
@@ -149,14 +152,34 @@ public String send() {
...
+        projectSlug = "";
+        projectVersions.clear();
+        setQuery("");
+    }
 
-        @Getter
-        private HProjectIteration projectIteration;
+    @Override
+    public List<HProject> suggest() {
+        projectVersions.clear();
```
davidmason: It looks like this method is supposed to just return a list of projects, but it is also clearing the versions. I think some code needs to be restructured so that this is not confusing, or maybe add a comment saying why it needs to clear the versions.

zanata-war/src/main/java/org/zanata/action/VersionGroupJoinAction.java
```diff
@@ -149,14 +152,45 @@ public String send() {
         }
     }
 
-    @AllArgsConstructor
-    public final class SelectableProject {
+    /**
+     * This is to reset data when user closes dialog or after sending email.
+     * See version-group/request_join_modal.xhtml#cancelJoinGroupEmail
+     */
+    public void resetData() {
```
davidmason: So it is to clear the form so that it is empty next time they open it? Let's name it so that it says that (`resetData` is extremely generic). I suggest `clearFormFields`.

zanata-war/src/main/java/org/zanata/action/VersionGroupJoinAction.java
```diff
@@ -149,14 +152,45 @@ public String send() {
...
+            projectVersions.clear();
+        }
+    }
+
+    @Override
+    public List<HProject> suggest() {
+        clearProjectVersions();
+        return projectDAO.getProjectsForMaintainer(
+                authenticatedAccount.getPerson(), getQuery(), 0,
+                Integer.MAX_VALUE);
```
davidmason: It looks like you just changed from `projectVersions.clear();` to `clearProjectVersions();`. I think it was already clear what that does. The question is why?

> It looks like this method is supposed to just return a list of projects, but it is also clearing the versions. I think some code needs to be restructured so that this is not confusing, or maybe add a comment saying why it needs to clear the versions.

zanata-war/src/main/resources/messages.properties
```diff
@@ -1178,18 +1177,18 @@ jsf.RequestJoinGroup=Request to add project version to group
 jsf.AlreadyInGroup=Already in Group
 jsf.email.joingrouprequest.AdditionalInfoMessage=To ensure your request is processed without delay, please provide any additional information that will help the group maintainers to process your request.
 !FIXME make key similar to related keys
-jsf.RequestToJoinLanguageTeamTitle=Request To Join "{0}" Language Team
-jsf.RequestRoleLanguageTeamTitle=Request Role in "{0}" Language Team
-
+jsf.email.RequestToJoinLanguageTeam.title=Request to join {0} [{1}] language team
```
davidmason: This needs a source comment that says what `{0}` and `{1}` will be replaced with. It is ambiguous enough that translators should not be left to guess this.

zanata-war/src/main/resources/messages.properties
```diff
@@ -1178,18 +1177,18 @@ jsf.RequestJoinGroup=Request to add project version to group
...
 jsf.email.JoinGroupRequest.Subject=Request to join group "{0}"
-jsf.email.joinrequest.Subject=User "{0}" wants to join the "{1}" language team
-jsf.email.rolerequest.Subject=User "{0}" request for additional role in "{1}" language team
-jsf.email.AdditionalInfoMessage=To ensure your request is processed without delay, please provide any additional information that will help the team coordinators identify you and process your request.
-# {0} = locale display name
-jsf.email.ContactCoordinatorTitle=Contact "{0}" Coordinator
-# {0} = locale ID, {1} = locale native name
-jsf.contactLanguageTeamCoordinatorForLocale=Contact "{0} ({1})" Language Team Coordinators
+jsf.language.email.joinrequest.Subject=User:{0} request to join [{1}] language team
```
davidmason: Suggest: `User "{0}" has requested to join the [{1}] language team

zanata-war/src/main/resources/messages.properties
```diff
@@ -1178,18 +1177,18 @@ jsf.RequestJoinGroup=Request to add project version to group
...
+jsf.language.email.AdditionalInfoMessage=Please provide any additional information that will help the team coordinators identify you and process your request.
+jsf.group.email.AdditionalInfoMessage=Please provide any additional information for group maintainer.
```
davidmason: Suggest: `Please provide any additional information that will help the group maintainers identify your project and process your request.`

zanata-war/src/main/resources/messages.properties
```diff
@@ -1178,18 +1177,18 @@ jsf.RequestJoinGroup=Request to add project version to group
...
+
+jsf.message.admin.inquiry.subject=Inquiry on Zanata
+jsf.message.admin.inquiry.title=Message to Zanata Administrator
 
+# {0} = locale ID, {1} locale display name
+jsf.message.coordinator.inquiry.subject=Inquiry on language {0} [{1}]
+jsf.message.coordinator.inquiry.title=Message to {0} [{1}] coordinator
```
davidmason: These 4 strings seem a bit awkward. Which is used where? Can you put an example of each of the email outputs in the pull request description so they are easy to see in context?

zanata-war/src/main/resources/messages_fr.properties
```diff
@@ -377,7 +377,7 @@ jsf.iteration.CopyProjectValidation.label=Copier les param\u00E8tres de validati
 jsf.iteration.CopyProjectValidations.message=Param\u00E8tres de validation copi\u00E9s \u00E0 partir du projet
 jsf.iteration.CopyProjectType.message=Type de projet copi\u00E9 \u00E0 partir du projet
 jsf.iteration.inheriteLanguage.label=H\u00E9riter les langues \u00E0 partir des param\u00E8tres du projet.
-# translation auto-copied from project Publican, version 4, document publican, author G�rard Baylard
+# translation auto-copied from project Publican, version 4, document publican, author G�rard Baylard
```
davidmason: There is some kind of encoding problem here. Better fix that.

zanata-war/src/main/resources/messages.properties
```diff
@@ -1263,8 +1262,8 @@ jsf.email.coordinator.ReceivedReason=You are a coordinator in the "{0}" language
 
 #------ request-to-join-language email ------
 # name, loginName, localeId, nativeName
-jsf.email.joinrequest.UserRequestingToJoin=Zanata user "{0}" with id "{1}" is requesting to join the {2} ({3}) Language Team
-jsf.email.joinrequest.AddUserInstructions=You can add {0} to the {1} team as translator using the "Add Team Member" action on the language team page and searching for "{2}".
+jsf.email.joinrequest.UserRequestingToJoin=Zanata user "{0}" with id "{1}" is requesting to join the {2} ({3}) language team
+jsf.email.joinrequest.AddUserInstructions=You can add "{0}" to the "{1}" team as translator using the "Add Team Member" action on the language team page.
```
davidmason: Do we allow multiple users to have the same display name? The username is the most reliable way to find a user, so I think we should still mention it explicitly, at least until we make a link that includes the username so that the admin does not need to search.

zanata-war/src/main/java/org/zanata/action/VersionGroupJoinAction.java
aeng: We need to clear the all the versions displayed in dialog from previous selected project when user is entering a new project search(autocomplete)

zanata-war/src/main/webapp/WEB-INF/layout/admin/contact_admin_modal.xhtml
```diff
@@ -7,78 +7,62 @@
   aria-labelledby="myModalLabel" aria-hidden="true">
 
   <script>
-    function cancelEmail() {
-      document.getElementById('#{rich:clientId('subject')}').value = '';
-      document.getElementById('#{rich:clientId('message')}').value = '';
+    $(document).ready(function() {
+      jQuery('#' + jqSelector('#{rich:clientId('contact-admin-subject')}')).
+        val('#{msgs['jsf.message.admin.inquiry.subject']}');
+    });
```
davidmason: What is the purpose of this? It looks like it is inserting a hard-coded string into a field that I can see below that should have the value `#{contactAdminAction.subject}`.

Is this setting the default value for it? Can we do that in contactAdminAction instead so that it is easier to figure out?

zanata-war/src/main/resources/messages_fr.properties
aeng: Its from release branch. Not sure how to fix this.

zanata-war/src/main/webapp/WEB-INF/layout/admin/contact_admin_modal.xhtml
```diff
@@ -7,78 +7,62 @@
...
   <div class="modal__dialog l--constrain-medium">
     <header class="modal__header">
       <h2 class="modal__title">
-        #{msgs['jsf.ContactAdmin']}
+        #{msgs['jsf.message.admin.inquiry.title']}
       </h2>
       <button type="button"
         class="modal__close button--link"
-        data-dismiss="modal" onclick="cancelEmail();return false;">
+        data-dismiss="modal" onclick="cancelContactAdmin();return false;">
```
davidmason: I think this could be shortened to `onclick="cancelContactAdmin"` since that already returns false. At longest it would be `onclick="return cancelContactAdmin()"`

zanata-war/src/main/webapp/WEB-INF/layout/admin/contact_admin_modal.xhtml
```diff
@@ -7,78 +7,62 @@
...
-          <s:decorate id="messageField" template="../field.xhtml">
-            <h:inputTextarea id="message" value="#{contactAdminAction.htmlMessage}"
-              required="true">
-              <a4j:ajax event="blur" render="messageField" execute="@this"/>
-            </h:inputTextarea>
-          </s:decorate>
-        </div>
+      <h:inputHidden id="contact-admin-subject" value="#{contactAdminAction.subject}"/>
+      <div class="modal__content l--pad-all-1">
+        <s:decorate id="messageField" template="../field.xhtml">
```
davidmason: We should not use relative paths to the generic templates like this. All the template paths should be relative to the `zanata-war/src/main/webapp` directory except when the template is specific to a particular component, so that they don't break if we change the directory structure of some xhtml pages.

I think this should be `/WEB-INF/layout/field.xhtml`

djansen-redhat:
There aren't updates to functional tests, specifically ContactLanguageTeamTest and JoinLanguageTeamTest. This looks likely to break them.

zanata-war/src/main/webapp/WEB-INF/layout/language/contact_coordinator_modal.xhtml
```diff
@@ -7,88 +7,71 @@
...
-          </h:outputLabel>
-          <s:decorate id="messageField" template="../field.xhtml">
-            <h:inputTextarea id="message" value="#{sendEmail.htmlMessage}"
-              required="true">
-              <a4j:ajax event="blur" render="messageField" execute="@this"/>
-            </h:inputTextarea>
-          </s:decorate>
-        </div>
+      <div class="modal__content l--pad-all-1">
+        <s:decorate id="messageField" template="../field.xhtml">
```
davidmason: Use absolute template path.

zanata-war/src/main/webapp/WEB-INF/layout/language/request_join_modal.xhtml
```diff
@@ -0,0 +1,114 @@
...
+  class="modal" id="joinLanguageDialog" tabindex="-1" role="dialog"
+  aria-labelledby="myModalLabel" aria-hidden="true">
+
+  <script>
+    $(document).ready(function() {
+      jQuery('#' + jqSelector('#{rich:clientId('request-join-emailType')}')).
+        val('#{languageJoinAction.emailType}');
+      jQuery('#' + jqSelector('#{rich:clientId('request-join-subject')}')).
+        val('#{languageJoinAction.subject}');
+      jQuery('#' + jqSelector('#{rich:clientId('request-join-id')}')).val('#{id}');
```
davidmason: Why are all these set with jQuery instead of just putting them in the template directly?

zanata-war/src/main/webapp/WEB-INF/layout/version-group/request_join_modal.xhtml
```diff
@@ -0,0 +1,132 @@
...
+    #version-list label {
+      padding-left: 0.40625em;
+    }
+
+    #version-list input[type=checkbox] {
+      width: 1em;
+      height: 1em;
+      vertical-align: middle;
+      margin-bottom: 0.20625em;
+    }
```
davidmason: Should these styles be in assets instead?

zanata-war/src/main/webapp/WEB-INF/layout/version-group/request_join_modal.xhtml
```diff
@@ -0,0 +1,132 @@
...
+      toggleButtons(false);
+      return false;
+    }
+
+    function toggleButtons(hasSelection) {
+      if(hasSelection) {
+        jQuery(getJoinGroupForm()).find('#send-button').removeAttr('disabled');
+      } else {
+        jQuery(getJoinGroupForm()).find('#send-button').attr('disabled', true);
+      }
```
davidmason: There is probably a function you can use to toggle the attribute based on the value of `hasSelection`. There is something like that for toggling class so it is worth checking.

zanata-war/src/main/webapp/WEB-INF/urlrewrite.xml
```diff
@@ -413,11 +382,6 @@
     <to>$1/version-group/view/$2</to>
   </outbound-rule>
 
-  <outbound-rule>
-    <from>^(/.+)?/version-group/request_to_join.seam\?emailType=(.+)&amp;versionGroupSlug=(.+)</from>
-    <to>$1/version-group/join/$3</to>
-  </outbound-rule>
```
davidmason: Were all these removed rules just thing that we are not using any more?

zanata-war/src/main/webapp/language/language.xhtml
```diff
@@ -140,6 +177,17 @@
         </div>
       </div>
     </div>
+    <s:fragment rendered="#{identity.loggedIn}">
+      <ui:include src="../WEB-INF/layout/language/contact_coordinator_modal.xhtml">
```
davidmason: This should not need to use `../WEB-INF`, should work with just `/WEB-INF` and will be more robust.

zanata-war/src/main/webapp/language/language.xhtml
```diff
@@ -140,6 +177,17 @@
...
+        <ui:param name="id" value="#{languageAction.language}" />
+        <ui:param name="displayName" value="#{languageAction.locale.retrieveDisplayName()}" />
+      </ui:include>
+      <s:fragment rendered="#{not languageAction.isUserInTeam()}">
+        <ui:include src="../WEB-INF/layout/language/request_join_modal.xhtml">
```
davidmason: This should not need to use `../WEB-INF`, should work with just `/WEB-INF` and will be more robust.

zanata-war/src/main/webapp/version-group/version_group.xhtml
```diff
@@ -215,19 +215,13 @@
...
-                      <f:param value="#{versionGroupHome.instance.name}"/>
-                    </h:outputFormat>
+                  <a href="#" class="i__item--right"
+                    data-toggle="modal" onclick="return false;"
+                    data-target="#joinGroupDialog">
+                    #{msgs.format('jsf.RequestAddProjectToGroup', versionGroupHome.instance.name)}
                     <i class="i i--project i__item__icon"></i>
-                  </s:link>
+                  </a>
+                  <ui:include src="../WEB-INF/layout/version-group/request_join_modal.xhtml"/>
```
davidmason: This should not need to use `../WEB-INF`, should work with just `/WEB-INF` and will be more robust.

davidmason:
Reviewed all the files. There are some questions to answer. The diff for all files is too big so please put the rest of the changes in small, focused diffs.

aeng:
+1. Good one. on it now

Regards, 

Alex Eng 
Senior Software Engineer 
DID: +61 3514 8262 
Mobile: +614 2335 3457 
Email: aeng@redhat.com 

Red Hat, Asia-Pacific Pty Ltd 
Level 1, 193 North Quay 
Brisbane 4000 
Office: +61 7 3514 8100 
Fax: +61 7 3514 8199 
Website: www.redhat.com 

----- Original Message -----

zanata-war/src/main/webapp/WEB-INF/layout/language/request_join_modal.xhtml
aeng: Because sendEmail is a generic bean for email actions and its used in contactAdmin(which is in everypage), as well as contact coordinator, request to join action. To handle these actions, we only set certain variables (form submit) only when user click "send" button. 

I do agree if we need to refactor sendEmail bean to multiple action handler for each type of email. But then it will be too big change for this.

zanata-war/src/main/webapp/WEB-INF/layout/version-group/request_join_modal.xhtml
aeng: jQuery has toggle function for css class but not for attributes. I can see the problem implement that since its not just add/remove an attribute, but also the value.

zanata-war/src/main/webapp/WEB-INF/urlrewrite.xml
aeng: Yes. We are no longer using that.

zanata-war/src/main/java/org/zanata/action/VersionGroupJoinAction.java
```diff
@@ -149,14 +152,47 @@ public String send() {
...
+        projectVersions.clear();
+        setQuery("");
+    }
 
-        @Getter
-        private HProjectIteration projectIteration;
+    // clear all the project versions in the selected project
+    private void clearProjectVersions() {
+        if(!projectVersions.isEmpty()) {
+            projectVersions.clear();
```
davidmason: - [ ] Is there any point to this? `.clear()` should work the same regardless whether the list is empty, so it looks like this is just extra code to maintain.

zanata-war/src/main/java/org/zanata/action/VersionGroupJoinAction.java
davidmason: @aeng I mean why not just leave it as `projectVersions.clear()`?

zanata-war/src/main/java/org/zanata/action/VersionGroupJoinAction.java
aeng: That method no longer needed. Forgot to remove it.

zanata-war/src/main/resources/messages_fr.properties
```diff
@@ -377,7 +377,7 @@ jsf.iteration.CopyProjectValidation.label=Copier les param\u00E8tres de validati
 jsf.iteration.CopyProjectValidations.message=Param\u00E8tres de validation copi\u00E9s \u00E0 partir du projet
 jsf.iteration.CopyProjectType.message=Type de projet copi\u00E9 \u00E0 partir du projet
 jsf.iteration.inheriteLanguage.label=H\u00E9riter les langues \u00E0 partir des param\u00E8tres du projet.
-# translation auto-copied from project Publican, version 4, document publican, author G�rard Baylard
+# translation auto-copied from project Publican, version 4, document publican, author Gérard Baylard
```
davidmason: This is still showing up as being different, but the lines look exactly the same. It must be something to do with the encoding. I hope our client is outputting the comment lines with the right encoding. Comment lines that do not have non-ascii characters are showing up as unchanged. This is a bit of a mystery, and I think we had better figure out what is happening so we know whether it can cause any problems.

- [ ] how did you update this file?

zanata-war/src/main/webapp/WEB-INF/layout/language/request_join_modal.xhtml
```diff
@@ -0,0 +1,114 @@
...
+              onValueChanged="bindRole" label="#{msgs['jsf.Reviewer']}"/>
+          </s:fragment>
+
+          <s:fragment rendered="#{!languageJoinAction.isCoordinator()}">
+            <zanata:checkbox value="coordinator" checked="#{languageJoinAction.requestAsCoordinator}"
+              onValueChanged="bindRole" label="#{msgs['jsf.Coordinator']}"/>
+          </s:fragment>
+        </div>
+
+        <s:decorate id="messageField" template="../field.xhtml">
```
davidmason: - [ ] can use absolute path here

zanata-war/src/main/webapp/WEB-INF/layout/version-group/request_join_modal.xhtml
```diff
@@ -0,0 +1,125 @@
...
+                    </h:selectBooleanCheckbox>
+                    <label>#{version.iterationSlug}</label>
+                  </li>
+                </ui:repeat>
+              </ul>
+            </s:fragment>
+          </s:div>
+        </div>
+
+        <s:decorate id="messageField" template="../field.xhtml">
```
davidmason: - [ ] can use absolute path here

zanata-war/src/main/webapp/WEB-INF/layout/language/request_join_modal.xhtml
davidmason: Ah, so we want to bind the fields such as `emailType` in the `sendEmail` bean, but we want the value to come from `languageJoinAction`. I agree we should refactor it, either to have the `sendEmail` get the values directly from the other bean, or have several beans that inherit from `sendEmail`. Should be ok to wait.

zanata-war/src/main/webapp/admin/roledetail.xhtml
```diff
@@ -13,13 +13,6 @@
       .groups label {
         padding-left: 0.40625em;
       }
-
-      .groups input[type=checkbox] {
-        width: 1em;
-        height: 1em;
-        vertical-align: middle;
-        margin-bottom: 0.20625em;
-      }
```
davidmason: - [ ] @aeng Are these styles in assets now? I notice there are several places that have the same set of styles.

aeng: @davidmason yes. all check box style is in assets now.

zanata-war/src/main/webapp/WEB-INF/layout/version-group/request_join_modal.xhtml
davidmason: @aeng does it work if you just do this?

```
jQuery(getJoinGroupForm()).find('#send-button').prop('disabled', !hasSelection);
```

If not, it can still be more readable by extracting the repeated code:

```
var sendButton = jQuery(getJoinGroupForm()).find('#send-button');
if(hasSelection) {
  sendButton.removeAttr('disabled');
} else {
  sendButton.attr('disabled', true);
}
```

zanata-war/src/main/webapp/WEB-INF/layout/version-group/request_join_modal.xhtml
```diff
@@ -0,0 +1,127 @@
...
+      jQuery(getJoinGroupForm()).find('#versionGroupAutocomplete-autocomplete__input').val('');
+      jQuery(getJoinGroupForm()).find('#request-join-group-send-button').removeClass('is-active');
+      toggleButtons(false);
+      return false;
+    }
+
+    function toggleButtons(hasSelection) {
+      var sendButton = jQuery(getJoinGroupForm()).find('#request-join-group-send-button');
+
+      if(hasSelection === 'true') {
```
davidmason: Are we really passing in the string `'true'` to this? That is a very unexpected thing to do. Can we get a boolean instead?

zanata-war/src/main/webapp/WEB-INF/layout/version-group/request_join_modal.xhtml
```diff
@@ -0,0 +1,127 @@
...
+        sendButton.removeAttr('disabled');
+      } else {
+        sendButton.attr('disabled', 'disabled');
+      }
```
davidmason: - [ ] so did the `.prop` approach not work? That will be useful knowledge for next time.

zanata-war/src/main/webapp/WEB-INF/layout/version-group/request_join_modal.xhtml
aeng: toggleButtons('#{versionGroupJoinAction.hasSelectedVersion()}') always passes string "true" for some reason.

davidmason: The reason is that there are quotes around the EL expression. Just remove them :speak_no_evil:

aeng: Doesn't work when you remove them. EL expression

davidmason: So what happens when you change it to `toggleButtons(#{versionGroupJoinAction.hasSelectedVersion()})`? and change the check back to `if (hasSelection) { ...` ?

aeng: It won't even trigger toggleButton. Need single quote for el expression.

aeng:
Jenkins passed. Status not updated.

davidmason:
@aeng :+1: :see_no_evil: :white_check_mark: :ok: :ok_hand:

aeng:

djansen-redhat:
- [ ] Odd behaviour on Role checkboxes - once I checked Translator, Reviewer and the email only said Reviewer; once I clicked Send with nothing checked and it sent with the previously checked items included

zanata-war/src/main/webapp/WEB-INF/layout/language/contact_coordinator_modal.xhtml
```diff
@@ -7,89 +7,65 @@
...
-            <h:inputTextarea id="message" value="#{sendEmail.htmlMessage}"
-              required="true">
-              <a4j:ajax event="blur" render="messageField" execute="@this"/>
-            </h:inputTextarea>
-          </s:decorate>
-        </div>
+      <div class="modal__content l--pad-all-1">
+        <s:decorate id="messageField" template="/WEB-INF/layout/field.xhtml">
+          <h:inputTextarea id="contact-coordinator-message" value="#{languageContactCoordinatorAction.message}"
+            required="true" styleClass="l--push-bottom-0" onkeyup="checkMessage()">
```
djansen-redhat: This doesn't work if the user pastes their message into the field

djansen-redhat:
:white_check_mark: Tested

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/f4267c26d0de00aa9019b42f7c97bd48d2f5b52d

PR #760, opened 2015-03-31
rhbz1165939 Do not display empty actions panel on groups page

djansen-redhat:


aeng:
The idea of `right_content` action panel is to allow link to any action we want to add in the future for version group.  But for now, it make sense not to show the panel as "creating new group" is the only action. :+1:

PR #763, opened 2015-04-02
Separate wildfly client version from server version

seanf:
I'm hoping this will help with running Zanata on WildFly 9.

Edit: looks like it will take more, but I think it's still a good change.


seanf:
This pull request seems to contain the wrong contents.  I'll have to investigate.

PR #764, opened 2015-04-02
Remove unwanted RichFaces dependencies

seanf:


aeng:
:+1: as long as the full functional test passes.

PR #765, opened 2015-04-06
Update the copyright year to 2015

djansen-redhat:


aeng:
I wonder if there's anything we can do when updating a war, it just updates the statement to 
`Copyright &#169; 2008-{currentYear} <a href="http://www.redhat.com/" target="_blank">Red Hat, Inc</a>` ?

zanata-war/src/main/java/org/zanata/ZanataInit.java
```diff
@@ -430,7 +431,8 @@ private void logBanner(VersionInfo ver, AppServerVersion appServerVersion) {
         }
         log.info("  AS version: " + appServerVersion.asVersion.orNull());
         log.info("  SCM: " + ver.getScmDescribe());
-        log.info("  Red Hat Inc 2008-2014");
+        log.info("  Red Hat Inc 2008-{}",
```
seanf: Just because the code is running in the year X doesn't mean the copyright year of publication has changed.  We should just put the year of publication: 2008.

Wait, is this even a copyright statement?  It doesn't say copyright or ©, so what is it supposed to be?

djansen-redhat: I don't really care, but the issue was that the later year was wrong in *any* case. Having the current year *looks* good, as the human brain tends to see old years as bad.
"© year-year company" is a generally acceptable copyright label, and in a legal stoush having one is better than not having one.

aeng: :arrow_up:  :+1: .

zanata-war/src/main/java/org/zanata/action/CopyrightAction.java
```diff
@@ -0,0 +1,48 @@
...
+    private static final long serialVersionUID = 1L;
+
+    @In
+    private Messages msgs;
+
+    public String copyrightNotice() {
+        return msgs.format("jsf.CopyrightNotice",
+                String.valueOf(Calendar.getInstance().get(Calendar.YEAR)));
+    }
+}
```
aeng: Don't think we need an action bean just for getting year. You can implement the method in applicationConfiguration bean.

djansen-redhat: I thought this might be a bit heavy-handed - will fix

djansen-redhat:
Updated as per comment.

aeng:
:+1: :clap:

djansen-redhat:

djansen-redhat:

definite:
:+1:

PR #768, opened 2015-04-08
Add Undertow modules for WildFly tests

seanf:
https://bugzilla.redhat.com/show_bug.cgi?id=1197955

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/768)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1218,6 +1218,7 @@
         <mojarra.module.name>wildfly-${module.wildfly.version}-module-mojarra-${mojarra.module.version}.zip</mojarra.module.name>
         <hibernate.module.version>4.2.15.Final</hibernate.module.version>
         <hibernate.module.name>wildfly-${module.wildfly.version}-module-hibernate-main-${hibernate.module.version}.zip</hibernate.module.name>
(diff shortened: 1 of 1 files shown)


seanf:
Looks like you can't upgrade undertow-core without upgrading undertow-servlet, and the same probably goes for the other undertow modules (like websockets).  So I have grabbed all the undertow modules from WildFly 9.0.0.Beta2, to see if they can work in WF 8.1.

seanf:
Looks like all the functional tests passed on WF8.1, apart from three which had this (intermittent) error:

    unknown error: Element is not clickable at point (373, 26). Other element would receive the click

seanf:
It looks like this wasn't testing what I thought at all.  When the functional tests passed on WF 8.1, I was using a badly-packaged Undertow module which apparently wasn't upgrading Undertow at all.  The properly packaged Undertow appears to break badly on WF 8.1 (XNIO is too old).  There's a chance it will work on 8.2, but other 8.2 problems need to be solved.

Reverted by commit e3ffc75a09875eb20b697ae315b742ae2d072e30.

seanf:
At this point, I think we would be better off ensuring Zanata runs well on WildFly 9, which already includes the Undertow 1.2 without hacking around.

PR #772, opened 2015-04-13
Implement server side file conversion for `properties` in file project-type

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1186972

1. common: https://github.com/zanata/zanata-common/pull/9
2. api: https://github.com/zanata/zanata-api/pull/26
3. client: https://github.com/zanata/zanata-client/pull/57
4. **server**: https://github.com/zanata/zanata-server/pull/772


zanata-war/src/main/java/org/zanata/action/VersionHomeAction.java
```diff
@@ -879,8 +887,12 @@ private void uploadAdapterFile() {
             HRawDocument rawDocument = new HRawDocument();
             rawDocument.setDocument(document);
             rawDocument.setContentHash(new String(Hex.encodeHex(md5hash)));
-            rawDocument.setType(DocumentType.typeFor(translationFileServiceImpl
-                    .extractExtension(fileName)));
+            if(docType.isPresent()) {
+                rawDocument.setType(DocumentType.getByName(docType.get()));
+            } else {
+                rawDocument.setType(DocumentType.fromSourceExtension(
+                    FileUtil.extractExtension(fileName)).iterator().next());
```
davidmason: - [ ] Why does this have `.iterator().next()` on it? Is `extractExtension()` returning a collection instead of just a single extension? I would only expect to find one extension on a filename.

aeng: What happens if there's multiple adapter that uses same source extension? e.g. PROPERTIES and PROPERTIES_UTF8.

davidmason: Oh yeah, that is a tricky one - the safest is to force the user to choose one when it is ambiguous like that.

zanata-war/src/main/java/org/zanata/adapter/GettextAdapter.java
```diff
@@ -0,0 +1,171 @@
...
+                e);
+        } catch (MalformedURLException e) {
+            throw new FileFormatAdapterException(
+                "Could not open the URI. The URI may be malformed: "
+                    + ((url == null) ? "URL is null" : url.toString()),
+                e);
+        } catch (IOException e) {
+            throw new FileFormatAdapterException(
+                "Could not open the URL. The URL is OK but the input stream could not be opened.\n"
+                    + e.getMessage(), e);
```
davidmason: Can these exception messages use format strings with placeholders instead? I forget whether that is standard for exceptions.

zanata-war/src/main/java/org/zanata/adapter/OkapiFilterAdapter.java
```diff
@@ -354,8 +363,23 @@ public void writeTranslatedFile(OutputStream output, URI originalFile,
         }
     }
 
+    /**
+     * Transform list of TextFlowTarget to map with TextFlowTarget.resId as key
+     *
+     * @param targets
+     * @return
+     */
+    private Map<String, TextFlowTarget> transformToMap(List<TextFlowTarget> targets) {
```
davidmason: Can we call this `mapByResId`?

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
```diff
@@ -0,0 +1,225 @@
...
+import org.zanata.exception.FileFormatAdapterException;
+import org.zanata.rest.dto.resource.Resource;
+import org.zanata.rest.dto.resource.TranslationsResource;
+
+import com.google.common.base.Charsets;
+import com.google.common.base.Optional;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Adapter for reading and write {@link org.zanata.common.DocumentType#PROPERTIES} file
```
davidmason: (grammar) "... for reading and writing ..." or "... to read and write ..."

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
```diff
@@ -0,0 +1,225 @@
...
+    public void writeTranslatedFile(OutputStream output, URI originalFile,
+        Resource resource, TranslationsResource translationsResource,
+        String locale, Optional<String> params)
+        throws FileFormatAdapterException, IllegalArgumentException {
+
+        writeTranslatedFile(output, resource, translationsResource, locale,
+                params, ISO_8859_1);
+    }
+
+    public Resource parseDocumentFile(URI fileUri, LocaleId sourceLocale,
```
davidmason: Is this an override, or just a new public method?

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
```diff
@@ -0,0 +1,225 @@
...
+        Optional<String> filterParams, String charset) throws FileFormatAdapterException,
+        IllegalArgumentException {
+
+        if (sourceLocale == null) {
+            throw new IllegalArgumentException("Source locale cannot be null");
+        }
+
+        if (fileUri == null) {
+            throw new IllegalArgumentException("Document URI cannot be null");
```
davidmason: - [ ] I wonder if we should add some non-null annotations to the interface for these things, so that these checks are not needed. There are some of these checks earlier in the diff that could be avoided with the right annotations too.

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
```diff
@@ -0,0 +1,225 @@
...
+        try {
+            inputStream = readStream(fileUri);
+            propReader.extractTemplate(doc, inputStream);
+        } catch (IOException e) {
+            throw new FileFormatAdapterException(
+                "Could not open the URL. The URL is OK but the input stream could not be opened.\n"
+                    + e.getMessage(), e);
+        } finally {
+            try {
+                inputStream.close();
```
davidmason: - [ ] Did you try the try-with-resources pattern for this one?

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
```diff
@@ -0,0 +1,225 @@
...
+            } catch (IOException e1) {
+            }
+        }
+        return doc;
+    }
+
+    public TranslationsResource parseTranslationFile(URI fileUri,
```
davidmason: - [ ] Is this an override?

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
```diff
@@ -0,0 +1,225 @@
...
+        String localeId, Optional<String> params, String charset)
+        throws FileFormatAdapterException, IllegalArgumentException {
+
+        if (StringUtils.isEmpty(localeId)) {
+            throw new IllegalArgumentException(
+                "locale id string cannot be null or empty");
```
davidmason: - [ ] This `localeId` is not actually used, so maybe we don't need to check, or we should be actually using it instead of just `LocaleId.EN_US`.

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        PropReader propReader = new PropReader(charset, LocaleId.EN_US,
```
davidmason: - [ ] Should this look up the locale from `localeId` string?

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
```diff
@@ -0,0 +1,225 @@
...
+        try {
+            inputStream = readStream(fileUri);
+            propReader.extractTarget(targetDoc, inputStream, srcDoc);
+        } catch (IOException e) {
+            throw new FileFormatAdapterException(
+                "Could not open the URL. The URL is OK but the input stream could not be opened.\n"
+                    + e.getMessage(), e);
+        } finally {
+            try {
+                inputStream.close();
```
davidmason: - [ ] did you try the try-with-resources pattern for this?

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
```diff
@@ -0,0 +1,225 @@
...
+            tempFile = File.createTempFile("filename", "extension");
+            if(charset.equals(ISO_8859_1)) {
+                PropWriter.writeTranslationsFile(resource,
+                        translationsResource, tempFile,
+                        PropWriter.CHARSET.Latin1, createSkeletons);
+            } else if (charset.equals(UTF_8)) {
+                PropWriter.writeTranslationsFile(resource,
+                    translationsResource,
+                    tempFile, PropWriter.CHARSET.UTF8, createSkeletons);
+            }
```
davidmason: - [ ] At this point if the charset was not either of the expected charsets, the file was not written. Better add an `else` with an error about incorrect charset.

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
```diff
@@ -0,0 +1,225 @@
...
+        try {
+            tempFile = File.createTempFile("filename", "extension");
+            if(charset.equals(ISO_8859_1)) {
+                PropWriter.writeTranslationsFile(resource,
+                        translationsResource, tempFile,
+                        PropWriter.CHARSET.Latin1, createSkeletons);
+            } else if (charset.equals(UTF_8)) {
+                PropWriter.writeTranslationsFile(resource,
+                    translationsResource,
+                    tempFile, PropWriter.CHARSET.UTF8, createSkeletons);
```
davidmason: - [ ] These two lines are almost identical. Probably easier to maintain if we split the charset check out:

```
if (charset.equals(ISO_8859_1)) {
  propWriterCharset = PropWriter.CHARSET.Latin1;
} else if (charset.equals(UTF_8)) {
  propWriterCharset = PropWriter.CHARSET.UTF8;
} else {
  // throw error - invalid charset
}

PropWriter.writeTranslationsFile(resource, translationsResource,
        tempFile, propWriterCharset, createSkeletons);
```

davidmason: - [ ] Then probably move the whole `if-else-if-else` block to a function `toPropWriterCharset(...)` and that whole chunk of code becomes just:

```
PropWriter.writeTranslationsFile(resource, translationsResource,
        tempFile, toPropWriterCharset(charset), createSkeletons);
```

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
```diff
@@ -0,0 +1,225 @@
...
+            }
+
+            byte[] buffer = new byte[4096]; // To hold file contents
+            int bytesRead;
+            FileInputStream input = new FileInputStream(tempFile);
+            while ((bytesRead = input.read(buffer)) != -1) {
+                output.write(buffer, 0, bytesRead);
```
davidmason: - [ ] This logic is used in a few places, and is too low-level to put with this code anyway. Better extract it to a utility method `writeFileToOutputStream(File file, OutputStream output)`.
- [ ] Was this copied from one of the existing adapters? If so, might be good to replace that with a method call too.

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
```diff
@@ -0,0 +1,225 @@
...
+        } finally {
+            try {
+                inputStream.close();
+            } catch (IOException e1) {
+            }
+        }
+        return targetDoc;
+    }
+
+    public void writeTranslatedFile(OutputStream output, Resource resource,
```
davidmason: - [ ] is this an override?

zanata-war/src/main/java/org/zanata/adapter/PropertiesUTF8Adapter.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Adapter for reading and write {@link org.zanata.common.DocumentType#PROPERTIES_UTF8} file
+ *
+ * TODO: Convert to okapi properties adapter once all client conversion is
+ * migrated to server
+ *
+ * @author Alex Eng <a href="mailto:aeng@redhat.com">aeng@redhat.com</a>
+ */
+public class PropertiesUTF8Adapter extends PropertiesAdapter {
```
davidmason: - [ ] I suggest instead of this class, make a top-level `GenericPropertiesAdapter` that takes a charset in its constructor:

```
public class GenericPropertiesAdapter implements FileFormatAdapter {
    public static final String charset;

    public GenericPropertiesAdapter(String charset) {
        // check that charset is valid, throw exception if not

        this.charset = charset;
    }

    // make all the methods just use this.charset

}
```

Then have `PropertiesAdapter` and `PropertiesUTF8Adapter` that just need to implement the constructor:

```
public class PropertiesAdapter extends GenericPropertiesAdapter {
    public GenericPropertiesAdapter() {
        super(ISO_8859_1);
    }
}
```

... and same for UTF_8

davidmason: Note: `GenericPropertiesAdapter` would have all the code that is in `PropertiesAdapter` now, just without having to have any of the if statements about charset.

zanata-war/src/main/java/org/zanata/adapter/XliffAdapter.java
```diff
@@ -0,0 +1,147 @@
...
+import org.zanata.adapter.xliff.XliffWriter;
+import org.zanata.common.LocaleId;
+import org.zanata.exception.FileFormatAdapterException;
+import org.zanata.rest.dto.resource.Resource;
+import org.zanata.rest.dto.resource.TranslationsResource;
+
+import com.google.common.base.Optional;
+
+/**
+ * Adapter for reading and write {@link org.zanata.common.DocumentType#XLIFF} file
```
davidmason: (grammar) "... for reading and writing ..." or "... to read and write ..."

zanata-war/src/main/java/org/zanata/adapter/XliffAdapter.java
```diff
@@ -0,0 +1,147 @@
...
+            tempFile = File.createTempFile("filename", "extension");
+            XliffWriter.writeFile(tempFile, resource, locale,
+                translationsResource, createSkeletons);
+
+            byte[] buffer = new byte[4096]; // To hold file contents
+            int bytesRead;
+            FileInputStream input = new FileInputStream(tempFile);
+            while ((bytesRead = input.read(buffer)) != -1) {
+                output.write(buffer, 0, bytesRead);
+            }
```
davidmason: - [ ] replace this with a call to the method `writeFileOutputToStream(...)` that I mentioned earlier.

zanata-war/src/main/java/org/zanata/adapter/XliffAdapter.java
```diff
@@ -0,0 +1,147 @@
...
+                "Unable to generate translated file", e);
+        } finally {
+            if (tempFile != null) {
+                if (!tempFile.delete()) {
+                    log.warn(
+                        "unable to remove temporary file {}, marked for delete on exit",
+                        tempFile.getAbsolutePath());
+                    tempFile.deleteOnExit();
+                }
+            }
```
davidmason: - [ ] This block appears several times now. We should extract it to a utility method like `tryReallyHardToDeleteFile(File file)`

zanata-war/src/main/java/org/zanata/file/SourceDocumentUpload.java
```diff
@@ -147,17 +148,23 @@ public Response tryValidatedUploadSourceFile(GlobalDocumentId id,
...
             }
 
-            if (DocumentType.typeFor(uploadForm.getFileType()) == DocumentType.GETTEXT_PORTABLE_OBJECT_TEMPLATE) {
-                InputStream potStream = getInputStream(tempFile, uploadForm);
-                parsePotFile(potStream, id, uploadForm);
-            } else {
+            HProjectIteration version =
+                projectIterationDAO.getBySlug(id.getProjectSlug(), id.getVersionSlug());
+
+            if(version.getProjectType() == ProjectType.File) {
```
davidmason: - [ ] Could version be null at this point?

zanata-war/src/main/java/org/zanata/rest/service/FileService.java
```diff
@@ -271,24 +298,28 @@ public Response downloadTranslationFile(String projectSlug,
...
             StreamingOutput output =
-                    new FormatAdapterStreamingOutput(uri, translations, locale,
-                            adapter, params);
+                    new FormatAdapterStreamingOutput(uri, res, transRes,
+                        locale, adapter, params);
+
+            String srcExt = FilenameUtils.getExtension(document.getName());
+            String transExt = hRawDocument.getType().getExtensions().get(srcExt);
+            String transFileName =  FilenameUtils.removeExtension(document
+                    .getName()) + "." + transExt;
```
davidmason: These 3 lines would make a nice little utility method: `public String generateTranslationFileName(HDocument document)`

zanata-war/src/main/java/org/zanata/service/TranslationFileService.java
```diff
@@ -106,6 +107,10 @@ TranslationsResource parseAdapterTranslationFile(File tempFile,
     Resource parseUpdatedPotFile(InputStream fileContents, String docId,
             String uploadFileName, boolean offlinePo);
 
+    boolean hasMultipleAdapter(String fileNameOrExtension);
```
davidmason: suggest `hasMultipleAdapters(...)`

zanata-war/src/main/java/org/zanata/service/TranslationFileService.java
```diff
@@ -186,24 +193,16 @@ Resource parseUpdatedAdapterDocumentFile(URI documentFile, String docId,
...
+    String getTranslationFileExtension(String projectSlug, String iterationSlug,
+        String docPath, String docName);
 
     /**
      * @return true if the specified document is of type po, false if it is any
      *         other type, including null.
      */
-            boolean
+    boolean
             isPoDocument(String projectSlug, String iterationSlug, String docId);
```
davidmason: - [ ] This should not be split over 2 lines.

zanata-war/src/main/java/org/zanata/service/impl/TranslationFileServiceImpl.java
```diff
@@ -117,19 +133,23 @@
...
-            String iterationSlug, String docId) throws ZanataServiceException {
-        if (fileName.endsWith(".po")) {
-            return parsePoFile(fileContents, projectSlug, iterationSlug, docId);
-        } else if (hasAdapterFor(fileName)) {
+            String iterationSlug, String docId, Optional<String> documentType) throws ZanataServiceException {
+
+        HProjectIteration version =
+            projectIterationDAO.getBySlug(projectSlug, iterationSlug);
+
+        if(version.getProjectType() == ProjectType.File) {
```
davidmason: - [ ] Could version be null at this point?

zanata-war/src/main/java/org/zanata/service/impl/TranslationFileServiceImpl.java
```diff
@@ -117,19 +133,23 @@
...
             File tempFile = persistToTempFile(fileContents);
             TranslationsResource transRes =
-                    parseAdapterTranslationFile(tempFile, projectSlug,
-                            iterationSlug, docId, localeId, fileName);
+                parseAdapterTranslationFile(tempFile, projectSlug,
+                    iterationSlug, docId, localeId, fileName, documentType);
             removeTempFile(tempFile);
             return transRes;
+        } else if(fileName.endsWith(".po")) {
```
davidmason: - [ ] You keep putting if statements without a space (`if(foo` instead of `if (foo`). I think our style has consistently been with a space. Does this pass checkstyle?

zanata-war/src/main/java/org/zanata/service/impl/TranslationFileServiceImpl.java
```diff
@@ -282,24 +287,35 @@ private boolean hasAdapterFor(String fileNameOrExtension) {
...
+                "Cannot choose an adapter because the provided string '"
+                    + fileNameOrExtension
+                    + "' does not match any known document type.");
+        }
+        FileFormatAdapter adapter = getAdapterFor(documentType);
+        if(hasMultipleAdapter(fileNameOrExtension)) {
+            /**
+             * TODO: throw runtime error. Need to wait for all upload file
+             * dialog implement multiple adapter check for file extension.
+             */
```
davidmason: - [ ] Why does it have to wait? Is it because it would be impossible for properties files to be uploaded at the moment? I would expect the requirement is that the dialogs allow the user to choose an adapter.

zanata-war/src/main/java/org/zanata/service/impl/TranslationFileServiceImpl.java
```diff
@@ -316,23 +332,23 @@ public FileFormatAdapter getAdapterFor(DocumentType type) {
...
-                            .lastIndexOf('.') + 1);
-        } else {
-            extension = fileNameOrExtension;
+    /**
+     * Return adapter for given documentType if present, otherwise return adapter
+     * with given fileName.
+     *
+     * @param documentType
+     * @param fileName
+     * @return
```
davidmason: Move the summary text here, since it is describing the return value. The summary text could be "Get an appropriate adapter for a document type or file name."

zanata-war/src/main/java/org/zanata/servlet/FileUploadServlet.java
```diff
@@ -1,193 +0,0 @@
...
- * Used for translation file upload from GWT editor. For endpoint, see servlet
- * binding for this class in web.xml
- *
- * @see {@link FileUploadDialog}
- *
- * @author Alex Eng <a href="mailto:aeng@redhat.com">aeng@redhat.com</a>
- *
- */
-@Slf4j
-public class FileUploadServlet extends HttpServlet {
```
davidmason: - [ ] So translation file upload from the editor uses a different endpoint now?

zanata-war/src/main/java/org/zanata/servlet/MultiFileUploadServlet.java
```diff
@@ -242,29 +243,28 @@ private void respondWithFiles(HttpServletResponse response, JSONArray filesJson)
...
 
         private SourceDocumentUpload sourceUploader;
-        private TranslationFileService translationFileServiceImpl;
 
         public FileUploadRequestHandler(HttpServletRequest request) {
             this.request = request;
             projectSlug = request.getParameter("p");
             versionSlug = request.getParameter("v");
-
+            //TODO: alex: add types parameter in all caller of /files/upload
```
davidmason: - [ ] Is there a bugzilla for the next stage of improving the upload dialog, so we can make a note of this in it?

zanata-war/src/main/java/org/zanata/servlet/MultiFileUploadServlet.java
```diff
@@ -242,29 +243,28 @@ private void respondWithFiles(HttpServletResponse response, JSONArray filesJson)
...
+            String fileTypesQuery = request.getParameter("types");
```
davidmason: - [ ] Is this something a user might type in? We should make sure it fits the expected pattern before we split it up.

zanata-war/src/main/java/org/zanata/servlet/MultiFileUploadServlet.java
```diff
@@ -392,7 +392,23 @@ private DocumentFileUploadForm createUploadFormForItem(FileItem item) throws IOE
...
+                    DocumentType docType = DocumentType.valueOf(parsedFileType);
+                    if(docType != null && docType.getSourceExtensions().contains(extension)) {
+                        fileType = docType.name();
+                        break;
+                    }
+                }
+            }
+            fileType =
+                    fileType != null ? FileUtil
+                            .extractExtension(item.getName()) : fileType;
```
davidmason: - [ ] Could just initialize fileType with the extracted extension and remove this line.

```
String filtType = FileUtil.extractExtension(item.getName());
```

zanata-war/src/main/java/org/zanata/servlet/MultiFileUploadServlet.java
```diff
@@ -392,7 +392,23 @@ private DocumentFileUploadForm createUploadFormForItem(FileItem item) throws IOE
...
+
+            form.setFileType(fileType);
```
davidmason: - [ ] Should put all the new code just above here into a method, so that this line is something like:

```
form.setFileType(getFileTypeForItem(item));
```

zanata-war/src/main/java/org/zanata/util/FileUtil.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Utility class for File
+ *
+ * @author Alex Eng <a href="mailto:aeng@redhat.com">aeng@redhat.com</a>
+ */
+public class FileUtil {
+
+    /**
+     *
```
davidmason: This should have a proper summary description since it is now for general use.

zanata-war/src/main/java/org/zanata/util/FileUtil.java
```diff
@@ -0,0 +1,90 @@
...
+
+
+    /**
+     * Generate documentId by concatenating path with fileName
+     *
+     * e.g "foo", "bar.txt" = "foo/bar.txt"
+     *
+     * @param path
+     * @param fileName
+     * @return
```
davidmason: - [ ] either remove this line, or actually say what it returns.

davidmason:
@aeng does this PR remove the ability to upload translations from the editor document list?

zanata-war/src/main/webapp/WEB-INF/layout/version/upload_translation_modal.xhtml
```diff
@@ -4,63 +4,87 @@
...
+        toggleUploadTranslationDocButton(isFileSelected, true);
       });
     });
 
-    function getUploadButton() {
+    function toggleUploadTranslationDocButton(enable, display) {
+      var uploadButton = getUploadTranslationDocButton();
+      jQuery(uploadButton).prop('disabled', !enable);
+      jQuery(uploadButton).toggleClass('is-hidden', !display);
+    }
```
davidmason: - [ ] This function is confusing at the call site. I think it should be split into `setUploadTranslationDocButtonEnabled(boolean)` and `SetUploadTranslationDocButtonVisible(boolean)` so that it is clear at the call site what is happening.

zanata-war/src/main/webapp/WEB-INF/layout/version/upload_translation_modal.xhtml
```diff
@@ -4,63 +4,87 @@
...
         }
+        toggleUploadTranslationDocButton(isFileSelected, true);
       });
     });
 
-    function getUploadButton() {
+    function toggleUploadTranslationDocButton(enable, display) {
+      var uploadButton = getUploadTranslationDocButton();
+      jQuery(uploadButton).prop('disabled', !enable);
+      jQuery(uploadButton).toggleClass('is-hidden', !display);
```
davidmason: - [ ] It looks like `uploadButton` is already a jQuery object, no need to wrap in `jQuery(...)`.

zanata-war/src/main/webapp/WEB-INF/layout/version/upload_translation_modal.xhtml
```diff
@@ -4,63 +4,87 @@
   xmlns:zanata="http://java.sun.com/jsf/composite/zanata"
   xmlns:a4j="http://richfaces.org/a4j"
   xmlns:rich="http://richfaces.org/rich"
+  xmlns:f="http://java.sun.com/jsf/core"
   class="modal" id="uploadTranslationDialog" tabindex="-1" role="dialog"
   aria-labelledby="myModalLabel" aria-hidden="true">
 
   <script>
     $(document).ready(function() {
```
davidmason: - [ ] are we using `$` or `jQuery`? This file has both, we should be consistent.

zanata-war/src/main/webapp/WEB-INF/layout/version/upload_translation_modal.xhtml
```diff
@@ -4,63 +4,87 @@
...
-      jQuery(uploadButton).removeClass('is-hidden');
-
-      $(getUploadFile()).change(function() {
-        var uploadButton = getUploadButton();
-        if(jQuery(getUploadFile()).val() != '') {
-          jQuery(uploadButton).removeAttr('disabled');
-        } else {
-          jQuery(uploadButton).attr('disabled', true);
+      toggleUploadTranslationDocButton(false, true);
+      $(getUploadFileComponent()).change(function() {
```
davidmason: - [ ] This is already a jQuery object, no need to wrap in `$(...)`

zanata-war/src/main/webapp/WEB-INF/layout/version/upload_translation_modal.xhtml
```diff
@@ -4,63 +4,87 @@
...
+    function getUploadFileComponent() {
       return jQuery('#' + jqSelector('#{rich:clientId('upload-file')}'));
     }
 
+    function clearDocumentTypeSelection() {
+      jQuery('#' + jqSelector('#{rich:clientId('documentType-selection')}')).html('');
+    }
+
+    function getUploadFileName() {
+      return jQuery(getUploadFileComponent()).val();
```
davidmason: - [ ] Already a jQuery object, no need to wrap it.

zanata-war/src/main/webapp/WEB-INF/layout/version/upload_translation_modal.xhtml
```diff
@@ -4,63 +4,87 @@
...
+
+    function getDocTypeSelection() {
+      return jQuery('#' + jqSelector('#{rich:clientId('documentTypes')}'));
+    }
+
     function cancelUpload() {
       zanata.loader.deactivate('#upload-translation-loader');
-      jQuery(getUploadFile()).val('');
-      jQuery(getUploadButton()).attr('disabled', true);
+      jQuery(getUploadFileComponent()).val('');
```
davidmason: - [ ] Already a jQuery object, no need to wrap it.

zanata-war/src/main/webapp/iteration/view.xhtml
```diff
@@ -84,16 +84,31 @@
     return false;
   }
 
-  function checkIfSpecificFileSelected() {
-    checkIfFileSelected(#{rich:element('specificDocFileInput')},
-      #{rich:element('specificDocSubmitUploadButton')});
+  function checkIfFileSelected() {
+    var isFileSelected = jQuery(getSourceDocInput()).val() != '';
```
davidmason: - [ ] `getSourceDocInput()` returns a jQuery object, no need to wrap it.

zanata-war/src/main/webapp/iteration/view.xhtml
```diff
@@ -84,16 +84,31 @@
...
+    toggleUploadSourceButton(isFileSelected);
+    if(isFileSelected) {
+      renderSourceDocTypeSelector(jQuery(getSourceDocInput()).val());
```
davidmason: - [ ] unnecessary jQuery wrapping

zanata-war/src/main/webapp/iteration/view.xhtml
```diff
@@ -84,16 +84,31 @@
...
+
+  function getSourceDocInput() {
+    return jQuery('#' + jqSelector('#{rich:clientId('specificDocFileInput')}'));
+  }
+
+  function getUploadSourceDocButton() {
+    return jQuery('#' + jqSelector('#{rich:clientId('specificDocSubmitUploadButton')}'));
+  }
+
+  function toggleUploadSourceButton(enable) {
```
davidmason: - [ ] suggest `setUploadSourceButtonEnabled(boolean)` so that it is clear what it does.

zanata-war/src/main/webapp/iteration/view.xhtml
```diff
@@ -84,16 +84,31 @@
...
+    var uploadButton = getUploadSourceDocButton();
+    jQuery(uploadButton).prop('disabled', !enable);
```
davidmason: - [ ] unnecessary jQuery wrapping

zanata-war/src/main/java/org/zanata/adapter/GettextAdapter.java
aeng: We can format anyway we want it to be but this is the exception format we used throughout the app.

zanata-war/src/main/java/org/zanata/adapter/OkapiFilterAdapter.java
aeng: I prefer it to be `transformToMapByRestId`

aeng:
Yes. The upload file options in GWT is not even in sync with options we provided in JSF upload.
I think its best to have a centralised location for uploading files(from JSF page).  

Regards, 

Alex Eng 
Senior Software Engineer 
DID: +61 3514 8262 
Mobile: +614 2335 3457 
Email: aeng@redhat.com 

Red Hat, Asia-Pacific Pty Ltd 
Level 1, 193 North Quay 
Brisbane 4000 
Office: +61 7 3514 8100 
Fax: +61 7 3514 8199 
Website: www.redhat.com 

----- Original Message -----

zanata-war/src/main/java/org/zanata/util/FileUtil.java
aeng: I think name says it all, FileUtil. 
I can list out all the implemented method there but it will be redundant information.

zanata-war/src/main/java/org/zanata/servlet/MultiFileUploadServlet.java
aeng: As mentioned earlier, multifile upload doesn't support docType selection yet if the file extension found in multiple doc type. We are simply returning the first docType found with the extension. This doesn't change the current multifile upload behaviour.

zanata-war/src/main/java/org/zanata/servlet/MultiFileUploadServlet.java
aeng: https://bugzilla.redhat.com/show_bug.cgi?id=1217671

zanata-war/src/main/java/org/zanata/servlet/FileUploadServlet.java
aeng: Removed upload support from GWT as it doesn't sync with the options in JSF. And its best to have 1 single place for uploading file.

zanata-war/src/main/java/org/zanata/service/impl/TranslationFileServiceImpl.java
aeng: Multi file upload uses this service as well and the docType selection is not implemented in there yet.

zanata-war/src/main/java/org/zanata/service/impl/TranslationFileServiceImpl.java
aeng: Checkstyle passed. I will make the change manually.

zanata-war/src/main/java/org/zanata/service/TranslationFileService.java
aeng: Updated to `hasMultipleDocumentTypes` which what is it.

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
aeng: `Nonnull` annotation doesn't throw any exception even value is null. Its just like a remainder to the caller that this value should be check beforehand. You will still need to check for null.

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
aeng: the en-us is source locale.

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
aeng: yeah. there's no use for it.

zanata-war/src/main/java/org/zanata/adapter/PropertiesLatinOneAdapter.java
```diff
@@ -0,0 +1,36 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this software; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA, or see the FSF
+ * site: http://www.fsf.org.
+ */
+
+package org.zanata.adapter;
+
+/**
+ * Adapter for read and write {@link org.zanata.common.DocumentType#PROPERTIES} file
```
davidmason: - [ ] (grammar) "... for reading and writing ..." or "... to read and write ..."

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
davidmason: I must be thinking of something with Lombok that adds the check. Still, non-null values are clearly required here so it would be best to have the annotations to provide that information in the method signature.

zanata-war/src/main/java/org/zanata/adapter/PropertiesAdapter.java
davidmason: > the en-us is source locale.

So if someone wants to use a Dutch properties source file, it will be impossible. Is that the correct implementation?

zanata-war/src/main/java/org/zanata/adapter/GenericPropertiesAdapter.java
```diff
@@ -0,0 +1,204 @@
...
+public class GenericPropertiesAdapter implements FileFormatAdapter {
+
+    public static final String ISO_8859_1 = Charsets.ISO_8859_1.name();
+
+    public static final String UTF_8 = Charsets.UTF_8.name();
+
+    private final String charset;
+
+    public GenericPropertiesAdapter(String charset) {
+        // check that charset is valid, throw exception if not
```
davidmason: - [ ] Is this a *TODO*? Seems like nothing is blocking it.

zanata-war/src/main/java/org/zanata/adapter/GenericPropertiesAdapter.java
```diff
@@ -0,0 +1,204 @@
...
+            return PropWriter.CHARSET.UTF8;
+        } else {
+            throw new FileFormatAdapterException(
+                "Not supported charset '" + charset + "' for properties file");
+        }
+    }
+
+    private void writeTranslatedFile(OutputStream output, Resource resource,
+        TranslationsResource translationsResource, String locale,
+        Optional<String> params, String charset)
```
davidmason: - [ ] No need to pass `charset` in here, it is a field.

zanata-war/src/main/java/org/zanata/adapter/GenericPropertiesAdapter.java
davidmason: Easy fix for this: change `charset` to type `PropWriter.CHARSET`, then in this constructor do `this.charset = toPropWriterCharset(charset);`.

That will do the check, and it means the code below can use `charset` directly.

davidmason: If you do that, please also move `toPropWriterCharset` just below the constructor, since that is the only place it is used.

zanata-war/src/main/java/org/zanata/adapter/GettextAdapter.java
```diff
@@ -0,0 +1,171 @@
...
+import org.zanata.util.FileUtil;
+import org.zanata.util.ServiceLocator;
+
+import com.google.common.base.Charsets;
+import com.google.common.base.Optional;
+import com.google.common.collect.Sets;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Adapter for reading and write {@link org.zanata.common.DocumentType#GETTEXT_PORTABLE_OBJECT}
```
davidmason: - [ ] .(grammar) "... for reading and writing ..." or "... to read and write ..."

zanata-war/src/main/java/org/zanata/adapter/PropertiesUTF8Adapter.java
```diff
@@ -0,0 +1,37 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this software; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA, or see the FSF
+ * site: http://www.fsf.org.
+ */
+
+package org.zanata.adapter;
+
+/**
+ * Adapter for reading and write
```
davidmason: - [ ] .(grammar) "... for reading and writing ..." or "... to read and write ..."

zanata-war/src/main/java/org/zanata/adapter/XliffAdapter.java
```diff
@@ -0,0 +1,130 @@
...
+import org.zanata.common.LocaleId;
+import org.zanata.exception.FileFormatAdapterException;
+import org.zanata.rest.dto.resource.Resource;
+import org.zanata.rest.dto.resource.TranslationsResource;
+
+import com.google.common.base.Optional;
+import org.zanata.util.FileUtil;
+
+/**
+ * Adapter for read and write {@link org.zanata.common.DocumentType#XLIFF} file
```
davidmason: - [ ] .(grammar) "... for reading and writing ..." or "... to read and write ..."

zanata-war/src/main/java/org/zanata/adapter/GenericPropertiesAdapter.java
```diff
@@ -0,0 +1,204 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URL;
+
+/**
+ * Properties file adapter for read and write.
```
davidmason: - [ ] .(grammar) "... for reading and writing"

davidmason:
- [ ] I recommend doing Ctrl+F for "if(" on https://github.com/zanata/zanata-server/pull/772/files to find all the if statements that are not followed with space. It finds 10 of them at the moment.

zanata-war/src/main/java/org/zanata/servlet/FileUploadServlet.java
davidmason: Just need to make sure that product owner / UX are ok with this, since it is not part of the scope for this feature.

aeng:
We should enforce this rule if we're all agree. Otherwise, I wouldn't worry about style that are not part of Zanata.

zanata-war/src/main/java/org/zanata/util/FileUtil.java
davidmason: - [ ] I mean for the comment in line 31, 32 etc. (now 43-48), so my comment applies to `public static String extractExtension(String fileNameOrExtension)`.

davidmason:
> We should enforce this rule if we're all agree. Otherwise, the same style would appear over again. At the moment, I wouldn't worry about having spacing or not.

I think we will turn this rule on when we can. The style we use for all if statements is with a space, so better to fix it in this PR so there is less to fix up when we turn on the rule.

aeng:
> I think we will turn this rule on when we can. The style we use for all if statements is with a space, so better to fix it in this PR so there is less to fix up when we turn on the rule.

Done.

zanata-war/src/main/java/org/zanata/servlet/FileUploadServlet.java
aeng: Got ok from @carlosmunoz  and michelle.

davidmason:
> >I think we will turn this rule on when we can. The style we use for all if statements is with a space, so better to fix it in this PR so there is less to fix up when we turn on the rule.
>
> Done.

Is is committed yet?

aeng:
> Is is committed yet?

Yup.

davidmason:
:+1:

definite:
When pulling with locale that has "@" (e.g. zh@classic),  server return 500
Caused by: java.lang.IllegalArgumentException: The locale identifier 'zh@classic' contains at least one invalid character.

aeng:
> When pulling with locale that has "@" (e.g. zh@classic), server return 500
Caused by: java.lang.IllegalArgumentException: The locale identifier 'zh@classic' contains at least one invalid character.

This is an existing issue. I will file a new bug against it.

aeng:
> When pulling with locale that has "@" (e.g. zh@classic), server return 500
Caused by: java.lang.IllegalArgumentException: The locale identifier 'zh@classic' contains at least one invalid character.

https://bitbucket.org/okapiframework/okapi/issue/459/localeid-doesnt-supported-locale-variant

definite:
The locale variant support should be addressed in https://bugzilla.redhat.com/show_bug.cgi?id=1220239

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/847501d1e1b78a70187362f6e4dd285638b392d5

PR #773, opened 2015-04-13
Ensure use of InnoDB engine on MySQL

seanf:
https://bugzilla.redhat.com/show_bug.cgi?id=1207575

- [x] Might add precondition changeset (always runs) which checks that all tables are InnoDB

- [x] Wait for https://github.com/zanata/zanata-server/pull/791

### DIFF
--- a/zanata-war/src/main/java/liquibase/sqlgenerator/ext/CreateTableGeneratorInnoDB.java
+++ b/zanata-war/src/main/java/liquibase/sqlgenerator/ext/CreateTableGeneratorInnoDB.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright 2015, Red Hat, Inc. and individual contributors as indicated by the
+ * @author tags. See the copyright.txt file in the distribution for a full
+ * listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation; either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * This software is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this software; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA, or see the FSF
+ * site: http://www.fsf.org.
+ */
+package liquibase.sqlgenerator.ext;
+
+import liquibase.database.Database;
+import liquibase.database.core.MySQLDatabase;
+import liquibase.exception.UnexpectedLiquibaseException;
+import liquibase.exception.ValidationErrors;
+import liquibase.logging.LogFactory;
+import liquibase.logging.Logger;
+import liquibase.sql.Sql;
+import liquibase.sql.UnparsedSql;
+import liquibase.sqlgenerator.SqlGeneratorChain;
+import liquibase.sqlgenerator.core.CreateTableGenerator;
+import liquibase.statement.core.CreateTableStatement;
+import liquibase.structure.DatabaseObject;
+
+import java.util.Collection;
+
+/**
+ * A create table statement generator to force the use of the InnoDB engine
+ * on MySQL/MariaDB, unless the engine has already been specified by other
+ * means (eg modifySql).
+ * <p>
+ * Note: SqlGenerator implementations must live in package
+ * "liquibase.sqlgenerator.ext" to be automatically registered.
+ * </p>
+ *
+ * @author Sean Flanigan <a
+ *         href="mailto:sflaniga@redhat.com">sflaniga@redhat.com</a>
+ */
+@SuppressWarnings("unused")
+public class CreateTableGeneratorInnoDB extends CreateTableGenerator {
+
+    @Override
+    public boolean supports(CreateTableStatement statement, Database database) {
+        // NB this will include MariaDB (with Liquibase 3.4+)
+        return database instanceof MySQLDatabase;
+    }
+
+    @Override
+    public int getPriority() {
+        // Higher priority than the built-ins
+        return PRIORITY_DATABASE + 1;
+    }
+
+    @Override
+    public ValidationErrors validate(CreateTableStatement createTableStatement,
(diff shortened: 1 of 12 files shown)


aeng:
:+1:

seanf:
@carlosmunoz I'd like your opinion on this.

zanata-war/src/main/java/liquibase/sqlgenerator/ext/CreateTableGeneratorInnoDB.java
```diff
@@ -0,0 +1,119 @@
...
+        boolean foundCreateTable = false;
+        Sql[] sqls = sqlGeneratorChain.generateSql(statement, database).clone();
+
+        // Append " ENGINE=INNODB" to CREATE TABLE unless ENGINE already
+        // specified. This is clumsy, but Liquibase doesn't have seem
+        // to have any better hooks for this.
+        for (int i = 0; i < sqls.length; i++) {
+            Sql sql = sqls[i];
+            String sqlText = sql.toSql();
+            String sqlUpper = sqlText.toUpperCase();
```
carlosmunoz: How about a `.trim()` just in case?

zanata-war/src/main/resources/db/changelogs/db.changelog-3.6.xml
```diff
@@ -30,4 +30,116 @@
...
+    <preConditions onFail="MARK_RAN">
+      <sqlCheck expectedResult="0">
+        SELECT COUNT(*) FROM INFORMATION_SCHEMA.TABLES
+        WHERE TABLE_SCHEMA = SCHEMA()
+        AND ENGINE = 'InnoDB'
+        AND TABLE_NAME='Activity';
+      </sqlCheck>
+    </preConditions>
+    <comment>Ensure InnoDB for Activity</comment>
+    <sql>ALTER TABLE Activity ENGINE = INNODB</sql>
```
carlosmunoz: Why the special check here?

zanata-war/src/main/resources/db/changelogs/db.changelog-validations.xml
```diff
@@ -0,0 +1,18 @@
...
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd">
+
+  <changeSet author="sflaniga@redhat.com" id="check-innodb" runAlways="true"
+    dbms="mysql">
+    <preConditions onFail="HALT">
+      <sqlCheck expectedResult="0">
+        SELECT count(*) FROM information_schema.tables
+        WHERE TABLE_SCHEMA = SCHEMA() AND engine != 'InnoDB'
```
carlosmunoz: :+1:

carlosmunoz:
Looks good in general terms. We are making some assumptions about the sql string, but I guess the check that you've added can raise a red flag if there are non-InnoDB tables in the midst.

zanata-war/src/main/java/liquibase/sqlgenerator/ext/CreateTableGeneratorInnoDB.java
seanf: Good idea.

zanata-war/src/main/resources/db/changelogs/db.changelog-3.6.xml
seanf: Well,

    

None of the other tables was big enough to concern me, but I thought it was *possible* that this conversion might take a while, so the precondition lets us skip it if we don't need it.  

Only on some environments have we been creating MyISAM tables; others are using InnoDB everywhere.  I suspect it is MySQL 5.1 that doesn't (always?) respect the session variable.

seanf:
@carlosmunoz note that this is targeting `release`, if that makes any difference.

djansen-redhat:
:white_check_mark: Tested update and halt on !InnoDB

seanf:
@djansen-redhat @aeng @carlosmunoz I have added a custom error message for the precondition check.

seanf:
When I tested it again, I couldn't find any evidence that Liquibase was picking up the "ensure InnoDB" extension(!).  I had to move it to its own jar before Liquibase would register it.

I've also added another extension which logs before executing each change, which should help a little with https://bugzilla.redhat.com/show_bug.cgi?id=1207980

djansen-redhat:
Is there a bug report related to this?

seanf:
> Is there a bug report related to this?

https://bugzilla.redhat.com/show_bug.cgi?id=1207575

aeng:
:+1: on https://github.com/zanata/zanata-server/commit/6b111cc5adf6299615370d8f2fe0068963aa2797

seanf:
@carlosmunoz How do you feel about the new module under zanata-server, ie zanata-liquibase?

I'm hoping we'll be able to move our changelogs and custom changes into that module too, which would make it easier to run liquibase from the command line.

carlosmunoz:
I don't mind it... although it looks small for the time being.

If we end up moving the changelogs, will liquibase be able to read them from inside a jar? or will we need two artifacts. In the later case, I would rather keep the changelogs where they currently reside.

seanf:
On 2015-04-23 00:53, Carlos Munoz wrote:
> I don't mind it... although it looks small for the time being.
> 
> If we end up moving the changelogs, will liquibase be able to read them
> from inside a jar? or will we need two artifacts. In the later case, I
> would rather keep the changelogs where they currently reside.

Yes, liquibase is supposed to be able to read changelogs from a jar.
The only question will be how many classes our custom changes depend on,
because they may have to be moved out of zanata.war into a jar.


-- 
Sean Flanigan

Principal Software Engineer
Red Hat

djansen-redhat:
Test:
- [ ] Move !InnoDB tables to InnoDB on changeset is run
- [ ] Abort on !InnoDB tables existing after changeset is run
- [ ] Warn on !InnoDB non-Zanata tables existing after changeset is run?

seanf:
@djansen-redhat: regarding your checklist:

1. only for tables created since the last time we cleaned this up.  So this doesn't apply to HTextFlow, for instance.
2. correct (assuming they are in the same schema as Zanata's tables)
3. no, we abort if the non-InnoDB tables are in the same schema, otherwise they are ignored.

djansen-redhat:
:white_check_mark: Tested

seanf:
@carlosmunoz are you happy with c69d8ec?

carlosmunoz:
:+1:

seanf:
I need to double check that the new extension is indeed logging changeset IDs before executing them (unless someone beats me to it).

PR #775, opened 2015-04-16
Separate wildfly client version from server version

seanf:
Replaces https://github.com/zanata/zanata-server/pull/763


davidmason:
:+1: :shipit:

PR #776, opened 2015-04-17
Disable implicit wait in Selenium tests

seanf:
We will have to add some explicit waits (before/around checks) to make this work, but the end result should be a lot more robust.

- [x] Waiting for https://github.com/zanata/zanata-server/pull/779
- [x] Waiting for https://github.com/zanata/zanata-server/pull/782

### DIFF
--- a/functional-test/pom.xml
+++ b/functional-test/pom.xml
@@ -99,7 +99,7 @@
     <dependency>
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
-      <scope>test</scope>
+      <scope>compile</scope>
     </dependency>
 
     <dependency>
--- a/functional-test/pom.xml
+++ b/functional-test/pom.xml
@@ -50,7 +50,7 @@
     <webdriver.display>:0</webdriver.display>
     <webdriver.log>${project.build.directory}/webdriver.log</webdriver.log>
     <webdriver.screenshot.dir>${project.build.directory}/screenshots</webdriver.screenshot.dir>
-    <webdriver.wait>10</webdriver.wait>
+    <webdriver.wait>20</webdriver.wait>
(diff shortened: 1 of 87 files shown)


djansen-redhat:

seanf:
Rebuilding with all functional tests, now that #779 is in release.

seanf:
Waiting for the `form__loader` fix in assets.

seanf:
See also http://assets-zanata.rhcloud.com/6.1.1/styleguide/

seanf:
See also https://github.com/zanata/zanata-server/pull/782

seanf:

seanf:

seanf:
We should retest after zanata/zanata-assets#3 is merged and deployed.

seanf:
Rebase and retest after https://github.com/zanata/zanata-server/pull/782 is merged.

seanf:
If it turns out we can't turn off implicit waits completely (or not yet), I suggest we implement our own versions of `findElement(By by)` and `findElements(By by)` which will not perform implicit waits, but will simply use JavaScript DOM lookups (probably with help from jQuery).

This is a list of the By selector types we currently use:

	$ sed -n 's/.*By[.]\([a-zA-Z]*\).*/\1/p' src/**/*.java | sort -u
	className
	cssSelector
	id
	linkText
	name
	tagName
	xpath

We would need to implement support for some or all of the above `By` types, but we could skip the others (like partialLinkText).

seanf:
If we end up implementing our own `findElement(By by)` and `findElements(By by)`, @davidmason suggested that we use the following functions rather than depend on jQuery (at least for the simpler `By` types):

 - https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll
 - https://developer.mozilla.org/en-US/docs/Web/API/NodeList/length

djansen-redhat:
Replaced by #804

PR #779, opened 2015-04-20
Wait for page loaders to be finished before proceeding

djansen-redhat:
Requires the class 'is-active' to be present in the web element,
as the loader is always present and visible.

### DIFF
--- a/functional-test/src/main/java/org/zanata/page/AbstractPage.java
+++ b/functional-test/src/main/java/org/zanata/page/AbstractPage.java
@@ -261,7 +261,8 @@ public class AbstractPage {
                 if (outstanding == null) {
                     if (log.isWarnEnabled()) {
                         String url = getDriver().getCurrentUrl();
-                        String pageSource = ShortString.shorten(getDriver().getPageSource(), 2000);
+                        String pageSource = ShortString.shorten(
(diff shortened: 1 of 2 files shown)


seanf:
Looks good, but I would suggest that `waitForPageSilence()` should automatically wait for loaders to finish too.  And then perhaps `waitForLoadersFinished()` should be private.

djansen-redhat:
Updated to address the above

functional-test/src/main/java/org/zanata/page/AbstractPage.java
```diff
@@ -53,6 +53,7 @@ public AbstractPage(final WebDriver driver) {
         this.driver = driver;
         assert driver instanceof JavascriptExecutor;
         waitForPageSilence();
+        waitForLoaders();
```
seanf: I think this is redundant now.

djansen-redhat: Oops, thanks

seanf:
:+1: Everything else looks good.

seanf:
Back-ported in commit 8b3cc44a71879fe89252ffdaff680842da5506af.

PR #784, opened 2015-04-27
Update to WildFly 9.0.0.CR1

seanf:
Disable jsessionid tracking to fix WildFly 9 problem: specifically, the problem where clicking "login" tries to take you to the login form, but with an appended jsessionid parameter, which leads to a 404.

https://bugzilla.redhat.com/show_bug.cgi?id=1197955

PR #786, opened 2015-04-28
Remove jsessionid and upgrade Undertow for WildFly 8.2

seanf:
Based on https://github.com/zanata/zanata-server/pull/784 but using WildFly 8.2 combined with Undertow modules from WildFly 9 (because proper WildFly 9 still has some problems).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1220,6 +1220,7 @@
         <mojarra.module.name>wildfly-${module.wildfly.version}-module-mojarra-${mojarra.module.version}.zip</mojarra.module.name>
         <hibernate.module.version>4.2.15.Final</hibernate.module.version>
         <hibernate.module.name>wildfly-${module.wildfly.version}-module-hibernate-main-${hibernate.module.version}.zip</hibernate.module.name>
+        <undertow.module.name>wildfly-9.0.0.Beta2-modules-undertow.zip</undertow.module.name>
       </properties>
 
       <build>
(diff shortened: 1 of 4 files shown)


seanf:
The full functional tests actually passed with this Frankenstein combination.

carlosmunoz:
So...
1. What is the problem with Wildfly 9?
2. Does this mean that we are targeting Wildfly 8 with a modified Undertow module?

seanf:
On 2015-04-28 15:36, Carlos Munoz wrote:
> So...
> 1. What is the problem with Wildfly 9?

Specifically, the problem where clicking "login" tries to take you to
the login form, but with an appended jsessionid parameter, which leads
to a 404.


> 2. Does this mean that we are targeting Wildfly 8 with a modified
> Undertow module?

Yes, this PR changes the default WildFly to be 8.2.0.Final, with some
extra Undertow 1.2 modules which were taken from WildFly 9.


-- 
Sean Flanigan

Principal Software Engineer
Red Hat

seanf:
It actually seems to work, but given the unknown risks in upgrading Undertow this way, we're going to skip WildFly 8.2.0.Final and go with WildFly 9 beta (at least for testing) instead: https://github.com/zanata/zanata-server/pull/784

I also cherry picked b168cf687bc63aab00522b7d9398053bdaab9b76 as https://github.com/zanata/zanata-server/pull/789.

PR #787, opened 2015-04-28
rhbz903964 - produce better error message when permission check failed

huangp:
https://bugzilla.redhat.com/show_bug.cgi?id=903964

For example, when invoke client with following command:
```z  -B push --push-type trans --merge-type import```

There are two(maybe more) possible permission failure:
* user is not part of the language team
* user don't have permission to do merge type import operation

In the past what we get back is just a 401 with no body in response. Now we can tell what exactly go wrong:
* when user is not a member for the locale
```
[ERROR] Execution failed: 
 * Error Message: PUT http://localhost:8080/zanata/rest/async/projects/p/jasper/iterations/i/1/r/MessageResources/translations/de?ext=comment&merge=import&assignCreditToUploader=false returned a response status of 401 Unauthorized;
 * Response From Server: Failed to obtain permission(modify-translation) with following facts([Locale(id=de, name=German), Project(name=jasper, slug=jasper, status=ACTIVE)])]
```

* when user is a translator but not project maintainer (no permission to do merge import)
```
[ERROR] Execution failed: Failed while pushing document translations: [Failed to obtain permission(import-translation) with following facts([Project version(slug=1, status=ACTIVE)])]
```

### DIFF
--- a/zanata-model/src/main/java/org/zanata/model/HLocaleMember.java
+++ b/zanata-model/src/main/java/org/zanata/model/HLocaleMember.java
@@ -47,7 +47,7 @@ import org.apache.commons.lang.builder.HashCodeBuilder;
 @Table(name = "HLocale_Member")
 @Setter
 @NoArgsConstructor
-public class HLocaleMember implements Serializable {
+public class HLocaleMember implements Serializable, HasUserFriendlyToString {
     private static final long serialVersionUID = 1L;
 
     private HLocaleMemberPk id = new HLocaleMemberPk();
@@ -56,6 +56,8 @@ public class HLocaleMember implements Serializable {
     private boolean isReviewer;
     private boolean isTranslator;
 
+    private transient String userFriendlyToString;
+
     public HLocaleMember(HPerson person, HLocale supportedLanguage,
             boolean isTranslator, boolean isReviewer, boolean isCoordinator) {
         id.setPerson(person);
@@ -63,6 +65,11 @@ public class HLocaleMember implements Serializable {
         setTranslator(isTranslator);
         setReviewer(isReviewer);
         setCoordinator(isCoordinator);
+        userFriendlyToString =
+                String.format(
+                        "Locale membership(person=%s, isTranslator=%s, isReviewer=%s, isCoordinator=%s)",
+                        person.getName(), isTranslator, isReviewer,
+                        isCoordinator);
     }
 
     @EmbeddedId
@@ -97,6 +104,11 @@ public class HLocaleMember implements Serializable {
     @Transient
     public HLocale getSupportedLanguage() {
         return id.getSupportedLanguage();
+    }
+
+    @Override
+    public String userFriendlyToString() {
+        return userFriendlyToString;
     }
 
     @Embeddable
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: zanata-model/src/main/java/org/zanata/model/HLocaleMember.java [hunk 1]
```diff
@@ -63,6 +65,11 @@ public HLocaleMember(HPerson person, HLocale supportedLanguage,
         setTranslator(isTranslator);
         setReviewer(isReviewer);
         setCoordinator(isCoordinator);
+        userFriendlyToString =
+                String.format(
+                        "Locale membership(person=%s, isTranslator=%s, isReviewer=%s, isCoordinator=%s)",
+                        person.getName(), isTranslator, isReviewer,
```
davidmason: This is user display name. Username might be more useful to show instead or as well.

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
```diff
@@ -141,12 +144,12 @@ public boolean hasPermission(String name, String action, Object... arg) {
         if (result) {
             if (log.isDebugEnabled()) {
                 log.debug("ALLOWED hasPermission({}, {}, {}) for user {}",
-                        name, action, arg, getAccountUsername());
+                        name, action, Lists.newArrayList(arg), getAccountUsername());
             }
         } else {
             if (log.isDebugEnabled()) {
```
davidmason: Does this check even do anything? I thought that `log.debug(...)` would implicitly perform this check anyway. Is it a performance optimization to prevent the creation of the new array list for `arg`?

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
```diff
@@ -182,7 +185,27 @@ public boolean hasPermission(String action, Object... targets) {
...
+                    log.warn(
+                            "target [{}] may not have user friendly string representation",
+                            target.getClass());
+                    meaningfulTargets.add(target.toString());
+                }
+            }
+            throw new AuthorizationException(
+                    String.format(
+                            "Failed to obtain permission('%s') with following facts(%s)",
+                            action, meaningfulTargets));
```
davidmason: Is `facts` technical jargon here about the permission checking? For users it might be better to say "information", it is probably fine though as long as there is enough in the detail.

zanata-model/src/main/java/org/zanata/model/HLocaleMember.java
huangp: username is in HAccount. Although HPerson has fetch=eager annotation on account field, you can always override this. I don't want to cause an extra database call if it's not there.

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
huangp: I think originally it was to prevent the getAccountUser() method call if it's not in debug mode. Now it will also prevent the creation of the list.

davidmason:
:+1:

PR #789, opened 2015-04-30
Add server logging for failure responses

seanf:


huangp:
:+1:

zanata-war/src/main/java/org/zanata/rest/editor/service/TranslationService.java
```diff
@@ -80,7 +82,10 @@ public Response get(String localeId, String ids) {
         }
         List<Long> idList = TransUnitUtils.filterAndConvertIdsToList(ids);
         if (idList.size() > TransUnitUtils.MAX_SIZE) {
-            return Response.status(Response.Status.FORBIDDEN).build();
+            String msg = String.format("More than %d results.", TransUnitUtils.MAX_SIZE);
+            log.warn(msg);
+            return Response.status(Response.Status.FORBIDDEN).entity(msg)
```
djansen-redhat: Why did this line wrap, but not the one 2 lines above it? Weird.

djansen-redhat:
Any hints on how to trigger these?

seanf:
> Any hints on how to trigger these?

You might be able to use a custom client which sent an invalid request (too many transunit IDs), or perhaps send an illegal document for translation, but otherwise it would require a unit test with mock objects or something like Byteman which can inject exceptions: http://byteman.jboss.org/.  All of which is probably overkill for such a simple change.

djansen-redhat:
Agreed.

PR #796, opened 2015-05-06
Rhbz1211849 change slug

huangp:
https://bugzilla.redhat.com/show_bug.cgi?id=1211849

Things to test (feel free to add more)
- [x] project slug can be changed
- [x] version slug can be changed
- [x] change slug on project or version page as well as other general settings won't trigger exception
- [x] project can be searched by new slug after change
- [x] old slug is available to use immediately after change
- [x] after project slug change, TM merge with project match set to reject can still find matches from other version
- [x] when a project or version slug has changed, existing open editor will receive notification and refresh to new URL
- [x] ~~when a project or version slug has changed, existing open editor will automatically save unsaved work as fuzzy before page refreshed to new URL~~ RFE


zanata-model/src/main/java/org/zanata/hibernate/search/ContainingWorkspaceBridge.java
```diff
@@ -42,8 +42,6 @@ public void set(String name, Object value, Document luceneDocument,
 
         addStringField(IndexFieldLabels.PROJECT_FIELD, project.getSlug(),
                 luceneDocument, luceneOptions);
-        addStringField(IndexFieldLabels.ITERATION_FIELD, iteration.getSlug(),
-                luceneDocument, luceneOptions);
```
carlosmunoz: Why are we getting rid of the indexing by slug? As far as I'm concerned we still need it.

zanata-war/src/main/java/org/zanata/action/ProjectHome.java
```diff
@@ -786,8 +800,6 @@ public void updateStatus(char initial) {
                 if (version.getStatus() != EntityStatus.OBSOLETE) {
                     version.setStatus(EntityStatus.OBSOLETE);
                     entityManager.merge(version);
-                    projectIterationUpdateEvent.fire(
-                            new ProjectIterationUpdate(version));
```
carlosmunoz: Were these events unused?

zanata-war/src/main/java/org/zanata/events/ProjectUpdate.java
```diff
@@ -35,4 +35,6 @@
     public static final String EVENT_NAME = "org.zanata.events.ProjectUpdate";
 
     HProject project;
+
+    String oldSlug;
```
carlosmunoz: Mmmm.... the events were removed, but the event classes are being modified here. I might be missing some new places where they are being used, but just in case adding this comment.

zanata-model/src/main/java/org/zanata/hibernate/search/ContainingWorkspaceBridge.java
huangp: We don't use iteration slug in any lucene query. We used to use it in project wide search but then we switched to use HQL instead. Do you know any where else it is used?

zanata-war/src/main/java/org/zanata/action/ProjectHome.java
huangp: It has moved to SlugEntityUpdateListener, the hibernate event interceptor. Here it will only fire event when you update project in JSF but not from REST. I think it will be more robust to put it in the new place.

zanata-war/src/main/java/org/zanata/events/ProjectUpdate.java
huangp: Yes see above. I needed the old slug so that I can look up existing workspace for editor (the key in hashmap uses old slug) and send events to them.

zanata-model/src/main/java/org/zanata/hibernate/search/ContainingWorkspaceBridge.java
carlosmunoz: Does TM merge use it? (or will it potentially need to in the future?)

zanata-war/src/main/java/org/zanata/action/ProjectHome.java
carlosmunoz: Got it. I agree. :+1:

zanata-war/src/main/java/org/zanata/events/ProjectUpdate.java
carlosmunoz: Yes, that clears that up. Events are still used, just in the hibernate listener now.

zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java
```diff
@@ -60,4 +68,62 @@ protected ScrollableResults queryResults(int ignoredOffset, FullTextSession sess
...
+        // it must use the same session in the DAO and to do the indexing
+        HTextFlowTargetStreamingDAO dao =
+                new HTextFlowTargetStreamingDAO(HTextFlowTarget.class, session);
+        ScrollableResults scrollableResults =
+                dao.getTargetsWithAllFieldsEagerlyFetchedForProject(
+                        project);
+        reindexScrollableResultSet(session, scrollableResults, handle);
+    }
+
+    private static void reindexScrollableResultSet(FullTextSession session,
```
carlosmunoz: In order for this method to be asynchronous it needs to be annotated with `@Async` and the Seam component needs to be annotated with `@ContainsAsyncMethods`. The method can return a `java.util.concurrent.Future` or void. For an example, take a look at `AsyncTaskITCase`

zanata-war/src/main/java/org/zanata/service/IndexingService.java
```diff
@@ -36,4 +38,12 @@
             Map<Class<?>, ReindexClassOptions> indexingOptions,
             AsyncTaskHandle<Void> handle)
             throws Exception;
+
+    /**
+     * This will re-index all HTextFlowTargets under a given project.
+     */
+    @Async
+    Future<Void> reindexHTextFlowTargetssForProject(HProject hProject,
```
carlosmunoz: There is an extra 's' (Targetss).

zanata-war/src/main/java/org/zanata/service/impl/SecurityServiceImpl.java
```diff
@@ -81,6 +81,12 @@ public HProject checkWorkspaceStatus(WorkspaceId workspaceId) {
                         .getProjectIterationId().getProjectSlug(), workspaceId
                         .getProjectIterationId().getIterationSlug());
 
+        if (project == null || projectIteration == null) {
+            // TODO pahuang due to slug change or in future project being deleted permanently
+            // see org.zanata.webtrans.server.SeamDispatch.execute()
+            // for a valid exception to throw
+            throw new AuthorizationException("Project or version has moved");
```
carlosmunoz: ... or project does not exist :smile:

zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java
```diff
@@ -0,0 +1,115 @@
...
+import org.zanata.model.SlugEntityBase;
+import org.zanata.service.IndexingService;
+import org.zanata.util.Event;
+
+/**
+ * This class is a hibernate event listener which listens on post commit events
+ * for HProject and HProjectIteration. If it detects a HProject slug change, it
+ * will perform re-indexing for all HTextFlowTargets under that project. It will
+ * also fire update event for HProject and HProjectIteration with their old slug
+ * in payload.
```
carlosmunoz: :+1: :clap:

zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java
```diff
@@ -0,0 +1,115 @@
...
+ * @see IndexingServiceImpl
+ *
+ * @author Patrick Huang <a
+ *         href="mailto:pahuang@redhat.com">pahuang@redhat.com</a>
+ */
+@Name("slugEntityUpdatedListener")
+@Scope(ScopeType.APPLICATION)
+@AutoCreate
+@Slf4j
+public class SlugEntityUpdatedListener implements PostUpdateEventListener {
```
carlosmunoz: This is a Seam component that is also a Hibernate listener. How do those two play together?

zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * @see org.zanata.webtrans.server.HibernateIntegrator
+ * @see org.zanata.webtrans.server.TranslationWorkspaceManagerImpl
+ * @see IndexingServiceImpl
+ *
+ * @author Patrick Huang <a
+ *         href="mailto:pahuang@redhat.com">pahuang@redhat.com</a>
+ */
+@Name("slugEntityUpdatedListener")
+@Scope(ScopeType.APPLICATION)
```
carlosmunoz: This is application scoped, but it could potentially be getting concurrent calls. Lets make sure there are no concurrency problems. Right off the bat I can see a `private Integer index` field that might be dangerous.

In terms of injected components, everything must either be an application or stateless scoped bean (which I think they are).

zanata-war/src/main/java/org/zanata/search/HTextFlowTargetIndexingStrategy.java
huangp: Just realize when I checked the code. This method is not async but the caller is: org.zanata.service.impl.IndexingServiceImpl#reindexHTextFlowTargetsForProject

I found a typo in the method name above. I will update it.

zanata-war/src/main/java/org/zanata/service/IndexingService.java
huangp: Yup

zanata-war/src/main/java/org/zanata/webtrans/server/HibernateIntegrator.java
```diff
@@ -37,6 +38,8 @@ public void integrate(Configuration configuration,
                     updateListener);
             eventListenerRegistry.appendListeners(EventType.POST_INSERT,
                     updateListener);
+            eventListenerRegistry.appendListeners(EventType.POST_COMMIT_UPDATE,
+                    ServiceLocator.instance().getInstance(SlugEntityUpdatedListener.class));
```
carlosmunoz: I guess this answers my previous question about how they play together.

But this gives me an idea, instead of having a single Application scoped bean and therefore worry about concurrency and stuff, can we have the listener itself be a POJO that calls a (preferably stateless but not necessarily) seam bean to do the event calling and all?

zanata-war/src/main/java/org/zanata/webtrans/server/TranslationWorkspaceManagerImpl.java
```diff
@@ -172,26 +177,47 @@ protected String getSessionId() {
     }
 
     @Observer(ProjectUpdate.EVENT_NAME)
-    public void projectUpdate(@Observes ProjectUpdate payload) {
-        projectUpdate(payload.getProject());
+    // transaction has already been committed and marked as rolled back only for
+    // current thread. We have to open a new transaction to load any lazy
+    // properties (otherwise exception like javax.resource.ResourceException:
+    // IJ000460: Error checking for a transaction: Transactions are not active)
+    @Async
```
carlosmunoz: I don't think this needs to be `@Async`

zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java
huangp: In HibernateIntegrator, it gets the listener instance by using ServiceLocator. Hibernate doesn't care how you get it, as long as it implements the interface.

zanata-war/src/main/java/org/zanata/service/impl/SlugEntityUpdatedListener.java
huangp: The index itself is not a problem. But it does reminded me a potential bug. It used to only handle HProject but then it changed to also handle HProjectIteration. it needs two indexes instead of one. I will fix it.

zanata-war/src/main/java/org/zanata/webtrans/server/HibernateIntegrator.java
huangp: Do you mean inside the POJO listener, use ServiceLocator to get access to seam beans?

zanata-war/src/main/java/org/zanata/webtrans/server/TranslationWorkspaceManagerImpl.java
huangp: It must be in a separate thread to be able to get a new transaction (the reason is in the comment). Because the event is fired in post update commit hibernate listener (which means the transaction in current thread is committed and you can not do anything to get access database again. Unless there is other way to get a new transaction or reuse this committed transaction, this is the only way to let it start a new transaction.

zanata-war/src/main/resources/messages.properties
```diff
@@ -248,7 +248,7 @@ jsf.ProjectId=Project ID
 jsf.ProjectDescription=Project Description
 jsf.ProjectType=Project Type
 jsf.ProjectType.title=Determines how the project is treated for upload and download by clients or through the website.
-jsf.SlugExample=Example: my-project. This can not be changed.
+jsf.SlugExample=Example: my-project. This will become part of your project's URL. Changing it may break people's bookmark.
```
carlosmunoz: "Changing it may break other people's bookmarks to the project"

carlosmunoz:
Done with review. Should this kind of change make sense to put in a functional test (as much as I dislike those)? Just to test that renaming the project or version redirects you to the right page...

zanata-war/src/main/java/org/zanata/webtrans/server/HibernateIntegrator.java
carlosmunoz: That is correct. The POJO would use the service locator to get a stateless bean and execute it's logic.

zanata-war/src/main/java/org/zanata/webtrans/server/TranslationWorkspaceManagerImpl.java
carlosmunoz: I thought that just by using the `Work` class you could do it in a new transaction. Not sure that you need a new thread for it to happen.

Also, I think you are able to fire events to be executed asynchronously (i.e new thread).

zanata-war/src/main/java/org/zanata/webtrans/server/HibernateIntegrator.java
huangp: What's the benefit this will give us versus making it an application scoped bean? The POJO will be a singleton as well (hibernate will only use one instance of a listener for all its events AFAIK)

zanata-war/src/main/java/org/zanata/webtrans/server/TranslationWorkspaceManagerImpl.java
huangp: Just the Work class will still give me exceptions. I think it will try to reuse previous transaction. Not sure is this is a JEE behavior or seam behavior.

carlosmunoz: Seam probably. We should keep an eye on this then when we migrate to CDI.

zanata-war/src/main/java/org/zanata/webtrans/server/HibernateIntegrator.java
carlosmunoz: It will be a single instance, but if it uses service locator in the scope of each method call, we guarantee that a different bean (because it's stateless) will be used every time the method is invoked, thus avoiding concurrency issues. Application scoped beans are very dangerous as we've come to learn.

carlosmunoz:
:+1: Changes look good.

djansen-redhat:
> make sense to put in a functional test (as much as I dislike those)?
:crying_cat_face:

djansen-redhat:
Changing project version "1.0" to "2.0" - which exists in the same project, causes an exception:
> Caused by: com.mysql.jdbc.exceptions.jdbc4.MySQLIntegrityConstraintViolationException: Duplicate entry '2.0-3060' for key 'slug'

carlosmunoz:
@huangp Added a few comments relating to variable naming and one about resource handling (more importantly).

huangp:
Changing it to originalSlug is fine except I don't think I have reliable tooling support to do it safely. Given the risk I'd leave it as is. 
I will change slugValue to inputSlugValue.
The ```@Slug``` or any bean validation annotation only works if you validate the whole bean (HProject or HProjectIteration). For individual field value, this is what eventually will happen. I don't want to create a bean or use the entity(danger!!)

carlosmunoz:
Got it. :+1:

huangp:
@djansen-redhat autosave when slug changed should have been fixed in last commit.

djansen-redhat:
Fails [EditorOptionsPresenterTest:canSetReadOnlyOnWorkspaceUpdate](http://zanata-build.lab.eng.bne.redhat.com/job/zanata-server-github-pull-requests/3539/org.zanata$zanata-war/testReport/junit/org.zanata.webtrans.client.presenter/EditorOptionsPresenterTest/canSetReadOnlyOnWorkspaceUpdate)

djansen-redhat:
Only exception not dealt with is the clicking of activity links on old version url.  I'm not sure how one can deal with this, other than a dropdown notification handling the NullPointerException from a loadInstance.

djansen-redhat:
:white_check_mark: Tested. Looks good, works well, but also seems very complex...

PR #799, opened 2015-05-08
Declare required resources in jboss-web.xml

seanf:
See also https://github.com/zanata/zanata-server/pull/721


carlosmunoz:
:+1:

seanf:
I think this will be needed to make WildFly 9 (CR1 or later) work without an EAGER cache container.

PR #809, opened 2015-05-14
Add internal version of Seam Home components.

carlosmunoz:
This is a temporary solution to allow for the migration to CDI from Seam.
Once the migration happens, these classes should be considered for refactoring or elimination.


seanf:
Looks fine, but if the files were actually copied from Seam 2.3.1, just say that.

seanf:
:+1:

carlosmunoz:
Squash merged in commit bbca366c2899b2f8224172e558c0c0a799a6f7d6

PR #816, opened 2015-05-17
REST endpoint for editor translation memory

davidmason:
Adds `/suggestions` endpoint with an appropriate DTO for the new editor.
This is the server component of https://bugzilla.redhat.com/show_bug.cgi?id=1209669

This purposely does not re-use all of the existing TM code since there is a bug in how the result objects are collated. I chose to leave most of that code alone except for a little fix, because it is used by copytrans, so is performance-critical. It should be fixed and tested separately from this so that adequate attention can be given to any impact on copytrans performance.
See https://bugzilla.redhat.com/show_bug.cgi?id=1220960

### Expected match similarity behaviour

Note: this is the existing similarity behaviour, but returning 0% instead of `NaN` when there would have been a divide-by-zero condition.

#### Case:

similarity comparisons look like they ignore case.
I guess that the initial match search would also ignore case.

#### Stop-words:

Stop-words do not appear to be considered in the match search (lucene search), but are considered in calculating the similarity percentage, as follows:

 - Stop words are common words like "the", "no", "and", "a". They are specified in [stopwords.txt](https://github.com/zanata/zanata-server/blob/new-tm-endpoints-rhbz1209669/zanata-war/src/main/resources/org/zanata/search/stopwords.txt).
 - Stop words are completely ignored for similarity comparison.
    - "A house" and "No house" should be 100% similar since "A" and "No" are ignored.
 - There is no fallback: strings with only stop words will always show 0% similarity.
    - "No" and "No" will show 0% similarity because "no" is a stop-word.
    - Yes, I do think this is confusing and that it would be better to change it.
    - I had added a fallback, but sflanigan did not agree with that approach. I decided not to change the bahaviour so that a debate would not hold up this PR. I intend to gather data about actual user expectations to avoid intuition-based debate on the issue, then make a new PR with similarity scores that look sensible to users.

#### Punctuation:

(No change from how it was treated before): characters are only considered punctuation if they

 - are '.' OR ','
 - AND are followed by 1 or more whitespace characters.

I had changed this, but rolled it back due to some concerns that were brought up. We very much need to improve this, but that improvement is out of scope for this feature. I will write a thorough test suite for the expected handling of punctuation (and tags), then create a separate PR with that update.

The editor code that uses this endpoint is in https://github.com/zanata/zanata-spa/pull/91


zanata-war/src/main/java/org/zanata/rest/editor/MediaTypes.java
```diff
@@ -32,6 +32,11 @@ public String toString() {
     public static final String APPLICATION_ZANATA_LOCALES_JSON =
             APPLICATION_ZANATA_LOCALES + JSON;
 
+    public static final String APPLICATION_ZANATA_SUGGESTIONS =
+            APPLICATION_VND_ZANATA + ".suggestions";
+    public static final String APPLICATION_ZANATA_SUGGESTIONS_JSON =
+            APPLICATION_ZANATA_SUGGESTIONS + JSON;
+
```
huangp: I am not sure what's the benefit of using vnd mime type v.s. generic application/json or application/xml. To me, a specific vnd type will mean: if I am writing a client for it, I have to remember to set the Accept header to this type otherwise I will get 415 media not supported. I am not against it more of a question to ask. @seanf @carlosmunoz

zanata-war/src/main/java/org/zanata/rest/editor/service/SuggestionsService.java
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * Try to get a valid locale for a given string. If the string does not match a supported
+     * locale, generate an error response with the given error message.
+     *
+     * @param localeString used to look up the locale
+     * @param errorFormat error to display. Include "%s" to substitute the locale string.
+     * @return a LocaleId if the given string matches one, otherwise an error response.
+     */
+    private Either<LocaleId, Response> getLocaleOrError(final String localeString, final String errorFormat) {
```
huangp: errorFormat is the first argument to String.format. May be better to keep this consistent by changing the order of the arguments in this method signature. Not a major thing though.

zanata-war/src/main/java/org/zanata/rest/editor/service/SuggestionsService.java
```diff
@@ -0,0 +1,142 @@
...
+
+
+    /**
+     * Try to get a valid search type constant for a given string. If the string does not
+     * match a valid search type, generate an error response with appropriate error message.
+     *
+     * @param searchTypeString used to look up the search type.
+     * @return A SearchType if the given string matches one, otherwise an error response.
+     */
+    private Either<SearchType, Response> getSearchType(String searchTypeString) {
```
huangp: When searchTypeString is null will it trigger NPE or IllegalArgumentException? May need to catch both?
Alternatively, we can loop through SearchType.values() array and see if we can find a match *case insensitively*. This may offer friendlier user experience but just a suggestion.

zanata-war/src/main/java/org/zanata/rest/editor/service/resource/SuggestionsResource.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+
+/**
+ * Endpoint to search for suggestions from translation memory and other sources.
+ */
+@Produces({ MediaType.APPLICATION_JSON })
+@Consumes({ MediaType.APPLICATION_JSON })
+public interface SuggestionsResource {
+
+    public static final String SERVICE_PATH = "/suggestions/{sourceLocale}/{transLocale}";
```
huangp: I would suggest moving the sourceLocale and transLocale from path params to query params. At least move transLocle because sourceLocale will most likely be en-US or en. Some locales with variable such as 'kw@kkcor' will look odd in path after encoding.

zanata-war/src/main/java/org/zanata/rest/editor/MediaTypes.java
carlosmunoz: I think it has something to do with service auto-discovery and custom type description. Not sure if it's giving us anything the way we use it. @seanf ?

seanf: Custom media types are kind of an open question in RESTful architecture - http://www.infoq.com/news/2010/01/subbu-custom-media-types - old article, but still true.

The problem is that `application/json` doesn't really tell you anything about the semantics of the returned JSON.

Can't the server say that it `@Consumes`/`@Produces` both `application/json` and the custom `vnd` type?

zanata-war/src/main/java/org/zanata/rest/editor/service/SuggestionsService.java
seanf: Seeing two String parameters makes me nervous, because it's so easy to get the invocation wrong.  Perhaps it would be better to have the callers do their own string formatting.

zanata-war/src/main/java/org/zanata/rest/editor/service/SuggestionsService.java
```diff
@@ -0,0 +1,142 @@
...
+        final Response.ResponseBuilder errorResponse = Response.status(BAD_REQUEST).entity(errorMessage);
+
+        try {
+            @Nullable HLocale hLocale = localeService.getByLocaleId(localeString);
+            if (hLocale == null) {
+                return Either.right(errorResponse.build());
+            }
+            return Either.left(hLocale.getLocaleId());
+        } catch (IllegalArgumentException e) {
+            return Either.right(errorResponse.build());
```
seanf: `IllegalArgumentException` sounds like the sort of exception which should be logged (and hopefully prevented before occurring).

zanata-war/src/main/java/org/zanata/rest/editor/MediaTypes.java
huangp: It will make more sense to return vnd together with xml type. json has no type so really not giving us anything IMHO. I think I've read somewhere mentioning to include API version as media type. Not something we do at the moment. Either way I am fine with whatever we choose. 
----- Original Message -----

seanf: JSON doesn't have explicit schema support, but in use it still has implicit schemas, just like NoSQL: http://martinfowler.com/articles/schemaless/

davidmason: I do not see a clear course of action here, and it seems that at the moment it would cause no problems to keep or remove the vnd mime type. Is everyone happy with leaving this as-is for now to keep it consistent with the adjacent media type strings?

- [ ] @huangp 
- [ ] @carlosmunoz 
- [ ] @seanf

zanata-war/src/main/java/org/zanata/rest/editor/service/SuggestionsService.java
davidmason: I chose this order so that the arguments are in the order from the method name. In this case I consider `String.format` to be an implementation detail so I don't want to let its parameter order influence the signature. It is leaking the fact that a format string is required though (I wish Java had concise type aliases to make this clear).

- [ ] I will see how the code looks with the strings formatted before being passed to this method.

zanata-war/src/main/java/org/zanata/rest/editor/service/SuggestionsService.java
davidmason: - [x] I changed it to case-insensitive, and the error/null handling was no longer needed. :tada:

zanata-war/src/main/java/org/zanata/rest/editor/service/resource/SuggestionsResource.java
davidmason: @huangp users will not see this URL, so it is just developers who may have that experience with those locales. Do you think it would be more usable for developers generating the URL as a path parameter?

It seems like the `@` has to be encoded either way (see [What characters are allowed unencoded in query strings?](http://www.456bereastreet.com/archive/201008/what_characters_are_allowed_unencoded_in_query_strings/)).

zanata-war/src/main/java/org/zanata/rest/editor/service/SuggestionsService.java
davidmason: @seanf in this case the `IllegalArgumentException` happens when some code in `localeService.getByLocaleId` tries to create a `LocaleId` from the given string - we manually throw the exception in its constructor.

- [ ] I could refactor `getByLocaleId` to duplicate the check before creating the `LocaleId`, or to catch the exception and return `null`. The latter has less duplication, and seems like a safe change, so I will try that.

zanata-war/src/main/java/org/zanata/rest/editor/service/resource/SuggestionsResource.java
huangp: My personal preference and also I think it's fairly common practice is to leave the path portion URL friendly(e.g. our slug only allow URL friendly characters). Query params can be encoded. It's also easier for developer writing client (at least with the new JAX-RS client, they automatically encode query params for you but not path)

zanata-war/src/main/java/org/zanata/rest/editor/service/SuggestionsService.java
huangp: Fair enough.

zanata-war/src/main/java/org/zanata/rest/editor/MediaTypes.java
seanf: We needn't tackle cleaning up the adjacent media types right now, but I think SuggestionsResource should produce APPLICATION_ZANATA_SUGGESTIONS_JSON and (on second thought, David) generic APPLICATION_JSON.  We don't need APPLICATION_ZANATA_SUGGESTIONS as such.  The client should request 
APPLICATION_ZANATA_SUGGESTIONS_JSON.

Note that the current code doesn't actually appear to use APPLICATION_ZANATA_SUGGESTIONS_JSON anywhere.

zanata-war/src/main/java/org/zanata/rest/editor/service/SuggestionsService.java
seanf: By making a format string one of the parameters, String.format already influenced the semantics of the method, so why not let it influence the signature?

Still, we shouldn't be too paranoid; it's a private method.

zanata-war/src/main/java/org/zanata/rest/editor/service/SuggestionsService.java
seanf: It's generally best to prevent exceptions from happening in the first place.  Catching exceptions without logging can lead to a lot of pain, and worth avoiding as a habit.

You could always extract the check from LocaleId's constructor to a public method, and call that first if you suspect an illegal locale code.  That's probably an improvement to LocaleId's API.

zanata-war/src/main/java/org/zanata/rest/editor/service/SuggestionsService.java
davidmason: > By making a format string one of the parameters, String.format already influenced the semantics of the method, so why not let it influence the signature?

I changed it so that is passes an already-formatted string instead, once I considered how error-prone it might be to have the format string distant from the call to `String.format`

zanata-war/src/main/java/org/zanata/rest/editor/service/resource/SuggestionsResource.java
davidmason: My thinking when I chose to make them path parameters was that required parameters make more sense to have in the path, and in this case it only makes sense to ask for suggestions to translate something from one locale to another locale.

I think the main reason I am attached to the path parameters is that I want the URL to look nice and clean, and it looks messier with `?sourceLocale=en-US&transLocale=fr` than with `/en-US/fr`. Maybe I can find some smaller parameter names to make something that satisfies that desire.

@huangp what do you think of `/suggestions?from=en-US&to=fr`?

huangp: Yep I like this more: /suggestions?from=en-US&to=fr 

----- Original Message -----

davidmason: Cool, I will go with that for now then.

huangp:
:+1:

zanata-war/src/main/java/org/zanata/service/impl/LocaleServiceImpl.java
```diff
@@ -263,8 +264,16 @@ public HLocale getByLocaleId(@Nonnull LocaleId locale) {
...
     public HLocale getByLocaleId(@Nonnull String localeId) {
-        return this.getByLocaleId(new LocaleId(localeId));
+        final LocaleId locale;
+        try {
+            locale = new LocaleId(localeId);
+        } catch (IllegalArgumentException e) {
+            log.warn("Tried to look up a locale with a malformed id", e);
+            return null;
+        }
+        return this.getByLocaleId(locale);
```
seanf: This plus the variable declaration could live inside the `try {}` as a one liner.  (Not important though.)

seanf:
:+1:

zanata-war/src/main/java/org/zanata/service/impl/LocaleServiceImpl.java
davidmason: :+1: I like it. Done.

seanf:
:+1:

aeng:
Test war file: http://zanata-build.lab.eng.bne.redhat.com:8080/job/zanata-spa-integration-build/28/

davidmason:
Is this what you want, ninja? :+1:

djansen-redhat:
- [ ] Empty match percentage on manual search: http://i.imgur.com/4HDGmmi.png

djansen-redhat:
@davidmason I've disabled that failing test.

djansen-redhat:
:white_check_mark: Tested. Looks good guys.

djansen-redhat:
Considering the above added detail, I believe I can pass this.
:white_check_mark: Tested

davidmason:
Squash-merged in https://github.com/zanata/zanata-server/commit/73f58678c5d455763f809f9da587b676f5cfb23a

PR #818, opened 2015-05-18
Update deprecated reference

aeng:


seanf:
@carlosmunoz Could you please review the changes to ConditionRuleActionType and RoleTypeType et al?

zanata-model/src/main/java/org/zanata/model/validator/UniqueValidator.java
```diff
@@ -62,7 +62,7 @@ private int countRows(Object value) {
                 session.getSessionFactory().getClassMetadata(value.getClass());
         String idName = metadata.getIdentifierPropertyName();
         // FIXME was EntityMode.POJO
-        Serializable id = metadata.getIdentifier(value);
+        Serializable id = metadata.getIdentifier(value, null);
```
seanf: Perhaps you should cast the null to a particular type, so that a future overload of `getIdentifier` doesn't break compilation.

zanata-war/src/test/java/org/zanata/webtrans/test/GwtTestSample.java
```diff
@@ -16,6 +16,6 @@ public void testSomething() {
         // avoided wherever possible in favour of plain JUnit tests.
         // Only code that requires a DOM should be tested with these
         // (e.g. most view classes).
-        Assert.assertTrue(true);
+        Assertions.assertThat(true).isTrue();
```
seanf: Maintaining useless code is always fun.  If we want to keep this line at all, we should just switch from `junit.framework.Assert` to `org.junit.Assert`, instead of making this code depend on assertj.  But really, just delete the line.  In fact, delete the whole useless test class.

seanf:
Looks good to me otherwise, but I think @carlosmunoz should review the JPA changes, not me.

carlosmunoz:
Since tests seemed to pass, I don't see anything to hold this up. One question though:
Why did we need to add a few Type Descriptors?

aeng:
@carlosmunoz Each descriptor has different type. Perhaps I can create a CharDescriptor type since its all using initial of the string. What do you think?

aeng:
@carlosmunoz on second thought, I think its clearer to have each type with its own descriptor. Let me know what you think.

carlosmunoz:
I think that makes sense. There is already a `CharTypeDescriptor` in hibernate, and I think we should either create a generic 'XXEnumType' that knows how to map enums to the first character, or if that is not possible, then at least a base class for the common code.

carlosmunoz:
You beat me to my reply, so when I said "that makes sense" I was talking about the first one. I think you need different 'Types' for each enum, but I think a single TypeDescriptor should work (if possible).

aeng:
@carlosmunoz care to review the descriptor again?

carlosmunoz:
:+1: Looks good.
(btw, you have me doing code reviews at 10:30 pm over here... :sleepy:)

aeng:
@carlosmunoz the night is still young boss.

aeng:
Not sure how to QA on this PR.

djansen-redhat:
I lol'd at that [comment](https://github.com/zanata/zanata-server/pull/818#issuecomment-103766001)

:white_check_mark: F-Tests verify this.

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/73a9960014d1ebc577325c385786618cb0385e3d

PR #820, opened 2015-05-18
Fix cannot cast FirefoxDriver to EventFiringWebDriver error

djansen-redhat:


functional-test/src/main/java/org/zanata/page/WebDriverFactory.java
```diff
@@ -143,17 +141,18 @@ public void updateListenerTestName(String testName) {
...
         }
-        enableScreenshots();
+        driver = enableScreenshots();
         eventListener.updateTestID(testName);
     }
 
     private WebDriver enableScreenshots() {
         log.debug("Enabling screenshot module...");
-        return EventFiringWebDriver.class.cast(driver).register(eventListener);
+        return new EventFiringWebDriver(driver).register(eventListener);
```
seanf: Could we just do this in `createDriver()` ?  The fact that `enableScreenshots()` returns a `WebDriver` (which is really an `EventFiringWebDriver`) rather than `void` is all a bit non-obvious.

djansen-redhat: Not sure. But I certainly can change it to a void return.

djansen-redhat:

seanf:
This build has been running for almost 4 hours, and it has a lot of 30 second pauses in the log.

functional-test/pom.xml
```diff
@@ -839,6 +839,7 @@
                 <echo>-Dwebdriver.chrome.bin=full path to chrome binary.</echo>
                 <echo>-Dwebdriver.chrome.driver=full path to chromedriver binary. Default is chromedriver on $PATH</echo>
                 <echo>-Dwebdriver.wait=global wait time in seconds for element searches. Default is 10.</echo>
+                <echo>-Dwebdriver.screenshot.dir=location to store screenshots during test execution.Default is ${project.build.directory}/screenshots</echo>
```
aeng: I assume we are doing mvn clean on every built?

seanf: We should be, at least for the build server.

djansen-redhat: Yes.

functional-test/src/main/java/org/zanata/page/WebDriverFactory.java
```diff
@@ -60,33 +55,47 @@
...
         return driver;
     }
 
-    private WebDriver createDriver() {
-        WebDriver driver = createPlainDriver();
+    private EventFiringWebDriver createDriver() {
+        String driverType = PropertiesHolder.getProperty(webDriverType.value());
+        EventFiringWebDriver newDriver;
+        switch (driverType.toLowerCase()) {
+            case "chrome":
```
aeng: This is not big deal, but since we've got `Constants` in FT, perhaps we should move `chrome` and `firefox` to there.

aeng:
Finished review. Nothing major. :+1:

functional-test/src/main/java/org/zanata/page/WebDriverFactory.java
djansen-redhat: Can't use non-primitive constants in a switch. (afaik)

aeng:
:+1:

djansen-redhat:
Merged in 362e8d776792e97173795e9486ba761e3798af57

PR #823, opened 2015-05-19
Standardise on JUnit; remove TestNG

seanf:
I'm trying to cut down the number of groups and remove the use of suite files.


seanf:
@huangp @carlosmunoz I've started a branch where we can work on removing TestNG

seanf:
@huangp looks like we need a new way of initialising `ForceLoadClass` or at least the work it was doing.

seanf:
@carlosmunoz @huangp What exactly does ForceLoadClass do?  And what's the story of the failure you can get without it?

carlosmunoz:
I have no idea what `ForceLoadClass` does. That transformer has been there for a while and we had never touched it.

huangp:
It's to do with SeamAutowire trying to modify a class. If the class is already loaded javassist will throw that exception. I am trying to make it so that SeamAutowire will be loaded and static block be executed (initialize all the modified classes like Contexts etc.) before any tests are run. 

----- Original Message -----

seanf:
I investigated a bit more, and to remove (some of) the duplicate quarantine-related annotations, we would need a fix for https://github.com/BinaryTweed/quarantining-test-runner/issues/4  Then subclass tests would be able to inherit the annotations I already put on abstract parent classes like `ZanataJpaTest`.

By the way, the quarantining stuff is the replacement for `ForceLoadClass` which used to be triggered by `TestTransformer`, but the replacement should provide better isolation between Seam-related tests.  (I put the quarantine annotations on every test which referenced `org.jboss.seam` or `SeamAutowire`.)

Also, JUnit tends to have more isolation than TestNG, because every test invocation is run with its own instance of the test class.

zanata-war/src/test/java/org/zanata/ImmutableDbunitJpaTest.java
```diff
@@ -20,45 +20,42 @@
...
+import org.junit.After;
+import org.junit.Before;
 
 /**
  * @author Sean Flanigan <a href="mailto:sflaniga@redhat.com">sflaniga@redhat.com</a>
  */
 public abstract class ImmutableDbunitJpaTest extends ZanataDbunitJpaTest {
 
-    @BeforeClass
+    @Before
```
carlosmunoz: Should this be JUnit's `@BeforeClass` instead?

zanata-war/src/test/java/org/zanata/ZanataDbunitJpaTest.java
```diff
@@ -48,33 +48,16 @@
...
         return binaryDir;
     }
 
-    @BeforeClass
-    @Parameters("replaceNull")
-    public void setReplaceNull(@Optional Boolean replaceNull) {
-        if (replaceNull == null)
-            return;
-        this.replaceNull = replaceNull;
-    }
```
carlosmunoz: So, these methods are no longer needed... :question:

zanata-war/src/test/java/org/zanata/file/TranslationDocumentUploadTest.java
```diff
@@ -23,14 +23,13 @@
...
+import org.junit.Before;
+import org.junit.Test;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
-import org.testng.annotations.AfterMethod;
-import org.testng.annotations.BeforeMethod;
-import org.testng.annotations.Test;
 import org.zanata.exception.ChunkUploadException;
 
-@Test(groups = { "unit-tests" })
```
carlosmunoz: In general, I guess we are losing test groups. I take it we were not using them.

zanata-war/src/test/java/org/zanata/ImmutableDbunitJpaTest.java
seanf: On 2015-06-02 13:56, Carlos Munoz wrote:
> In zanata-war/src/test/java/org/zanata/ImmutableDbunitJpaTest.java
> <https://github.com/zanata/zanata-server/pull/823#discussion_r31491550>:
> 
>>  
>>  /**
>>   * @author Sean Flanigan <a href="mailto:sflaniga@redhat.com">sflaniga@redhat.com</a>
>>   */
>>  public abstract class ImmutableDbunitJpaTest extends ZanataDbunitJpaTest {
>>  
>> -    @BeforeClass
>> +    @Before
> 
> Should this be JUnit's |@BeforeClass| instead?

That would be a more direct translation, but JUnit's BeforeClass can't
access non-statics.  This particular case shouldn't hurt test times too
much, because ImmutableDbunitJpaTest isn't used a lot.

> 
> —
> <https://github.com/zanata/zanata-server/pull/823/files#r31491550>.
> 


-- 
Sean Flanigan

Principal Software Engineer
Red Hat

zanata-war/src/test/java/org/zanata/ZanataDbunitJpaTest.java
seanf: On 2015-06-02 13:58, Carlos Munoz wrote:
> In zanata-war/src/test/java/org/zanata/ZanataDbunitJpaTest.java
> <https://github.com/zanata/zanata-server/pull/823#discussion_r31491577>:
> 
>>      protected String getBinaryDir() {
>>          return binaryDir;
>>      }
>>  
>> -    @BeforeClass
>> -    @Parameters("replaceNull")
>> -    public void setReplaceNull(@Optional Boolean replaceNull) {
>> -        if (replaceNull == null)
>> -            return;
>> -        this.replaceNull = replaceNull;
>> -    }
> 
> So, these methods are no longer needed... :question:

They haven't really been needed since we stopped injecting test
parameters via xml files.  Which always made IDE testing difficult, anyway.


> —
> <https://github.com/zanata/zanata-server/pull/823/files#r31491577>.
> 


-- 
Sean Flanigan

Principal Software Engineer
Red Hat

zanata-war/src/test/java/org/zanata/file/TranslationDocumentUploadTest.java
seanf: On 2015-06-02 14:11, Carlos Munoz wrote:
> In
> zanata-war/src/test/java/org/zanata/file/TranslationDocumentUploadTest.java
> <https://github.com/zanata/zanata-server/pull/823#discussion_r31491869>:
> 
>>  import org.zanata.exception.ChunkUploadException;
>>  
>> -@Test(groups = { "unit-tests" })
> 
> In general, I guess we are losing test groups. I take it we were not
> using them.

No, we weren't really using them any more, except to separate unit tests
from integration tests.  I think file naming conventions will suffice,
but we could use JUnit's Categories if we want finer control.

> 
> —
> <https://github.com/zanata/zanata-server/pull/823/files#r31491869>.
> 


-- 
Sean Flanigan

Principal Software Engineer
Red Hat

carlosmunoz:
:+1: Reviewed.

seanf:
@carlosmunoz commit d7d8016 cuts down on the boilerplate a bit.

carlosmunoz:
:+1:

seanf:
Merged in 3e0ec6ef4d72bf3731d43a2390144c3e61dc01c8.

PR #833, opened 2015-05-29
Upgrade to RESTEasy 3

seanf:
Code compiles, but tests are broken.  

We also need to do something about the JAXB/JSON mappings for TextContainer (two mappings for "content").  `org.codehaus.jackson.map.ser.BeanSerializerModifier` might be needed. http://www.cowtowncoder.com/blog/archives/2011/02/entry_443.html

See also http://stackoverflow.com/questions/7161638/how-do-i-use-a-custom-serializer-with-jackson


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-server/pull/833" target="_blank">ReviewNinja

seanf:
@carlosmunoz Have you got anywhere with a solution or workaround for the duplicate mappings in TextContainer?

carlosmunoz:
@seanf Not yet. Although I have a good idea on how to approach it.

zanata-war/src/main/webapp-jboss/WEB-INF/jboss-deployment-structure.xml
```diff
@@ -3,12 +3,15 @@
...
   <deployment>
-    <!-- These exclusions let us use RESTEasy 2 instead of the
+    <!-- These exclusions let us use the war file's RESTEasy instead of the
          container's version -->
     <exclude-subsystems>
       <subsystem name="jaxrs" />
     </exclude-subsystems>
     <exclusions>
+      <module name="javax.ws.rs.api" />
+      <module name="javaee.api" />
```
huangp: Excluding this causes the class not found exception (missing "javax.inject-api"). We either declare a jboss module dependency "javax.inject.api" or we include the jar in war.

huangp:
@seanf are you happy with this PR?

seanf:
@huangp I'm still reviewing it.  By the way, I'm not sure if much is left of my original RESTEasy 3 changes, but have you reviewed all of them (or your revised versions of all of them)?

zanata-war/src/main/java/org/zanata/rest/ZanataResteasyBootstrap.java
```diff
@@ -50,12 +50,6 @@ public void registerHotDeployedClasses() {
     @Override
     protected void initDispatcher() {
         super.initDispatcher();
-                getDispatcher().getProviderFactory()
-                .getServerPreProcessInterceptorRegistry()
-                .register(ZanataRestSecurityInterceptor.class);
-        getDispatcher().getProviderFactory()
-                .getServerPreProcessInterceptorRegistry()
-                .register(ZanataRestVersionInterceptor.class);
     }
```
seanf: This whole method could go.  (I know, this is probably mine.)

zanata-war/src/test/java/org/zanata/arquillian/Deployments.java
```diff
@@ -65,7 +66,9 @@ public static void main(String[] args) {
                 // JavaMelody's ServletFilter/Listener interfere with test
                 // deployments
                 new RejectDependenciesStrategy(false,
-                        "net.bull.javamelody:javamelody-core")).asFile();
+                        "net.bull.javamelody:javamelody-core",
+                        "com.google.collections:google-collections"))
```
seanf: Please add a comment for this one.  Did you ever work out where the google-collections dependency came from?

huangp:
I didn't touch your original change. I just fix all the tests. I have now reviewed your original changes and I am fine with it.

seanf:
Reviewed https://github.com/zanata/zanata-server/compare/2b08324f1a978c513315be5518e2aee08b911af9...772525afbbfc13d199f2fc204184d29d8d49dfaa#diff-81988d116fbd66e0ea6993a5390b69d9R70

Looks good; just those two minor things.

seanf:
:+1:

djansen-redhat:
:see_no_evil: Verified by Sir Jenkins.

huangp:
squash merged in 766cfc0

PR #839, opened 2015-06-01
Improve project search in Zanata with lucene

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1147304
https://bugzilla.redhat.com/show_bug.cgi?id=1123186

### DIFF
--- a/zanata-war/src/main/java/org/zanata/dao/ProjectDAO.java
+++ b/zanata-war/src/main/java/org/zanata/dao/ProjectDAO.java
@@ -7,6 +7,8 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 import org.apache.commons.lang.StringUtils;
+import org.apache.lucene.analysis.StopAnalyzer;
+import org.apache.lucene.analysis.WhitespaceAnalyzer;
 import org.apache.lucene.analysis.standard.StandardAnalyzer;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.queryParser.ParseException;
@@ -14,6 +16,7 @@ import org.apache.lucene.queryParser.QueryParser;
 import org.apache.lucene.search.BooleanClause;
 import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.TermQuery;
+import org.apache.lucene.search.WildcardQuery;
 import org.apache.lucene.util.Version;
 import org.hibernate.Query;
 import org.hibernate.Session;
@@ -25,13 +28,12 @@ import org.jboss.seam.annotations.In;
 import org.jboss.seam.annotations.Name;
 import org.jboss.seam.annotations.Scope;
 import org.zanata.common.EntityStatus;
+import org.zanata.hibernate.search.CaseInsensitiveWhitespaceAnalyzer;
 import org.zanata.hibernate.search.IndexFieldLabels;
 import org.zanata.model.HAccount;
 import org.zanata.model.HPerson;
 import org.zanata.model.HProject;
 import org.zanata.model.HProjectIteration;
-
-import com.google.common.collect.Lists;
 
 @Name("projectDAO")
 @AutoCreate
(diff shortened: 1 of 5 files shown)


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-server/pull/839" target="_blank">ReviewNinja

zanata-war/src/main/java/org/zanata/dao/ProjectDAO.java
```diff
@@ -255,15 +255,35 @@ public int getTotalObsoleteProjectCount() {
...
+    private Query getSearchProjectQuery(@Nonnull String searchQuery,
+            boolean includeObsolete) {
+        StringBuilder sb = new StringBuilder();
+        sb.append("from HProject p where ");
+
+        if(!includeObsolete) {
+            sb.append("p.status <> ")
+                .append("'" + EntityStatus.OBSOLETE.getInitial() + "' and ");
+        }
+        sb.append("lower(p.name) like :query or lower(p.slug) like :query");
```
huangp: I think you need to put brackets around "lower(p.name) like :query or lower(p.slug) like :query". Even better put it as a named query in HProject since includeObsolete is almost always false now.

seanf:
- [ ] Handle non-substring searches

If this problem is addressed by switching to case-insensitive substring searches, I think we will be introducing a different problem.  For instance, if I have a multi-word project name like BGR Field Translations, we will no longer be able to find it by the first and last words.

The previous Hibernate Search solution may be a bit word-centric for the case mentioned in https://bugzilla.redhat.com/show_bug.cgi?id=1123186, but at least it lets you put the words (terms) in any order.

Perhaps we need to use Lucene's more advanced query options to do with fuzziness and punctuation.

aeng:
@seanf I think SQL search will solve more issues reported by user (hyphen and special char search doesn't work). Of course, word centric search will be a plus, but how many words can a project name be? and what are the odds of user searching for first and last word of project name?

seanf:
First and last word is just an example, but with the project "BGR Field Translations", I don't think it would be unusual to search for, say, BGR Translations.  And the longer the project name, the more likely some of the words are less important, and likely to be omitted from search queries by users.

Most good search engines allow term-based searches, not just substring searches, so some users will expect to be able to put words in any order.  That's why specialised text search libraries like Lucene and friends exist.  

The punctuation problems should be fixable with Lucene, perhaps by tweaking the query parser/Analyzer to strip out punctuation.  I think the partial word problem in https://bugzilla.redhat.com/show_bug.cgi?id=1123186 can be handled using a wildcard query.

zanata-war/src/main/java/org/zanata/dao/ProjectDAO.java
```diff
@@ -311,6 +286,34 @@ private FullTextQuery getTextQuery(@Nonnull String searchQuery,
...
+     *            special char
+     * @param field
+     *            - lucene field
+     */
+    private BooleanQuery buildSearchFieldQuery(@Nonnull String searchQuery,
+        @Nonnull String field) throws ParseException {
+        BooleanQuery query = new BooleanQuery();
+
+        //replace hyphen with space for lucene search
+        searchQuery = searchQuery.replace("-", " ");
```
huangp: we allow ".", "-" and "_" in project slug. Not sure whether you need to replace the other two as well here.

aeng: good point. I shall test in those.

aeng: I've remove the replace of `-`, its not needed anymore with the new analyzer `(whitespace case insensitive)`. Now index for project name and desc includes special char.

huangp:
:+1:

djansen-redhat:
:white_check_mark: Tested, waiting for build

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/1b2c207a1c0ec9eea604a00311c665cb36ae3a12

PR #848, opened 2015-06-04
Downgrade Surefire for release

seanf:
We can't release Zanata with a SNAPSHOT dependency, so we will have to downgrade for the moment.


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-server/pull/848" target="_blank">ReviewNinja

carlosmunoz:
:+1:

PR #853, opened 2015-06-05
Fix broken internal link, update mkdocs.yml deprecated syntax

aeng:


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-server/pull/853" target="_blank">ReviewNinja

carlosmunoz:
:+1: the yaml config looks much better now. Is the menu layout still fixed to 2 levels max?

aeng:
@carlosmunoz only deprecated syntax, only supports 2 level still.

http://www.mkdocs.org/user-guide/writing-your-docs/#multilevel-documentation

PR #854, opened 2015-06-08
Fall back on stop-words comparison for similarity when nothing else is available.

davidmason:
A divide-by-zero condition was generating a NaN value for similarity for strings
that contain only stop-words. When that condition would be met, this will instead
repeat the comparison without ignoring stop-words, then finally just return 0.0
if there are still no tokens to compare.

### DIFF
--- a/zanata-war/src/main/java/org/zanata/search/LevenshteinTokenUtil.java
+++ b/zanata-war/src/main/java/org/zanata/search/LevenshteinTokenUtil.java
@@ -129,8 +129,8 @@ public class LevenshteinTokenUtil {
     /**
      * Splits into tokens (lower-case).
      *
-     * @param s
-     * @return
+     * @param s the string to tokenise
+     * @return an array of lowercase tokens (words)
      */
     static String[] tokenise(String s) {
         String[] tokens = s.toLowerCase().split(SPLIT_REGEX);
@@ -141,6 +141,16 @@ public class LevenshteinTokenUtil {
             }
         }
         return list.toArray(new String[list.size()]);
+    }
+
+    /**
+     * Like tokenise(String) but does not discard stop words.
+     *
+     * @param s the string to tokenise
+     * @return an array of lowercase tokens (words)
+     */
+    static String[] tokeniseAndKeepStopWords(String s) {
+        return s.toLowerCase().split(SPLIT_REGEX);
     }
 
     private static int countExtraStringLengths(List<String> strings,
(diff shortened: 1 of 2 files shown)


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-server/pull/854" target="_blank">ReviewNinja

zanata-war/src/main/java/org/zanata/search/LevenshteinTokenUtil.java
```diff
@@ -159,56 +169,113 @@ private static int countExtraStringLengths(List<String> strings,
...
+     *  - Strings at the same index are compared.
+     *  - Stop-words are ignored in comparisons. See #stopwords.
+     *  - When both lists are empty, they are considered identical (returns 1.0)
+     *  - Empty strings are considered identical to other empty strings.
+     *
+     * If a string is made up only of stop-words, the calculation will be
+     * repeated without ignoring stop-words. This is so that a sensible score
+     * can be returned when there is nothing else to compare.
+     *
+     * If comparisons with and without stop-words generate usable information,
```
huangp: Should this be "un-usable"? Because the fallback is 0.0?

zanata-war/src/main/java/org/zanata/search/LevenshteinTokenUtil.java
```diff
@@ -129,8 +129,8 @@ public static double getSimilarity(final String s1, final String s2) {
     /**
      * Splits into tokens (lower-case).
      *
-     * @param s
-     * @return
+     * @param s the string to tokenise
+     * @return an array of lowercase tokens (words)
      */
     static String[] tokenise(String s) {
         String[] tokens = s.toLowerCase().split(SPLIT_REGEX);
```
huangp: Did you see https://bugzilla.redhat.com/show_bug.cgi?id=805737#c20 Probably fix the REGEX as well since you are here.

seanf:
If two strings only have stop words in common, perhaps we should keep things simple (for us and for users), and just return 0% similarity.  (And then maybe discard 0% matches as useless.)

davidmason:
> If two strings only have stop words in common, perhaps we should keep things simple (for us and for users), and just return 0% similarity. (And then maybe discard 0% matches as useless.)

I do not want a user to search for "No" and get back a result that is "No" and has 0% match. I think we should review the exclusion of stop words at all for this calculation, since the values users see do not match what they see in the diff. That is too major a change to make without some research into user expectations though.

seanf:
A result that just has "No" is basically useless. TM isn't meant to be a
glossary, or a dictionary.  By discarding stop-word-only matches, we should
be able to improve the overall quality of TM results.

If a fuzzy match comes back as 0% similar, we should throw it away.

seanf:
:-1: There's a bit too much in here.  Let's just fix the divide by zero for now.  The behaviour with regard to stop words (or discarding stop word-only matches) shouldn't be otherwise changed in this bug fix.

djansen-redhat:
Do we intend to revisit this?

davidmason:
> Do we intend to revisit this?

@djansen-redhat I intend to, but I think there is too much bikeshedding around this. I intend to get some data from users about their needs and expectations and convert that into a thorough unit test suite rather than spend any more time guessing the best approach to this.

PR #855, opened 2015-06-09
Ignore flaky change password test

djansen-redhat:


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-server/pull/855" target="_blank">ReviewNinja

aeng:
Question, how are we going to handle the flaky test? are we going to revisit them?

aeng:
Apart from the question, everything is good. :+1:

djansen-redhat:
@aeng yes, after the release I will revisit the flaky tests.  It may be simple, but I don't have the time to do much other than manually verify them for now.

PR #856, opened 2015-06-09
Make frontend module more offline friendly

huangp:
It will install node and npm into $user.home/Downloads.
It will also try to use npm cache as much as possible.

### DIFF
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -2,7 +2,7 @@
 
 At the moment it only contains "user profile page" bundle.
 
-To build it, just run 
+To build it, just run
 
 ```mvn install```
 
@@ -15,4 +15,5 @@ Following mvn arguments is overrideable from command line:
 <npm.cli.script>${node.install.directory}/node/npm/bin/npm-cli.js</npm.cli.script>
 ```
 
-By default it will try to install npm modules using cache unless the cache entry is older than configured value (at time of writing, it's 999999 seconds).
+By default it will try to install npm modules from npm registry (default cache TTL is 10 seconds).
+If you activate profile ```-DnpmOffline``` the cache-min option will become 9999999 which means it will try to install npm modules from cache first.
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -1,4 +1,4 @@
-## This is a module to build frontend javascript projects
+## This is a module to build zanata frontend javascript projects
 
 At the moment it only contains "user profile page" bundle.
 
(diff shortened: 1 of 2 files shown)


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-server/pull/856" target="_blank">ReviewNinja

aeng:
Are we caching the node modules as well?

frontend/pom.xml
```diff
@@ -17,7 +17,7 @@
     <web.target>${project.build.directory}/web</web.target>
     <bundle.output>${project.build.directory}/build</bundle.output>
     <bundle.dest>${bundle.output}/META-INF/resources</bundle.dest>
-    <node.install.directory>${project.build.directory}</node.install.directory>
+    <node.install.directory>${download.dir}/zanata-frontend/node-${node.version}-npm-${npm.version}</node.install.directory>
```
seanf: So, it should be possible for the developer to override this with `-Dnode.install.directory`?  If so, good, but perhaps it should be documented in a readme somewhere (since it is taking a minor liberty with the dev's file system).

frontend/pom.xml
```diff
@@ -85,6 +85,9 @@
               <arguments>
                 <argument>${npm.cli.script}</argument>
                 <argument>install</argument>
+                <argument>--cache-min</argument>
+                <!-- in seconds -->
+                <argument>999999</argument>
```
seanf: Are there any potential drawbacks with long cache timeouts?  In Maven, over-caching should only lead to using old SNAPSHOTs as far as I know, but I'm not sure if the JavaScript dependencies make such a distinction (release vs snapshot).

huangp: There is no snapshot in js dependencies. I've tried it on my local machine. It will still build some native dependencies (written in C/C++ I think) but it seems not making any internet connections any more. The only down side I can think of is, it shares with your global npm/node configuration as well as cache.

huangp:
@aeng yes. The cache option is to cache npm modules.

frontend/pom.xml
```diff
@@ -125,4 +129,18 @@
     </plugins>
   </build>
 
+  <profiles>
+    <profile>
+      <id>npm offline</id>
```
seanf: I would suggest `npmOffline`, just for consistency.

frontend/README.md
```diff
@@ -0,0 +1,19 @@
+## This is a module to build frontend javascript projects
+
+At the moment it only contains "user profile page" bundle.
+
+To build it, just run
+
+```mvn install```
+
+Following mvn arguments is overrideable from command line:
```
seanf: I suggest something like:
"The following Maven properties can be overridden on the command line with `-Dkey=value`:"

seanf:
:+1:

djansen-redhat:
:white_check_mark: Tested, looks fine to me.

PR #858, opened 2015-06-09
Use generated-resources/deps

seanf:
This makes it easier to regenerate dependencies.properties after a
rebuild in IntelliJ IDEA, using "Generate Sources and Update Folders..."
button in the Maven Projects tool window.


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-server/pull/858" target="_blank">ReviewNinja

carlosmunoz:
:+1:

PR #862, opened 2015-06-11
Update api and common

davidmason:
Got 8 functional test failures in a branch with this. Trying it independently to see whether the tests pass.

PR #864, opened 2015-06-11
Simplify appserver properties

seanf:
This should make it easier to switch between EAP and WildFly for testing.  Assuming you have set `EAP6_URL`, and you're happy to use the version of WildFly specified in the pom, you just switch between `-Dappserver=jbosseap6` and `-Dappserver=wildfly8`.  

No need to specify any of these (and it's easier to switch if you don't):
- cargo.installation
- cargo.basename
- appserver.dir.name

You can still specify these if you want to change the version of WildFly itself or the version of the management api:
- wildfly.version
- wildfly.client.version

### DIFF
--- a/etc/scripts/extractAppserver.groovy
+++ b/etc/scripts/extractAppserver.groovy
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2015, Red Hat, Inc. and individual contributors
+ * as indicated by the @author tags. See the copyright.txt file in the
+ * distribution for a full listing of individual contributors.
+ *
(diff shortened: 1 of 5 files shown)


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-server/pull/864" target="_blank">ReviewNinja

seanf:
This PR includes the same change as https://github.com/zanata/zanata-server/pull/862

seanf:

huangp:
:+1:

PR #865, opened 2015-06-12
Remove ignore dup for schema/xml.xsd

seanf:
There are some differences between the copy of `schema/xml.xsd` in jboss-metadata-common and zanata-adapter-xliff, but I'm not sure jboss-metadata-common is still on the test classpath, so I'm hoping we might not need the exemption after all.


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-server/pull/865" target="_blank">ReviewNinja

seanf:
Looks like we still need it.

PR #867, opened 2015-06-12
Ignore slow tests

seanf:
`@SlowTest` is currently not handled by our JUnit setup, so we have to `@Ignore` such tests, at least for now.

### DIFF
--- a/zanata-war/src/test/java/org/zanata/SlowTest.java
+++ b/zanata-war/src/test/java/org/zanata/SlowTest.java
@@ -26,6 +26,8 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
+ * (Currently not handled by test framework: please use
+ * <code>@Ignore("slow test")</code> as well.)
  * Marker annotation to identify tests that might be potentially slow to run.
  *
  * @author Carlos Munoz <a
--- a/zanata-war/src/test/java/org/zanata/rest/service/ResourceUtilsJpaTest.java
+++ b/zanata-war/src/test/java/org/zanata/rest/service/ResourceUtilsJpaTest.java
@@ -69,7 +69,7 @@ public class ResourceUtilsJpaTest extends ZanataJpaTest {
         // TODO check the results in 'to'
     }
 
-    @Ignore
+    @Ignore("slow test")
     // This should be executed manually in IDE
     @Test
     @SlowTest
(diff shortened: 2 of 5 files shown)


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-server/pull/867" target="_blank">ReviewNinja

carlosmunoz:
@seanf Why did we need this change? `@SlowTest` was only a marker annotation as far as I know, and it wasn't being handled in any special way.

seanf:
@carlosmunoz If we were already running these tests, then the TestNG :arrow_right: JUnit slowdown must be unrelated, but as we hadn't intended to run the slow tests, I think it still makes sense to ignore them.

carlosmunoz:
Mmm... again, not sure if were in fact not intending to run these tests. 

I'm fine with the pull request in terms of code review, just not sure that we want to ignore the slow tests.

seanf:
> I'm fine with the pull request in terms of code review, just not sure that we want to ignore the slow tests.

?? Not sure I understand.  How can you be happy with the code review if you think the change itself is a mistake?


Anyway, `CopyTransServiceImplTest.testCopyTrans` is the only SlowTest which wasn't already Ignored.  And it turns out it's not that slow (by our *terrible* standards), because it runs all four combinations (some tests have dozens) in two seconds total.

So I'm going to remove that particular `@Ignore`, but I think I'll keep this PR open while I try to optimise some of the slower tests (including parameterised tests).

carlosmunoz:
> ?? Not sure I understand. How can you be happy with the code review if you think the change itself is a mistake?

I don't see any issues with the code itself if our intention is to ignore these tests for any reason (I trust your judgement on that). Personally I don't see such reason, but these days you are closer to the code than I am.

> Anyway, CopyTransServiceImplTest.testCopyTrans is the only SlowTest which wasn't already Ignored. And it turns out it's not that slow (by our terrible standards), because it runs all four combinations (some tests have dozens) in two seconds total.

> So I'm going to remove that particular @Ignore, but I think I'll keep this PR open while I try to optimise some of the slower tests (including parameterised tests).

Sounds like a plan.

seanf:
Running into a lot of intermittent tests, but the build is finally green.  @carlosmunoz I assume you're happy with the last commit?

seanf:
@carlosmunoz Could you please take another look at this PR?

seanf:

carlosmunoz:
:+1: Reviewed.
A couple of ignored tests due to them being more on the "performance profiling" category.

seanf:
(Rebased on top of `master`.)

PR #869, opened 2015-06-15
Reuse surefire forks

seanf:


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-server/pull/869" target="_blank">ReviewNinja

huangp:
:+1:

seanf:
I just checked an old commit, and before https://github.com/zanata/zanata-server/pull/823 (when `reuseForks` was `false` but `forkCount` was `1.5C`), I can only see one VM launched [1] for all the surefire tests.  Must be a difference between Surefire with TestNG vs Surefire with JUnit. (Those same settings led to concurrency failures when I first switched to JUnit, which is why I set `forkCount=1`.  And when `forkCount=1`, it seems `reuseForks=false` suddenly does what it says.) 

So that helps to explain the huge slowdown after we switched to JUnit and really started using one fork per test class.  It also means that I haven't suddenly reduced our test isolation - the tests weren't previously isolated after all.

[1] Based on the message `OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0`.  I guess we should think about fixing that warning.

PR #872, opened 2015-06-15
Replace Cobertura with JaCoCo

seanf:
This allows us to run unit tests once only, either with coverage
(`-Dcoverage`) or without.  I can't find a way to run tests exactly once
with Cobertura.  If surefire is configured to skip tests, Cobertura
won't run them either, and I can't find a way to generate a Cobertura
report on its own.

JaCoCo is configured to enforce coverage ratios, initially just for
zanata-war.

### DIFF
--- a/functional-test/pom.xml
+++ b/functional-test/pom.xml
@@ -27,6 +27,8 @@
     <cargo.container.home>${project.build.directory}/jboss/container</cargo.container.home>
     <context.path>zanata</context.path>
 
+    <coverage.halt.failure>false</coverage.halt.failure>
+
     <!--mysql-maven-plugin will use these setup-->
     <ds.database>root</ds.database>
     <ds.username>root</ds.username>
--- a/pom.xml
+++ b/pom.xml
@@ -21,9 +21,7 @@
 
   <properties>
     <byteman.version>2.1.2</byteman.version>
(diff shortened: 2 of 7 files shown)


review-ninja:
<a href="http://app.review.ninja/zanata/zanata-server/pull/872" target="_blank">ReviewNinja

huangp:
:+1:

huangp:
- [ ] fix jenkins. Remove cobertura plugin and replace it with jacoco plugin?

seanf:
Thanks @huangp 

I have changed the Jenkins job so that it no longer fails the build if there are no Cobertura reports, and it now specifies `-Dcoverage`.  However, thanks to https://github.com/zanata/zanata-server/pull/869 Cobertura's extra unit tests now only take 4 minutes for zanata-war.  Thus it might not be worth switching to JaCoCo at this stage, given the performance problems with its Jenkins plugin (https://issues.jenkins-ci.org/browse/JENKINS-26254 and https://issues.jenkins-ci.org/browse/JENKINS-16010).

seanf:
We can re-open if https://issues.jenkins-ci.org/browse/JENKINS-26254 and https://issues.jenkins-ci.org/browse/JENKINS-16010 are ever fixed.

PR #888, opened 2015-06-26
Store metadata for translation update. Update user statistic queries 

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1235070


zanata-model/src/main/java/org/zanata/model/type/TranslationEntityType.java
```diff
@@ -0,0 +1,44 @@
...
+package org.zanata.model.type;
+
+import lombok.Getter;
+import org.zanata.model.HTextFlowTarget;
+import org.zanata.model.tm.TransMemoryUnit;
+
+import java.io.Serializable;
+
+/**
+ * @author Alex Eng <a href="mailto:aeng@redhat.com">aeng@redhat.com</a>
```
huangp: Can we add javadoc for this class please

zanata-model/src/main/java/org/zanata/model/type/TranslationSourceType.java
```diff
@@ -0,0 +1,94 @@
...
+    GWT_EDITOR_ENTRY("GWT"),
+    JS_EDITOR_ENTRY("JS"),
+    API_UPLOAD("API"),
+    WEB_UPLOAD("WEB"),
+    UNKNOWN("UNK");
+
+    public static final Collection<TranslationSourceType> AUTOMATED_ENTRIES;
+
+    static {
+        AUTOMATED_ENTRIES = Collections.unmodifiableCollection(new HashSet(
```
huangp: Could use guava Sets.newHashSet(E...) instead

zanata-war/src/main/java/org/zanata/rest/service/AsynchronousProcessResourceService.java
```diff
@@ -156,6 +157,19 @@ public ProcessStatus startSourceDocCreationOrUpdate(final String idNoSlash,
                                                     // progress
     }
 
+    /**
+     * This is used by functional test ZanataRestCaller for testing purposes.
```
huangp: Is this only used by functional test? If yes we need to put it in zanata-test-war module. But I think this is used by client as well.

aeng: Double checked, its used by client too.

huangp:
:+1:  except adding more javadoc to a few of the new classes I am okay with the review

djansen-redhat:
Test:
- [ ] Contribution is counted when user translates to fuzzy
- [ ] Contribution is counted when user translates to translated
- [ ] Contribution is counted when user translates to translated, from fuzzy, without change
- [ ] Contribution is counted when user translates to Approved
- [ ] Contribution is not counted when user runs Copy Version to a new version
- [ ] Contribution is not counted when user runs CopyTrans on a version
- [ ] Contribution is not counted when user runs Merge Version on a version
- [ ] Contribution is not counted when user runs TMMerge from editor on a version
- [ ] CopyTrans etc action is visible in editor text flow comments

djansen-redhat:
It appears that history is being copied as-is, and the user is receiving credit for the history events.
We should mark all history that is copied as "automatedEntry" as well.

djansen-redhat:
:white_check_mark: Tested.

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/70e87501e4b0bdbc8c03be446e384184dfa61285

PR #889, opened 2015-06-30
Prepare for CDI (mostly security-related)

seanf:


pom.xml
```diff
@@ -391,6 +507,36 @@
         <scope>test</scope>
       </dependency>
 
+      <dependency>
+        <groupId>org.jglue.cdi-unit</groupId>
+        <artifactId>cdi-unit</artifactId>
+        <version>3.1.2</version>
+        <scope>test</scope>
+        <exclusions>
+          <exclusion>
```
huangp: Do we still need to exclude this even we explicitly define a dependency down below?

zanata-war/src/main/java/org/zanata/action/ProjectHome.java
```diff
@@ -1063,8 +1063,8 @@ private void ensureMutualExclusivity(
     /**
      * Update the about page to the entered value, and show a success message.
      */
-    @Restrict("#{s:hasPermission(projectHome.instance, 'update')}")
     public void updateAboutPage() {
+        identity.checkPermission(instance, "update");
```
huangp: Maybe we should just use entity listener (e.g. pre-update) to check for permission. This will save us from adding this identity.checkPermission everywhere when it's about to update the entity. Could be something for later.

huangp: In fact, we could just override the super.update() method in EntityHome. What do you think?

huangp:
:+1: for @seanf changes.

pom.xml
seanf: It might not be needed, but I'm just going by http://jglue.org/cdi-unit-user-guide/#weldVersions

zanata-war/src/main/java/org/zanata/action/ProjectHome.java
seanf: That (checkPermission in `EntityHome.update()`) sounds like a good refactoring, but for now I'd like to keep it to a dead simple translation where possible.

djansen-redhat:
I'm putting the monkey on this one.
:see_no_evil: (verified by Leeroy Jenkins)

seanf:
Merged as 7610015760c0503155ce01041e68b3f3f6812df6.

PR #890, opened 2015-06-30
Add explicit dependency on jms.

davidmason:
This was causing a deployment failure on JBoss EAP 6.3.3 with a root cause of

  java.lang.ClassNotFoundException: javax.jms.MessageListener
    from [Module "deployment.zanata.war:main" from Service Module Loader]

Related: https://issues.jboss.org/browse/WFLY-4642


seanf:
:+1:

PR #891, opened 2015-07-01
Refactor project settings

davidmason:
This just moves the content from 2 of the settings tabs to separate files. No template code should be any different.


aeng:
:+1: as long as the functional test passes.

djansen-redhat:
:see_no_evil:

PR #886, opened 2015-06-23
Migrate Seam 2 to CDI

seanf:
https://zanata.atlassian.net/browse/ZNTA-530

PR #894, opened 2015-07-06
Make method non-final (not allowed for CDI beans)

seanf:


seanf:
Moving the constructor changes to https://github.com/zanata/zanata-server/pull/901

seanf:

seanf:
I'm going to include this commit in https://github.com/zanata/zanata-server/pull/886 instead.

PR #900, opened 2015-07-07
Exclude copies from user stats (rhbz1235070 backport)

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1235070

Review is done in https://github.com/zanata/zanata-server/pull/888. Only `db.changelog-3.7.xml` has changed

Note: dev database will need to wipe out for the liquibase to work due to merges of 3.8 to 3.7 changelog.


seanf:
@aeng Any conflicts when backporting?

seanf:
I realise the code has been reviewed once, but I think the changes are bigger and more complex than they need to be, especially for a backport.

Why didn't we leave the old `getEntityType()` alone, and choose another name for the new method?  That's a big refactoring for a backport.  Perhaps `getCopiedEntityType()`?

It's not really clear what the new `getEntityType()` is (no javadoc; terse method name).  I see that its type is the enum `TranslationEntityType`, but I suppose it represents a CopiedEntityType or OriginalEntityType (or LinkedEntityType?).  The enum should have understandable enum values, not three-character ones: we just need to be able to get the appropriate 3-char string when mapping to and from the database.

In fact, there seems to be a lot of overlap between the old `EntityType` and the new `TranslationEntityType` - I think the data model will be simpler and more understandable if we use a single enum, rather than introduce a second, related but slightly different enum.

I don't think changing the above need affect the liquibase changesets.  I suggest creating another commit to improve this pull request: 
 * merging `TranslationEntityType` into `EntityType`,
 * bringing back the old `getEntityType()`, 
 * renaming the new method to `getCopiedEntityType()` or similar, 
 * and adding the appropriate Hibernate mapping for the three letter entity types.

Here's an example of mapping an enum (or multiple enums with a common interface) to an arbitrary string "label" in the database: https://gist.github.com/jedvardsson/96af1b86f78762ac12ef

Finally, the new commit can be cherry-picked for `integration/master`.

aeng:
So are you happy with this pull request to merge into release?

aeng:
`EntityType` is not persisted in database. It's for identifying the EntityType in `Activity` table. `TranslationEntityType` is the new type that actually persisted in database (3 char) in HTextFlowTarget and HTextFlowTargetHistory. Both of them are totally different.

aeng:
@seanf ignored all my previous comments. I will implement as suggested.

aeng:
@seanf done with the changes in latest commit https://github.com/zanata/zanata-server/commit/a97853e5468be054d13465bc93388c6ebbdc5585. Please review.

seanf:
:+1: Looks good, @aeng .  We'd better put the bug back ON_QA, I suppose.

aeng:
@seanf agree. I will get it QA before merge

djansen-redhat:
:white_check_mark: Tested

aeng:
Merged in release. https://github.com/zanata/zanata-server/commit/b2d7ea504ae8acee9fcbc87c5b528417e5e76aa1

seanf:
@aeng Did you cherry pick the last three commits for `integration/master` too?

aeng:
@seanf i cherry-pick the squash commits, https://github.com/zanata/zanata-server/commit/fd950955676d356e838fc2b654997a4c4386e622

seanf:
Okay, thanks @aeng

PR #901, opened 2015-07-07
Move constructor logic to @Create methods

seanf:
In both Seam and CDI, the bean's constructor will be invoked twice:
once for the proxy object, and once for the actual instance, so any
logic should be in @Create/@PostConstruct methods, not the constructor.

In Seam, @javax.annotation.PostConstruct methods appear to cause Seam to
treat the bean as an EJB, which means some injections don't happen before
the @PostConstruct method (or at all, possibly).  (These methods will be
migrated to @PostConstruct for CDI.)


seanf:

seanf:

carlosmunoz:
:+1: 
I would have imagined those constructors were being used in tests.

djansen-redhat:
This is fine.

PR #902, opened 2015-07-08
New permission for create project

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1214502

### DIFF
--- a/zanata-war/src/main/java/org/zanata/security/SecurityFunctions.java
+++ b/zanata-war/src/main/java/org/zanata/security/SecurityFunctions.java
@@ -85,7 +85,7 @@ public class SecurityFunctions {
      * Project ownership rules
      **************************************************************************/
 
-    /* Any authenticated user can create a project */
+    /* 'project-creator' can create project */
     @GrantsPermission(actions = "insert")
     public static boolean canCreateProject(HProject target) {
         return getIdentity().hasRole("project-creator");
--- a/zanata-war/src/main/java/org/zanata/security/SecurityFunctions.java
+++ b/zanata-war/src/main/java/org/zanata/security/SecurityFunctions.java
@@ -88,7 +88,7 @@ public class SecurityFunctions {
     /* Any authenticated user can create a project */
     @GrantsPermission(actions = "insert")
     public static boolean canCreateProject(HProject target) {
-        return getIdentity().isLoggedIn();
+        return getIdentity().hasRole("project-creator");
     }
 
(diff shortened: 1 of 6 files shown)


zanata-war/src/main/java/org/zanata/security/SecurityFunctions.java
```diff
@@ -88,7 +88,7 @@ public static boolean canManageRoles(String target) {
     /* Any authenticated user can create a project */
```
davidmason: Update comment to match the new behaviour.

zanata-war/src/main/java/org/zanata/util/EssentialDataCreator.java
```diff
@@ -106,6 +106,14 @@ public void prepare() {
                 adminExists = false;
             }
 
+            if (!accountRoleDAO.roleExists("project-creator")) {
+                log.info("Creating 'project-creator' role");
+                if (accountRoleDAO.create("project-creator", MANUAL, "user") == null) {
+                    throw new RuntimeException(
+                        "Couldn't create 'project-creator' role");
```
davidmason: "Could not" (avoid contractions in technical writing).

davidmason:
:+1:

zanata-war/src/main/java/org/zanata/util/EssentialDataCreator.java
```diff
@@ -101,11 +101,19 @@ public void prepare() {
...
             }
 
+            if (!accountRoleDAO.roleExists("project-creator")) {
+                log.info("Creating 'project-creator' role");
+                if (accountRoleDAO.create("project-creator", MANUAL, "user") == null) {
+                    throw new RuntimeException(
+                        "Could not create 'project-creator' role");
+                }
+            }
+
```
seanf: I think we are at (or past) the point where this bit of logic should be extracted into a method, called once for each role we want.  (Or a for loop.)

djansen-redhat:
:white_check_mark: Tested

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/7e90a9b74ecb888710e67bbd0a87d4058e00699a

seanf:
Shouldn't the squash merge 7e90a9b have a name like "feat(role): New permission for create project" instead of "feat: role: New permission for create project"?

aeng:
@seanf you're right.

PR #904, opened 2015-07-10
Webhook header needs to include cryptographic signature in header for identification

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1213630


zanata-model/src/main/java/org/zanata/model/HProject.java
```diff
@@ -177,6 +177,9 @@
     @Field
     private EntityStatus status = EntityStatus.ACTIVE;
 
+    @Size(max = 255)
+    private String webhookKey;
```
seanf: Why 255?  Will the UI have a validator which agrees with this maximum?

zanata-war/pom.xml
```diff
@@ -1382,6 +1382,13 @@
     </dependency>
 
     <dependency>
+      <groupId>org.mock-server</groupId>
+      <artifactId>mockserver-netty</artifactId>
+      <version>3.9.17</version>
+      <scope>test</scope>
+    </dependency>
```
seanf: Is this still needed?

zanata-war/src/main/java/org/zanata/events/DocumentMilestoneEvent.java
```diff
@@ -50,6 +47,7 @@
     private String version;
     private String docId;
     private LocaleId locale;
+    private String url;
```
seanf: Javadoc, please?

zanata-model/src/main/java/org/zanata/model/HProject.java
seanf: Perhaps `webhookSecret` would be clearer.  And/or a javadoc.

seanf: Oh, and could you please add `@Nullable`, just to be explicit?

zanata-war/src/main/java/org/zanata/service/impl/WebHooksPublisher.java
```diff
@@ -22,37 +22,69 @@
...
+            ResteasyClient client = new ResteasyClientBuilder().build();
+            ResteasyWebTarget target = client.target(url);
+
+            Invocation.Builder postBuilder =
+                    target.request().accept(acceptType);
+
+            String sha = signWebhookHeader(data, secretKey.get(), serverUrl);
+            if (secretKey.isPresent()) {
+                postBuilder.header(WEBHOOK_HEADER, sha);
+            }
```
seanf: Won't line 67 blow up if the secretKey is not present?  In any case, you don't need to generate the sha unless the secret is present.

zanata-war/src/main/java/org/zanata/service/impl/WebHooksPublisher.java
```diff
@@ -22,37 +22,69 @@
...
+            return postBuilder.post(Entity.entity(data, mediaType));
         } catch (Exception e) {
-            log.error("Error on webHooks post {}", e, url);
-            return null;
+            log.error("Error on webHooks post {}, {}", url, e);
```
seanf: The standard capitalisation seems to be Webhook rather than WebHook, and thus this log message should probably say either "Webhooks" or "webhooks".

zanata-war/src/main/java/org/zanata/util/HmacUtil.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private static final String HMAC_SHA1_ALGORITHM = "HmacSHA1";
+
+    private static final String HMAC_MD5_ALGORITHM = "HmacMD5";
+
+    private static final String HMAC_SHA256_ALGORITHM = "HmacSHA256";
+
+    public static String hmacSha1(final String key, final String valueToDigest)
+            throws IllegalArgumentException {
+        return hmacSha1(StringUtils.getBytesUtf8(key),
```
seanf: Maybe add a javadoc about the UTF-8 encoding.  

It's important that the character encoding used here, and the character encoding used when encoding our JSONType objects, are matching.  It's extremely hard to verify that by reading code, with the current implementation.

zanata-war/src/main/resources/db/changelogs/db.changelog-3.8.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd">
+
+  <changeSet id="1" author="aeng@redhat.com">
+    <comment>WebhookKey to HProject</comment>
+    <addColumn tableName="HProject">
+      <column name="webhookKey" type="varchar(255)">
```
seanf: I think webhookSecret would be a bit clearer.  It would be good to use that term (or at least "secret") consistently in the code.

zanata-war/src/main/webapp/WEB-INF/layout/project/settings-tab.xhtml
```diff
@@ -203,21 +210,39 @@
...
+      <div class="form--inline">
+        <input type="text" id="secretKeyInput" autocomplete="off" maxlength="255"
+          placeholder="#{msgs['jsf.project.WebHooks.secretKey.placeholder']}"
+          class="form--inline__input" value="#{projectHome.instance.webhookKey}"/>
+
+        <div class="form--inline__suffix">
+          <button id="add-webhook-key" class="form--inline__addon button--primary button loader"
+            onclick="addWebhookSecretKey(jQuery('#secretKeyInput').val())">
+            <span class="loader__label">#{msgs['jsf.project.AddWebhookSecretKey']}</span>
+          </button>
```
seanf: Is there an actual button for adding the key?  Github just has an optional "Secret" field when creating or updating a Webhook.

It might be good to link to a page (eg on docs.zanata.org) which explains our Webhook authentication scheme.

zanata-war/src/test/java/org/zanata/service/impl/WebHooksPublisherTest.java
```diff
@@ -0,0 +1,113 @@
...
+import static org.mockserver.model.HttpCallback.callback;
+import static org.mockserver.model.HttpRequest.request;
+import static org.mockserver.verify.VerificationTimes.exactly;
+
+/**
+ * @author Alex Eng <a href="mailto:aeng@redhat.com">aeng@redhat.com</a>
+ */
+public class WebHooksPublisherTest {
+
+    private int port = 8180;
```
seanf: Hard-coded port numbers are a pain in the neck when managing Jenkins jobs.  Can you bind to port "0", and then ask the server which port it actually received, for use in the client?

This test should probably be classified as an IT, not a unit test.

zanata-war/src/test/java/org/zanata/util/HmacUtilTest.java
```diff
@@ -0,0 +1,62 @@
...
+        assertThat(sha5).isEqualTo(sha6);
+
+        String sha7 = HmacUtil.hmacSha1(key2, data2);
+        String sha8 = HmacUtil.hmacSha1(key2, data2);
+        assertThat(sha7).isEqualTo(sha8);
+
+        assertThat(sha1).isNotEqualTo(sha3).isNotEqualTo(sha5)
+                .isNotEqualTo(sha7);
+    }
+}
```
seanf: I would recommend adding another test case with some precalculated HMACs, perhaps generated using Trello's JavaScript implementation.

zanata-model/src/main/java/org/zanata/model/HProject.java
aeng: Yes, UI has limit of 255 too.

zanata-war/pom.xml
aeng: Yup. Used to create mock server for testing

zanata-war/src/main/webapp/WEB-INF/layout/project/settings-tab.xhtml
aeng: Good idea. I've link the doc in the webhook heading.

zanata-war/src/main/java/org/zanata/util/HmacUtil.java
```diff
@@ -0,0 +1,90 @@
...
+        return Base64.encodeBase64String(rawHmac);
+    }
+
+    // Get an hmac_sha1 key from the raw key bytes
+    public static Mac getHmacSha1(final byte[] key) {
+        return getInitializedMac(HMAC_SHA1_ALGORITHM, key);
+    }
+
+    // Get an hmac_sha1 Mac instance and initialize with the signing key
+    public static Mac
```
seanf: I think these two methods should be private.

zanata-war/src/main/java/org/zanata/util/UrlUtil.java
```diff
@@ -120,24 +125,36 @@ public String versionUrl(String projectSlug, String versionSlug) {
                 + versionSlug;
     }
 
-    public String editorDocumentListUrl(String projectSlug, String versionSlug,
-            LocaleId targetLocaleId, LocaleId sourceLocaleId) {
-        return getContextPath() + "/webtrans/translate?project=" + projectSlug
+    public String editorDocumentListUrl(boolean fullPath, String projectSlug,
+                    String versionSlug, LocaleId targetLocaleId,
+                    LocaleId sourceLocaleId) {
```
seanf: I think it would be less confusing to add the extra parameter `fullPath` at the end.

zanata-war/src/main/java/org/zanata/events/DocumentMilestoneEvent.java
```diff
@@ -27,29 +27,58 @@
...
+     * Target locale id.
+     * {@link org.zanata.common.LocaleId}
+     */
     private LocaleId locale;
+
+    /**
+     * Editor url in target document.
+     * {@link org.zanata.util.UrlUtil#fullEditorDocumentUrl}
+     */
+    private String url;
```
seanf: Could we give this a more descriptive name, like `editorDocumentUrl`?  Do we link to the editor, or to the viewer?  Can it be visited by an anonymous user, or a non-translator?

zanata-war/src/main/java/org/zanata/events/DocumentMilestoneEvent.java
```diff
@@ -27,29 +27,58 @@
...
+
+    /**
+     * Message for milestone reached.
+     */
     private String milestone;
```
seanf: Is this the name of a milestone, or a message about reaching a milestone?  If it's the latter, the property name is a bit misleading.

zanata-war/src/main/java/org/zanata/service/impl/DocumentServiceImpl.java
```diff
@@ -266,12 +271,19 @@ private void processWebHookDocumentMilestoneEvent(
...
+                    String docUrl =
+                            urlUtil.fullEditorDocumentUrl(project.getSlug(),
+                                version.getSlug(), event.getLocaleId(),
+                                LocaleId.EN_US, document.getDocId());
+
                     DocumentMilestoneEvent milestoneEvent =
                             new DocumentMilestoneEvent(project.getSlug(),
                                     version.getSlug(), document.getDocId(),
-                                    event.getLocaleId(), message);
+                                    event.getLocaleId(), docUrl, message);
```
seanf: Perhaps the URL should have been added at the end of the field list in DocumentMilestoneEvent, (and thus the end of the constructor params).

zanata-war/src/main/java/org/zanata/service/impl/WebHooksPublisher.java
```diff
@@ -22,37 +22,70 @@
...
-            return request.post();
+            ResteasyClient client = new ResteasyClientBuilder().build();
+            ResteasyWebTarget target = client.target(url);
+
+            Invocation.Builder postBuilder =
+                    target.request().accept(acceptType);
+
+            if (secretKey.isPresent()) {
+                String sha =
+                        signWebhookHeader(data, secretKey.get(), serverUrl);
```
seanf: - [ ] We should sign with the webhook callbackURL, not Zanata's server URL.

zanata-war/src/main/java/org/zanata/events/DocumentMilestoneEvent.java
aeng: Its message. DocumentServiceImpl#documentStatisticUpdated

zanata-war/src/main/java/org/zanata/service/impl/WebHooksPublisher.java
```diff
@@ -22,37 +22,70 @@
...
+            log.error("Error on webhooks post {}, {}", url, e);
         }
+        return null;
     }
 
-    public static ClientResponse publish(@Nonnull String url,
-        @Nonnull JSONType event) {
-        return publish(url, event.getJSON(), MediaType.APPLICATION_JSON_TYPE,
-                MediaType.APPLICATION_JSON_TYPE);
+    private static String signWebhookHeader(String data, String key,
```
seanf: - [ ] This method could use a unit test, with some sample data, a sample webhook callback URL, and the expected result.

zanata-war/src/main/java/org/zanata/events/DocumentMilestoneEvent.java
aeng: Its editor url, and only logged in user can access to it.

seanf: @aeng So it's accessible to any logged-in user, but potentially read-only?

aeng: @seanf correct.

seanf:
:+1: Looks good, assuming tests pass.

djansen-redhat:
If a key is not specified:
> 2015-07-15T01:29:24,404Z ERROR [org.zanata.service.impl.WebHooksPublisher] (DefaultQuartzScheduler_Worker-3) Unable to generate hmac sha1 for  {"project":"djansen-test","version":"1.0","docId":"404.txt","locale":"am","editorDocumentUrl":"http://localhost:8080/zanata/webtrans/translate?project=djansen-test&iteration=1.0&localeId=am&locale=en-US#view:doc;doc:404.txt","milestone":"100% Translated","eventType":"org.zanata.events.DocumentMilestoneEvent"}http://requestb.in/w28gb8w2
2015-07-15T01:29:24,405Z ERROR [org.zanata.service.impl.WebHooksPublisher] (DefaultQuartzScheduler_Worker-3) Error on webhooks post http://requestb.in/w28gb8w2, java.lang.IllegalArgumentException: java.lang.IllegalArgumentException: Empty key

djansen-redhat:
Mention double SHA1 on base64 in docs

djansen-redhat:
:+1: :white_check_mark: Tested, just waiting for green light

seanf:
@aeng In the docs, I think it would be good to give a bit of pseudocode (or real code), demonstrating how the receiver can verify the header.  Perhaps something like this (please double check):

    String expectedHeader = base64(hmacSha1(base64(hmacSha1(body.getBytes() + callbackURL.getBytes(UTF8)))));
    if (expectedHeader.equals(request.getHeader("X-Zanata-Webhook"))) {
        processWebhook(body);
    }

aeng:
@seanf +1

seanf:
:+1: No need to run the tests again for a docs change, but @djansen-redhat, are you happy with the modified docs?

djansen-redhat:
@seanf @aeng looks good to me

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/bd33f08c933c8572725d0d8f131cd654aaee5b50

PR #906, opened 2015-07-13
Fix ProfileTest generate api key test

djansen-redhat:
It appears that there was a race condition with the alert disappearing
and the number of expected background functions.

### DIFF
--- a/functional-test/src/test/java/org/zanata/feature/language/JoinLanguageTeamTest.java
+++ b/functional-test/src/test/java/org/zanata/feature/language/JoinLanguageTeamTest.java
@@ -66,8 +66,9 @@ public class JoinLanguageTeamTest extends ZanataTestCase {
         assertThat(languagePage.getMemberUsernames())
                 .contains("translator")
                 .as("Translator is a listed member of the pl team");
-        assertThat(hasEmailRule.emailsArrivedWithinTimeout(1, 5,
+        assertThat(hasEmailRule.emailsArrivedWithinTimeout(1, 30,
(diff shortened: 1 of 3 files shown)


aeng:
:+1:

djansen-redhat:
@aeng I encountered another issue while checking this in, so I fixed them together

functional-test/src/test/java/org/zanata/feature/language/JoinLanguageTeamTest.java
```diff
@@ -66,8 +66,9 @@ public void translatorJoinsLanguageTeam() throws Exception {
         assertThat(languagePage.getMemberUsernames())
                 .contains("translator")
                 .as("Translator is a listed member of the pl team");
-        assertThat(hasEmailRule.emailsArrivedWithinTimeout(1, 5,
+        assertThat(hasEmailRule.emailsArrivedWithinTimeout(1, 30,
```
aeng: Any reason its 30 seconds now? will it increase when the jenkins load is heavy?

djansen-redhat: No, it's just a matter of being an external interface (mail) that may, occasionally, take a touch more than 5 seconds.  We wait 20-30 seconds for everything else, so I think this is ok.

aeng:
:+1:

djansen-redhat:
:white_check_mark: Verified by Leeroy Jenkins

PR #907, opened 2015-07-14
fix(copy version): refresh version entry in UI after copy version

aeng:
https://bugzilla.redhat.com/show_bug.cgi?id=1242679

### DIFF
--- a/zanata-war/src/main/java/org/zanata/action/VersionHome.java
+++ b/zanata-war/src/main/java/org/zanata/action/VersionHome.java
@@ -370,8 +370,8 @@ public class VersionHome extends SlugHome<HProjectIteration> implements
     }
 
     public void copyVersion() {
-        getInstance().setStatus(EntityStatus.READONLY);
         getInstance().setSlug(inputSlugValue);
+        getInstance().setStatus(EntityStatus.READONLY);
 
         // create basic version here
         HProject project = getProject();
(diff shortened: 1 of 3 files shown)


zanata-war/src/main/java/org/zanata/action/VersionHome.java
```diff
@@ -370,7 +370,6 @@ public String createVersion() {
     }
 
     public void copyVersion() {
-        getInstance().setStatus(EntityStatus.READONLY);
```
huangp: I thought we mark it as read only before the copy is finished. Has this changed?

aeng: The problem with that is when copy version completed, the UI still shows the version as "READONLY" due to the hibernate cache. 
Perhaps I should redirect to project page if user trying to access the version page? what do you think?

huangp: That may be better. Otherwise if someone get into editor or do a push, it may clobber the in progress data

aeng: I think to minimise the changes of security rules, its best that I restore the readonly status.

huangp:
:+1:

djansen-redhat:
@aeng So we couldn't update the UI after the RO lock is removed?

djansen-redhat:
:white_check_mark: Tested

aeng:
In this pull request, nope.

PR #908, opened 2015-07-14
Replace Seam Security

huangp:
- replace all ```@Restrict``` to use our own annotations and write up our own security interceptor
- replace all ```#{s:hasRole``` and ```#{s:hasPermission``` to use ```identity``` instead
- remove ```<restrict>``` and ```require-login``` from pages.xml
- replace seam IdentityManager and JpaIdentityStore 

Some TODOs:
- [x] javadoc on CheckPermission.value
- [x] check permission('seam.user', 'read') in UserSearch.loadUsers
- [x] UserSearch.loadUsers should use ZanataIdentityManager, not JpaIdentityStore directly


seanf:
@huangp Is this ready for review?

huangp:
@seanf not yet. I am working on not have ZanataIdentity extend seam Identity class.

huangp:
@seanf @carlosmunoz this is now ready for review

seanf:
As part of this I think we should eliminate as many Seam Security classes as we can, eg RunAsOperation, Principal, Subject, Credentials.  Where our classes are substantially based on Seam classses, please make sure they have header comments (ie at the top) explaining where the code came from.

seanf:
- [ ] There are still a number of hits when I grep for `import org.jboss.seam.security`.

seanf:
@huangp Let me know when you've finished making changes.  I'm going to have to review the complete set of changes all together.

seanf:

huangp:

seanf:

seanf:

zanata-model/src/main/java/org/zanata/EntityAction.java
```diff
@@ -0,0 +1,18 @@
+package org.zanata;
```
seanf: Why not something.security?  There's too much in `org.zanata` already.

zanata-model/src/main/java/org/zanata/Identity.java
```diff
@@ -0,0 +1,55 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+
+package org.zanata;
```
seanf: This should be in a security package too.

zanata-model/src/main/java/org/zanata/model/HAccountRole.java
```diff
@@ -74,6 +74,7 @@ public String getName() {
         return groups;
     }
 
+    // TODO [CDI] check whether we actually use this field (doesn't seem to in prod db and in code)
     @RoleConditional
     public boolean isConditional() {
```
seanf: @carlosmunoz Have you got any idea what `RoleConditional` does?  It seems to predate the role stuff you implemented for Fedora authentication.

zanata-war/src/main/java/org/zanata/action/DashboardAction.java
```diff
@@ -59,7 +60,8 @@
 
 @Name("dashboardAction")
 @Scope(ScopeType.PAGE)
-@Restrict("#{identity.loggedIn}")
+@ZanataSecured
+@CheckLoggedIn
```
seanf: - [ ] Is `@ZanataSecured` needed to make `@CheckLoggedIn` work?  
- [ ] Should we put `@Interceptors(ZanataSecurityInterceptor.class)` on the `@interface CheckLoggedIn` too?
- [ ] Is there a sanity test to make sure that classes which need `@ZanataSecured` actually have it?

huangp: * Is @ZanataSecured needed to make @CheckLoggedIn work? 

Yes that's how Seam interceptor works. You need to have a class level annotation to bind it to an/some interceptor(s) then the other annotations (CheckLoggedIn, CheckRole, CheckPermission) are just for that interceptor to perform actual work. Our in-house async framework also work the same way. Once migrated to CDI, we will need to migrate the interceptor to CDI interceptor. 

    * Should we put @Interceptors(ZanataSecurityInterceptor.class) on the @interface CheckLoggedIn too? 

No need to do that. 

    * Is there a sanity test to make sure that classes which need @ZanataSecured actually have it? 
No unfortunately. We could use scannotation or some other tools to scan the entire package on classpath, and check to see any of the class that has security annotations. But the cost may be too high considering our classpath. Unless you want to use external tools like ```find zanata-war/src/main/java/ -type f -name "*.java" | xargs grep "@CheckLoggedIn"``` and ```find zanata-war/src/main/java/ -type f -name "*.java" | xargs grep "@ZanataSecured"``` 

----- Original Message -----

seanf: >>    * Should we put @Interceptors(ZanataSecurityInterceptor.class) on the @interface CheckLoggedIn too?

> No need to do that.

But it would guarantee that a class-level `@CheckLoggedIn` would work, would it not?  Without having to remember the `@ZanataSecured`.

seanf: >>    * Is there a sanity test to make sure that classes which need @ZanataSecured actually have it?

> No unfortunately. We could use scannotation or some other tools to scan the entire package on classpath, and check to see any of the class that has security annotations. But the cost may be too high considering our classpath. Unless you want to use external tools like ```find zanata-war/src/main/java/ -type f -name "*.java" | xargs grep "@CheckLoggedIn"``` and ```find zanata-war/src/main/java/ -type f -name "*.java" | xargs grep "@ZanataSecured"```

I think we should have such a test.  Otherwise it's too easy to think we've secured a class when we haven't. Scannotation (or Reflections) only needs to check `org.zanata.*` packages.  I'm sure we have tests which run slower, to test less critical things.

huangp: It will require a different Interceptor or we change ZanataSecurityInterceptor to also accept ```@CheckLoggedIn```. I don't think it will give us much value. Especially it can be put on method and for those cases you still need to remember to put ```@ZanataSecured``` plus changing it once migrated to CDI. I am writing a test to verify the annotation now. 
----- Original Message -----

seanf: No, we won't need it if we have the annotation test.

seanf:
@huangp There are still several hits when I grep for import org.jboss.seam.security.

zanata-overlay/distros/eap-6/standalone/configuration/standalone-zanata.xml
```diff
@@ -409,7 +409,7 @@
                 </security-domain>
                 <security-domain name="zanata.internal">
                     <authentication>
-                        <login-module code="org.jboss.seam.security.jaas.SeamLoginModule" flag="required"/>
+                        <login-module code="org.zanata.security.jaas.InternalLoginModule" flag="required"/>
```
seanf: - [ ] Make sure you put this change in the release notes.  
- [ ] We will need to ensure this gets into Ansible too, at the right time.

zanata-model/src/main/java/org/zanata/model/HAccountRole.java
carlosmunoz: No idea to be honest. What does the javadoc say?

zanata-war/src/main/java/org/zanata/security/SmartEntitySecurityListener.java
```diff
@@ -2,58 +2,57 @@
...
- * EntityPermissionChecker.instance() when security is disabled, which leads to
+ * Disable permission checking when security is disabled, which solves the
  * problems if there is no active Seam application context (in tests).
  *
  * @author Sean Flanigan <a
  *         href="mailto:sflaniga@redhat.com">sflaniga@redhat.com</a>
  *
  */
-public class SmartEntitySecurityListener extends EntitySecurityListener {
+public class SmartEntitySecurityListener {
```
seanf: How are all the Pre/Post methods in this class triggered now?

zanata-war/src/main/java/org/zanata/security/SpNegoIdentity.java
```diff
@@ -54,9 +54,7 @@ public void authenticate() {
         ZanataIdentity identity =
                 ServiceLocator.instance().getInstance(ZanataIdentity.class);
         if (identity.isLoggedIn()) {
-            if (Events.exists()) {
-                Events.instance().raiseEvent(Identity.EVENT_ALREADY_LOGGED_IN);
-            }
+            getAlreadyLoggedInEvent().fire(new AlreadyLoggedInEvent());
```
seanf: How is this working in tests, if it doesn't check whether Event scope is active?  Did we not need the if check at all?

seanf:
- [ ] We should have more comments to make it clear (and give credit) when we have borrowed/adapted many methods or entire classes from Seam (look at the diff and you'll see what has been added):

- UserAction, ZanataRoleAction, ZanataRoleSearch, ZanataUserSearch, ZanataIdentity (could use more), InternalLoginModule, and possibly PasswordUtil.

This is part of the reason I wanted to keep the Seam-derived code somewhat separate from our code (then we could just have a comment saying that this abstract class is adapted from org.jboss.seam.security.Foo).  The Seam-derived code is still fairly generic security logic, even if it no longer uses annotations.

zanata-war/src/main/java/org/zanata/security/SpNegoIdentity.java
huangp: it's checked inside Event class itself. No need to check it when you fire it

zanata-war/src/main/java/org/zanata/security/SmartEntitySecurityListener.java
huangp: Same way. Pre and Post are JPA/hibernate annotation. It just delegate the check to ZanataIdentity as before. Except this line we need to revisit:
```
PersistenceProvider provider = PersistenceProvider.instance();
Class beanClass = provider.getBeanClass(entity);
```
I think we are safe to just use entity.getClass(). the code snippet above just cater for entity being an EJB. And in our case it's always just a plain hibernate entity.

zanata-model/src/main/java/org/zanata/model/HAccountRole.java
seanf: On 2015-07-30 13:17, Carlos Munoz wrote:
> In zanata-model/src/main/java/org/zanata/model/HAccountRole.java
> <https://github.com/zanata/zanata-server/pull/908#discussion_r35835150>:
> 
>> @@ -74,6 +74,7 @@ public String getName() {
>>          return groups;
>>      }
>>  
>> +    // TODO [CDI] check whether we actually use this field (doesn't seem to in prod db and in code)
>>      @RoleConditional
>>      public boolean isConditional() {
> 
> No idea to be honest. What does the javadoc say?
> 
> —
> <https://github.com/zanata/zanata-server/pull/908/files#r35835150>.
> 

Naturally, it says "Flags an entity field or method as representing
whether a role is conditional or not."  Isn't that enlightening?

Reading JpaIdentityStore.listGrantableRoles I suspect it could be used
to implement the FAS role, or perhaps project-dependent roles like
project-maintainer or project-translator.

I think we should leave it in for now.

-- 
Sean Flanigan

Principal Software Engineer
Red Hat

zanata-war/src/main/java/org/zanata/security/SmartEntitySecurityListener.java
seanf: On 2015-07-30 13:57, Patrick Huang wrote:
> In
> zanata-war/src/main/java/org/zanata/security/SmartEntitySecurityListener.java
> <https://github.com/zanata/zanata-server/pull/908#discussion_r35836443>:
> 
>>   * problems if there is no active Seam application context (in tests).
>>   *
>>   * @author Sean Flanigan <a
>>   *         href="mailto:sflaniga@redhat.com">sflaniga@redhat.com</a>
>>   *
>>   */
>> -public class SmartEntitySecurityListener extends EntitySecurityListener {
>> +public class SmartEntitySecurityListener {
> 
> Same way. Pre and Post are JPA/hibernate annotation. It just delegate
> the check to ZanataIdentity as before.

Oh, it's coming back to me now: this class works because it is listed in
orm.xml, not because it extends a special Seam class.

> Except this line we need to revisit:
> 
> |PersistenceProvider provider = PersistenceProvider.instance(); Class
> beanClass = provider.getBeanClass(entity); |
> 
> I think we are safe to just use entity.getClass(). the code snippet
> above just cater for entity being an EJB. And in our case it's always
> just a plain hibernate entity.

Okay, fair enough.

> 
> —
> <https://github.com/zanata/zanata-server/pull/908/files#r35836443>.
> 


-- 
Sean Flanigan

Principal Software Engineer
Red Hat

zanata-war/src/main/java/org/zanata/seam/interceptor/ZanataSecurityInterceptor.java
```diff
@@ -0,0 +1,240 @@
...
+            if (roleRestrictions == null) {
+                roleRestrictions = new HashSet<>();
+            }
+
+            roleRestrictions.add(role);
+        }
+
+        public void check() {
+            // TODO [CDI] revisit this
+            if (!ZanataIdentity.isSecurityEnabled()) {
```
seanf: Revisit what, exactly?  That line, or the whole method?

zanata-war/src/main/java/org/zanata/security/FacesSecurityEvents.java
```diff
@@ -34,6 +40,8 @@
  * Override the {@link org.jboss.seam.security.FacesSecurityEvents} component to
  * alter default values.
  *
+ * TODO [CDI] need to find a CDI replacement.
```
seanf: for what?

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
```diff
@@ -32,40 +44,87 @@
...
     public static final String JAAS_DEFAULT = "default";
+    public static final String ROLES_GROUP = "Roles";
 
     private static final long serialVersionUID = -5488977241602567930L;
 
+    protected static boolean securityEnabled = true;
+    // Seam Context variables
+    // TODO [CDI] revisit this
+    private static final String LOGIN_TRIED = "org.jboss.seam.security.loginTried";
+    private static final String SILENT_LOGIN = "org.jboss.seam.security.silentLogin";
```
seanf: Revisit the direct injection of booleans into Seam contexts, you mean?

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
```diff
@@ -101,24 +160,84 @@ public void checkLoggedIn() {
...
-            Events.instance().raiseEvent(Logout.EVENT_NAME,
-                    new Logout(getCredentials().getUsername()));
+        if (getCredentials() != null) {
+            getLogoutEvent().fire(new Logout(getCredentials().getUsername()));
+        }
+        if (isLoggedIn()) {
+            unAuthenticate();
+            Session.instance().invalidate();
+            // used by Seam's RememberMe and RuleBasedPermissionResolver
+//            if (Events.exists()) Events.instance().raiseEvent(Identity.EVENT_LOGGED_OUT);
```
seanf: In that case, it's time we deleted FixedRememberMe.

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
```diff
@@ -260,10 +616,66 @@ public String login(AuthenticationType authType) {
...
         return null;
     }
+
+    public Principal getPrincipal() {
+        return principal;
+    }
+
+    public void checkRestriction(String expression) {
+        log.warn("permission check as EL is not supported yet");
+        throw new UnsupportedOperationException("identity.checkRestriction(exp) is not supported");
```
seanf: - [ ] As far as I know, we don't plan to support EL for security checks at all, which means we shouldn't have any annotations which accept EL.  EL seems to be the purpose of `value` in `@CheckPermission`, but we hardly seem to be using `@CheckPermission` anywhere.  Is there some migration that we are missing?

seanf: > As far as I know, we don't plan to support EL for security checks at all

Right, @carlosmunoz ?

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
```diff
@@ -260,10 +616,66 @@ public String login(AuthenticationType authType) {
...
+
+            if (systemOp == null) {
+                systemOp = new ThreadLocal<>();
+            }
+
+            systemOp.set(operation.isSystemOperation());
+
+            operation.execute();
+        } finally {
+            systemOp.set(false);
```
seanf: I wonder if this ThreadLocal could leak memory, in an environment where worker threads weren't reused.  Do you think it would be safer to call `systemOp.remove()`?

seanf:
@carlosmunoz Could you please look over the following classes?  I want a second pair of eyes to review this code.  I think you'll find it easier to review if you check out the branch in your IDE.

- [ ] zanata-war/src/main/java/org/zanata/action/UserAction.java
- [ ] zanata-war/src/main/java/org/zanata/action/ZanataRoleAction.java
- [ ] zanata-war/src/main/java/org/zanata/action/ZanataRoleSearch.java
- [ ] zanata-war/src/main/java/org/zanata/action/ZanataUserSearch.java
- [ ] zanata-war/src/main/java/org/zanata/seam/interceptor/ZanataSecurityInterceptor.java
- [ ] zanata-war/src/main/java/org/zanata/seam/security/ZanataIdentityManager.java
- [ ] zanata-war/src/main/java/org/zanata/seam/security/ZanataJpaIdentityStore.java
- [ ] zanata-war/src/main/java/org/zanata/security/jaas/InternalLoginModule.java

seanf:
@huangp I've reviewed most of the code, but I still want to go through `pages.xml` where `<restrict>`s have been removed and verify the corresponding security checks in the JSF beans.

seanf:
- [ ] @huangp Remind me again why CustomPermissionResolver still extends a Seam class.

zanata-war/src/main/java/org/zanata/security/FacesSecurityEvents.java
huangp: This class extends org.jboss.seam.security.FacesSecurityEvents. Need to do it in deltaspike or something else. 

----- Original Message -----

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
huangp: Revisit the names and the usage. I can rename the value to not having seam package in it for now. But in CDI land, they need to be changed to become qualified objects. Maybe just leave them as is now. 

----- Original Message -----

zanata-war/src/main/java/org/zanata/security/FacesSecurityEvents.java
seanf: On 2015-07-31 09:29, Patrick Huang wrote:
> In zanata-war/src/main/java/org/zanata/security/FacesSecurityEvents.java
> <https://github.com/zanata/zanata-server/pull/908#discussion_r35933312>:
> 
>> @@ -34,6 +40,8 @@
>>   * Override the {@link org.jboss.seam.security.FacesSecurityEvents} component to
>>   * alter default values.
>>   *
>> + * TODO [CDI] need to find a CDI replacement.
> 
> This class extends org.jboss.seam.security.FacesSecurityEvents. Need to
> do it in deltaspike or something else.

It doesn't extend anything in my checkout of this branch.  Does this
comment still apply?

Actually, I think the class is entirely unused now.


-- 
Sean Flanigan

Principal Software Engineer
Red Hat

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
seanf: Okay, but please try to make TODO comments self-explanatory.  We have to read them too!

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
carlosmunoz: Correct.

huangp: You are right we don't use @CheckPermission anywhere. There is no @Restrict which uses EL in our code. I think at one stage we have migrated all those into explicit call to identity.checkPermission. 
CheckPermission.value can be used as permission action name in the future. See org/zanata/security/CheckPermissionDecisionVoter.java:59 
I've removed the method ZanataIdentity.checkRestriction(expression). But in ZanatatSecurityInterceptor, it will throw UnsupportedOperationException if someone tries to use the value field. Is this alright? 

----- Original Message -----

seanf: - [ ] @huangp In that case, we should get rid of the `value` parameter for `@CheckPermission` (or the whole annotation), along with methods like this one for handling it.

huangp:
No need to extend it anymore. Fixed now. 

----- Original Message -----

zanata-war/src/main/java/org/zanata/security/FacesSecurityEvents.java
huangp: Sorry my bad. It was extending the seam class and then I took it out. But it still uses seam classes in the implementation: 

import org.jboss.seam.international.StatusMessage; 
import org.jboss.seam.international.StatusMessages; 

It's still used as an event observer. I think it just puts some messages to jsf. @carlosmunoz may know better. 

----- Original Message -----

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
seanf: On 2015-07-31 09:51, Patrick Huang wrote:
> In zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
> <https://github.com/zanata/zanata-server/pull/908#discussion_r35934714>:
> 
>>          if (authenticatedAccount != null) {
>>              return authenticatedAccount.getUsername();
>>          }
>>          return null;
>>      }
>> +
>> +    public Principal getPrincipal() {
>> +        return principal;
>> +    }
>> +
>> +    public void checkRestriction(String expression) {
>> +        log.warn("permission check as EL is not supported yet");
>> +        throw new UnsupportedOperationException("identity.checkRestriction(exp) is not supported");
> 
> You are right we don't use @CheckPermission anywhere. There is no
> @Restrict <https://github.com/Restrict> which uses EL in our code. I
> think at one stage we have migrated all those into explicit call to
> identity.checkPermission. CheckPermission.value can be used as
> permission action name in the future. See
> org/zanata/security/CheckPermissionDecisionVoter.java:59 I've removed
> the method ZanataIdentity.checkRestriction(expression). But in
> ZanatatSecurityInterceptor, it will throw UnsupportedOperationException
> if someone tries to use the value field. Is this alright?

I'm not sure if it's worth keeping around as dead code, but if we do
want to keep it, we should document it properly (eg javadoc on
CheckPermission.value), and remove the bits which won't ever make sense,
like this method.

-- 
Sean Flanigan

Principal Software Engineer
Red Hat

zanata-war/src/main/java/org/zanata/action/UserAction.java
```diff
@@ -135,22 +135,22 @@ public String getName(String username) {
...
 
-    @Override
     @Begin
     public void editUser(String username) {
-        super.editUser(username);
+        this.username = username;
+        roles = identityManager.getGrantedRoles(username);
+        enabled = identityManager.isUserEnabled(username);
+        newUserFlag = false;
         newUserFlag = false;
```
carlosmunoz: Same line twice

zanata-war/src/main/java/org/zanata/action/ZanataRoleAction.java
```diff
@@ -0,0 +1,131 @@
...
+ * @author Patrick Huang
+ *         <a href="mailto:pahuang@redhat.com">pahuang@redhat.com</a>
+ */
+@Name("zanataRoleAction")
+@Scope(CONVERSATION)
+@Install(precedence = APPLICATION)
+@ZanataSecured
+@CheckLoggedIn
+@CheckRole("admin")
+public class ZanataRoleAction implements Serializable {
```
carlosmunoz: I know I'm late to the review, and hate to bring this up late, but is there any chance we can stop naming this `Zanata...`? IMO the package name should tell you that.

huangp: It's named that way because it used to extend seam's RoleAciton (just so we don't have to use fully qualified class). It can be renamed to RoleAction now. But I will leave it after all the review has been done otherwise the diff will look horrible. 

----- Original Message -----

zanata-war/src/main/java/org/zanata/security/FacesSecurityEvents.java
seanf: Okay.

We don't need to do it for this PR, but how are we going to replace StatusMessage(s)?  From the class name, it sounds like it should overlap with FacesMessages, but it doesn't seem to.  I can't work out what it actually does with the messages after interpolating the variables.  Wait: what it does is add them to the global `messages` list, Seam's FacesMessages extends StatusMessages, and it displays any waiting global messages in `beforeRenderResponse()`.  

So FacesSecurityEvents could generate the status messages and call our `FacesMessages.addGlobal` to achieve the same effect.  

- [ ] @huangp If you agree, could you please put in a TODO to that effect?

seanf:
> I looked up the sed script again.  So `#{s:hasPermission...}` will be replaced by `@CheckPermission`.   Mind putting that fact in?  It helps to explain why we're putting in all this unused code.

- [ ] @huangp Is that migration still going to work, or did you already replace `s:hasPermission` with something else?

huangp:
s:hasPermission has been replaced with identity:hasPermission in jsf and possibly in pages.xml. There is no code uses something like @Restrict("s:hasPermission"). 

----- Original Message -----

zanata-war/src/main/webapp/WEB-INF/pages.xml
```diff
@@ -24,15 +24,10 @@
...
 
-  <page view-id="/dashboard/*" login-required="true">
-    <restrict>#{identity.loggedIn}</restrict>
-  </page>
-
   <page view-id="/edit_home_content.xhtml">
     <action execute="#{breadcrumbs.clear}" />
     <action
       execute="#{breadcrumbs.addLocation('', msgs['jsf.server.EditHomePage.label'])}" />
-    <restrict>#{s:hasRole('admin')}</restrict>
```
seanf: Where did this check move to?  I can see `@CheckRole("admin")` on ServerConfigurationBean; is there anything else?

zanata-war/src/main/webapp/WEB-INF/pages.xml
```diff
@@ -136,13 +131,12 @@
     <action execute="#{activate.activate}" />
   </page>
 
-  <page view-id="/account/email_validation.xhtml" login-required="true">
+  <page view-id="/account/email_validation.xhtml">
```
seanf: Shouldn't ValidateEmailAction have CheckLoggedIn or something?

seanf:
@huangp In that case, are we really expecting to use @CheckPermission at all, even with CDI?

zanata-war/src/main/webapp/WEB-INF/pages.xml
seanf: Oh, I see there's a check in a Create method.  

- [ ] Just to be doubly safe, I wouldn't mind seeing the annotations too.  Perhaps we could intercept bean creation in CDI with an extension.

zanata-war/src/main/java/org/zanata/action/ValidateEmailAction.java
```diff
@@ -60,6 +61,11 @@
     @In
     EmailChangeService emailChangeService;
 
+    @Create
+    public void onCreate() {
+        identity.checkLoggedIn();
```
seanf: @carlosmunoz Can you think of some way (in Seam, or in CDI) of intercepting bean creation, controlled by a class-level annotation like `@CheckLoggedIn`?  The idea here is to prevent the JSF page from loading with an authorisation exception, similar to what we used to do in pages.xml.  The only thing I can think of is a CDI extension, but I'm not sure how much work it would be.

carlosmunoz: So, isn't this line doing exactly that? I thought a failed `<restrict>` in `pages.xml` would result in an authorization exception as well.

But, to answer your question, nothing really comes to mind. CDI interceptors can be placed at the class level, in which case they will apply to all methods, but I'm fairly certain this doesn't include the creation of the bean.

seanf: On 2015-07-31 13:18, Carlos Munoz wrote:
> In zanata-war/src/main/java/org/zanata/action/ValidateEmailAction.java
> <https://github.com/zanata/zanata-server/pull/908#discussion_r35943963>:
> 
>> @@ -60,6 +61,11 @@
>>      @In
>>      EmailChangeService emailChangeService;
>>  
>> +    @Create
>> +    public void onCreate() {
>> +        identity.checkLoggedIn();
> 
> So, isn't this line doing exactly that?

It is, it's just not very declarative.


> I thought a failed |<restrict>|
> in |pages.xml| would result in an authorization exception as well.

I think it will, for now, but we're trying to move away from it in
preparation for Seam->CDI.

> But, to answer your question, nothing really comes to mind. CDI
> interceptors can be placed at the class level, in which case they will
> apply to all methods, but I'm fairly certain this doesn't include the
> creation of the bean.

This page mentions @PostConstruct for interceptor classes:
  https://docs.oracle.com/javaee/6/tutorial/doc/gkigq.html

and this one mentions @PostCreate for interceptor classes:
  http://docs.oracle.com/javaee/6/tutorial/doc/gkedm.html

but I'm not sure if we can use either or both of them with plain CDI.

-- 
Sean Flanigan

Principal Software Engineer
Red Hat

zanata-war/src/main/java/org/zanata/action/ZanataRoleAction.java
seanf: Seam's RoleAction wasn't restricted to admin.  Why is ZanataRoleAction?

zanata-war/src/main/webapp/WEB-INF/pages.xml
```diff
@@ -218,7 +203,6 @@
   </page>
 
   <page view-id="/admin/usermanager.xhtml" action="#{userSearch.loadUsers}">
-    <restrict>#{s:hasPermission('seam.user', 'read')}</restrict>
```
seanf: I can't find this security check in the UserAction bean.

zanata-war/src/main/webapp/WEB-INF/pages.xml
huangp: The page itself is secured by jsf in webapp/home.xhtml:19. Basically hidden from user without admin role. 

----- Original Message -----

zanata-war/src/main/webapp/WEB-INF/pages.xml
```diff
@@ -262,8 +246,9 @@
     <action execute="#{breadcrumbs.addLocation('', msgs['jsf.Groups'])}" />
   </page>
 
-  <page view-id="/version-group/create_version_group.xhtml">
-    <restrict>#{s:hasPermission(versionGroupHome.instance, 'insert')}</restrict>
+  <!--TODO [CDI] we need to find a way to secure this page's access in CDI (the actual perist method is already secured) -->
+  <page view-id="/version-group/create_version_group.xhtml"
+    action="#{identity.checkPermission(versionGroupHome.instance, 'insert')}">
```
seanf: Could we split up VersionGroupHome, put all the mutating operations into their own class, and make sure only the group maintainer pages use the mutating bean?

zanata-war/src/main/java/org/zanata/action/ZanataRoleAction.java
huangp: RoleAction is used by roledetail.xhtml and rolemanager.xhtml. They are both under admin/ which were restricted to admin only. Same reason for UserAction. 

----- Original Message -----

zanata-war/src/main/webapp/WEB-INF/pages.xml
huangp: All the user management work in UserAction is delegated to ZanataIdentityManager which will check all the security in its methods. But I do forgot to secure userSerach.loadUsers. it should also delegate to ZanataIdentityManager not the JpaIdentityStore directly. 
----- Original Message -----

huangp:
I think it makes declarative security easier. We can decide it once migrated to CDI 

----- Original Message -----

zanata-war/src/main/java/org/zanata/action/ZanataRoleAction.java
seanf: On 2015-07-31 15:25, Patrick Huang wrote:
> In zanata-war/src/main/java/org/zanata/action/ZanataRoleAction.java
> <https://github.com/zanata/zanata-server/pull/908#discussion_r35947306>:
> 
>> +import org.zanata.security.annotations.ZanataSecured;
>> +
>> +import static org.jboss.seam.ScopeType.CONVERSATION;
>> +import static org.jboss.seam.annotations.Install.APPLICATION;
>> +
>> +/**
>> + * @author Patrick Huang
>> + *         <a href="mailto:pahuang@redhat.com">pahuang@redhat.com</a>
>> + */
>> +@Name("zanataRoleAction")
>> +@Scope(CONVERSATION)
>> +@Install(precedence = APPLICATION)
>> +@ZanataSecured
>> +@CheckLoggedIn
>> +@CheckRole("admin")
>> +public class ZanataRoleAction implements Serializable {
> 
> RoleAction is used by roledetail.xhtml and rolemanager.xhtml. They are
> both under admin/ which were restricted to admin only. Same reason for
> UserAction.

I don't think that current decision should be hard-coded at the level of
RoleAction and UserAction.  A lot of users who are currently admins
shouldn't be, so I can see these permissions being separated from admin
in future.


-- 
Sean Flanigan

Principal Software Engineer
Red Hat

huangp: So what's your suggestion? Should I remove that check from ZanataRoleAction, ZanataRoleSearch and ZanataUserAction, ZanataUserSearch? They are secured by /admin/* in pages.xml from being accessed. They are also secured by identity manager. and of course the SmartEntityListener at the lowest level. If we want to change the permission rule, all these layers needs to change. 

----- Original Message -----

seanf:
Okay, so in the post-CDI refactoring, we might change methods like this:

void someMethod(Entity someEntity) {
  identity.checkPermission(someEntity, "some-permission")
  // do something
}


into this:

@CheckPermission("some-permission")
void someMethod(@PermissionTarget Entity someEntity) {
  // do something
}

Unfortunately, it seems like the relevant permission target is nearly
always a field of the bean (eg ProjectHome.instance), not a parameter of
the method, so I'm not sure how much use @CheckPermission will really
get.  Any thoughts, Carlos?

Anyway, let's leave it in for now, but please put some javadoc on
CheckPermission.value() to explain that it is the name of the permission
being checked.


On 2015-07-31 15:36, Patrick Huang wrote:
> I think it makes declarative security easier. We can decide it once
> migrated to CDI
> 
> ----- Original Message -----
> 
>> From: "Sean Flanigan" <notifications@github.com>
>> To: "zanata/zanata-server" <zanata-server@noreply.github.com>
>> Cc: "Patrick Huang" <pahuang@redhat.com>
>> Sent: Friday, July 31, 2015 11:39:42 AM
>> Subject: Re: [zanata-server] Replace Seam Security (#908)
> 
>> @huangp In that case, are we really expecting to use @CheckPermission
> at all,
>> even with CDI?
> 
>> —
>> Reply to this email directly or view it on GitHub .
> 
> —
> <https://github.com/zanata/zanata-server/pull/908#issuecomment-126571236>.
> 


-- 
Sean Flanigan

Principal Software Engineer
Red Hat

zanata-war/src/main/java/org/zanata/action/ZanataRoleAction.java
seanf: On 2015-07-31 15:54, Patrick Huang wrote:
> In zanata-war/src/main/java/org/zanata/action/ZanataRoleAction.java
> <https://github.com/zanata/zanata-server/pull/908#discussion_r35948181>:
> 
>> +import org.zanata.security.annotations.ZanataSecured;
>> +
>> +import static org.jboss.seam.ScopeType.CONVERSATION;
>> +import static org.jboss.seam.annotations.Install.APPLICATION;
>> +
>> +/**
>> + * @author Patrick Huang
>> + *         <a href="mailto:pahuang@redhat.com">pahuang@redhat.com</a>
>> + */
>> +@Name("zanataRoleAction")
>> +@Scope(CONVERSATION)
>> +@Install(precedence = APPLICATION)
>> +@ZanataSecured
>> +@CheckLoggedIn
>> +@CheckRole("admin")
>> +public class ZanataRoleAction implements Serializable {
> 
> So what's your suggestion?

I suggest we keep it as close to the Seam status quo as possible, for now.

> Should I remove that check from
> ZanataRoleAction, ZanataRoleSearch and ZanataUserAction,
> ZanataUserSearch?

Unless they or their Seam relatives had admin checks before, yes.

> They are secured by /admin/* in pages.xml from being
> accessed.

And that's still fine for the moment.  It's just an extra layer of
defence for what we currently consider "admin" pages.

As your TODO in pages.xml says, we will need a CDI option for this.
That might turn out to be putting @CheckRole("admin") on a lot of beans
like ZanataRoleAction, but I hope it doesn't come to that.  Not yet, anyway.


> They are also secured by identity manager. and of course the
> SmartEntityListener at the lowest level.

I think SmartEntitySecurityListener and ZanataIdentity should perhaps
check the same permissions[*] (based on EntityAction, not strings like
"create" and "update"), but I don't want to take on refactoring Seam
just yet, so let's just keep it the same as before.

[*] Then again, checking the same permissions over and over again is a
bit pointless and high maintenance.

I suppose as a general rule, we really want to check permissions at the
lowest practical abstraction level, to avoid the danger of extra code
paths which fail to check permissions.  For basic CRUD stuff, that
probably means checking at the entity level.  However, for the sake of
being user friendly, we need to perform early checks of the same
permissions at the front end (UI, perhaps REST too), so that we can turn
off (grey out or hide) disallowed features in the UI.

Carlos, any advice on this?


> If we want to change the
> permission rule, all these layers needs to change.

Only the layers which check explicitly for the admin role.  Otherwise,
we just have to make sure the relevant permissions are granted to the
right users.


-- 
Sean Flanigan

Principal Software Engineer
Red Hat

seanf:
@huangp I've reviewed everything for now.  Don't forget to use David's script http://userscripts-mirror.org/scripts/show/455500 to keep track of the open issues.

huangp:
@seanf @carlosmunoz build passed

seanf:
Why doesn't ZanataUserSearch.loadUsers use hasPermission('seam.user', 'read'), instead of CheckRole(admin)?

zanata-war/src/main/java/org/zanata/security/FacesSecurityEvents.java
seanf: @huangp What happened with this?

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
seanf: - [ ] Please put in a TODO comment at least, unless you think I'm wrong about this.

huangp:
It has it in IdentityManager. And the checkRole('admin') is for any jsf beans that are under admin/* 

----- Original Message -----

zanata-overlay/distros/eap-6/standalone/configuration/standalone-zanata.xml
huangp: In Release notes now

seanf:
@huangp looks like you forgot to delete the empty test `ordinaryUserDoNotHaveAdminPermission`

zanata-war/src/main/java/org/zanata/security/FacesSecurityEvents.java
huangp: Comment added

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
huangp: Commend added

seanf:
- [ ] @huangp What happened to the `identity.checkPermission("seam.role", "read")` which used to be in `ZanataRoleAction.onCreate`?

huangp:
" Seam's RoleAction wasn't restricted to admin. Why is ZanataRoleAction?" This is one of your previous question and you suggested to keep it close to the Seam quo. 

----- Original Message -----

seanf:
All looks good, @huangp !

seanf:
- [ ] @huangp Did you still plan to rename some of the packages or classes?

huangp:
Renamed all the newly created classes.

zanata-model/src/main/java/org/zanata/model/HAccount.java
```diff
@@ -108,7 +102,7 @@ public HPerson getPerson() {
     }
 
     @NaturalId
-    @UserPrincipal
+//    @UserPrincipal
```
djansen-redhat: Was there a reason to keep this/these?

seanf: I assumed it was in case we switched to a more generic solution (driven by
annotations) again in future, to make it easier to put in some annotations
again.

On 7 August 2015 at 13:42, Damian Jansen <notifications@github.com> wrote:

> In zanata-model/src/main/java/org/zanata/model/HAccount.java
> <https://github.com/zanata/zanata-server/pull/908#discussion_r36488926>:
>
> > @@ -108,7 +102,7 @@ public HPerson getPerson() {
> >      }
> >
> >      @NaturalId
> > -    @UserPrincipal
> > +//    @UserPrincipal
>
> Was there a reason to keep this/these?
>
> —
> <https://github.com/zanata/zanata-server/pull/908/files#r36488926>.
>



-- 

-- 
Sean Flanigan

Principal Software Engineer
Red Hat

djansen-redhat: Probably best to note that alongside the commenting. But that's just nitpicking.

djansen-redhat:
As a normal user, trying to create a version for their own project:
http://pastebin.test.redhat.com/303648

djansen-redhat:
:white_check_mark: Tested

huangp:
squashed and merged 9d0387e55e5d21439555d5f18c44cb9326175541

PR #910, opened 2015-07-15
fix(gwt): avoid packaging .junit_symbolMaps/

seanf:
This should reduce the size of zanata.war, and probably speed up builds
a bit too.


davidmason:
@seanf I probably would have classified this as type "chore" since it looks like it is primarily a build change. What do you think?

seanf:
Hmm, could be either, the unnecessary ~25MB of .war file could be viewed as a non-functional bug.  It's not just changing the build process, it's also changing the artifact.

seanf:
@davidmason Could you review this PR, please?

davidmason:
:+1:

PR #921, opened 2015-07-22
fix(create-project): update docs, add permission check for create project

aeng:
Fix from bug found in demo, 
- update documentation
- add permission check in create-project button in dashboard

### DIFF
--- a/zanata-war/src/main/webapp/WEB-INF/layout/dashboard/projects.xhtml
+++ b/zanata-war/src/main/webapp/WEB-INF/layout/dashboard/projects.xhtml
@@ -26,15 +26,17 @@
               <span class="is-invisible">#{msgs['jsf.dashboard.projects.maintainedProjects.label']}</span>
             </a>
           </li>
-          <li class="tab__end">
-            <a href="#" jsfc="s:link"
-              id="create-project-link"
-              view="/project/create_project.xhtml"
-              title="#{msgs['jsf.dashboard.projects.createNewProject.label']}">
-              <i aria-hidden="true" class="i i--plus i i--large"></i>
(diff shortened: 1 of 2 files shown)


zanata-war/src/main/webapp/WEB-INF/layout/dashboard/projects.xhtml
```diff
@@ -26,15 +26,17 @@
...
-          <li class="tab__end">
-            <a href="#" jsfc="s:link"
-              id="create-project-link"
-              view="/project/create_project.xhtml"
-              title="#{msgs['jsf.dashboard.projects.createNewProject.label']}">
-              <i aria-hidden="true" class="i i--plus i i--large"></i>
-              <span class="is-invisible">#{msgs['jsf.dashboard.projects.newProject.label']}</span>
-            </a>
-          </li>
+          <s:fragment rendered="#{s:hasPermission(projectHome.instance, 'insert')}">
```
huangp: Please use identity.hasPermission instead of s:hasPermissin. In the CDI migration all such seam specific EL needs to be migrated

huangp:
:+1:

davidmason:
@aeng just a note on commit naming: this one could have been done as 2 commits: docs(...) and fix(...). It would just mean we could more easily look at when all the docs changes happened, if we ever wanted to do that.

seanf:
Normally, I would say if you change something and need to document that change, it should be one commit, and should be listed as a `fix(...)`, but in this case, it's a fix for one thing, and a docs change for something else, so it would be better as two commits (at least if they weren't such small commits).

If you feel the need to say the commit is two different things, it may be a sign that it should be two commits.

PR #923, opened 2015-07-27
Improve TM results

seanf:
cf. https://bugzilla.redhat.com/show_bug.cgi?id=1086459

This PR *will not* change the TM sort order in the editor, but it will change the underlying search so that (a) more of the top matches are considered (20 instead of 10), and (b) the underlying search will prefer newer results where the Lucene score is identical.  Thus if there are multiple 100% text matches with the same score, the newest ones should make the cut.

(This will probably need to be backported to `release`.)


zanata-war/src/main/java/org/zanata/util/SysProperties.java
```diff
@@ -0,0 +1,53 @@
...
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package org.zanata.util;
+
+/**
+ * @author Sean Flanigan <a href="mailto:sflaniga@redhat.com">sflaniga@redhat.com</a>
+ */
+public class SysProperties {
```
carlosmunoz: Although for the most part I can read what this is doing, I really like the idea of having every class have a description (however short) of what it aims to do.

seanf:
Thanks @carlosmunoz; I've added javadocs to that class.

carlosmunoz:
:+1:

djansen-redhat:
:white_check_mark: Tested

PR #924, opened 2015-07-27
Boost Lucene query using TFT metadata

seanf:
This PR builds on top of #923, by changing Lucene's scoring algorithm to favour results which share some or all of the same:

* TextFlowTarget (id)
* Project (slug)
* Iteration (slug)
* Document (docId)
* TextFlow (resId)

### DIFF
--- a/zanata-war/src/main/java/org/zanata/service/impl/TranslationMemoryServiceImpl.java
+++ b/zanata-war/src/main/java/org/zanata/service/impl/TranslationMemoryServiceImpl.java
@@ -21,9 +21,13 @@
 
 package org.zanata.service.impl;
 
-import java.util.*;
+import static com.google.common.collect.Collections2.filter;
+
+import javax.enterprise.inject.Alternative;
 
 import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+
 import org.apache.commons.lang.StringUtils;
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.index.Term;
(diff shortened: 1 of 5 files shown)


seanf:
Note: only the last commit needs to be reviewed, because the earlier commits are part of #923.

seanf:

seanf:

zanata-war/src/main/java/org/zanata/service/impl/TranslationMemoryServiceImpl.java
```diff
@@ -90,6 +101,27 @@
...
+            SysProperties.TM_BOOST_PROJECT, 2.0f);
+
+    private static final float BOOST_DOCID = SysProperties.getFloat(
+            SysProperties.TM_BOOST_DOCID, 1.5f);
+
+    private static final float BOOST_RESID = SysProperties.getFloat(
+            SysProperties.TM_BOOST_RESID, 1.5f);
+
+    private static final float BOOST_ITERATION = SysProperties.getFloat(
+            SysProperties.TM_BOOST_ITERATION, 1.0f);
```
carlosmunoz: Does this mean that we can now configure the boosting properties via System props? If so, do we want to document it?

seanf: > Does this mean that we can now configure the boosting properties via System props? If so, do we want to document it?

@carlosmunoz It should be configurable now, in emergencies, but I don't think we want to make it a supported feature.  You shouldn't be messing with these boost values unless you have read this source code, or under advice from someone who has.

carlosmunoz: Ok. :+1:

djansen-redhat:
:white_check_mark: Tested - I believe this works...

PR #933, opened 2015-08-19
New Project Roles  ZNTA-524

davidmason:
I have done an initial review myself and fixed everything I noticed, other than a couple of places with FIXME/TODO comments that I will get to soon.

https://zanata.atlassian.net/browse/ZNTA-524

### DIFF
--- a/zanata-war/src/main/java/org/zanata/action/ProjectHomeAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/ProjectHomeAction.java
@@ -1,23 +1,22 @@
 /*
+ * Copyright 2014, Red Hat, Inc. and individual contributors as indicated by the
+ * @author tags. See the copyright.txt file in the distribution for a full
+ * listing of individual contributors.
  *
- *  * Copyright 2014, Red Hat, Inc. and individual contributors as indicated by the
- *  * @author tags. See the copyright.txt file in the distribution for a full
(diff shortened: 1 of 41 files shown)


davidmason:
@huangp do you have time to do an initial review on this?

I still have 1 modal to add, but most of its code will be identical to the existing modal so it should be an easy follow-up review.

zanata-war/src/main/java/org/zanata/action/ProjectHomeAction.java
```diff
@@ -438,4 +489,451 @@ public void resetPageData() {
...
+            } else {
+                setShowMembersInGroup(false);
+                return o1.getName().toLowerCase()
+                    .compareTo(o2.getName().toLowerCase());
+            }
+
+            // FIXME aeng: this is either confusing, or it is a logic error.
+            //       If the sort option is ROLE, this method will always return 0.
+            //       Please add an explanation of why this is, or fix the logic.
+            return 0;
```
davidmason: Can anyone figure out why this should behave this way? I suspect that maybe people are purposely not sorted when they are grouped, but it does not say so anywhere (and I would expect the people to be sorted alphabetically within groups).

davidmason:
Closing to recreate based on latest integration-master (diff is showing some of the changes from master).

PR #940, opened 2015-08-24
fix(ZNTA-640): prevent NullPointerException and provide more contextual information

davidmason:
https://zanata.atlassian.net/browse/ZNTA-640

There are two possible things that can be `null`, probably the latter since it has not been around as long. I'm logging both just in case.

I am not sure if it is appropriate to use an empty string instead of a valid abbreviation to prevent the exception. @aeng you wrote `TranslationSourceType` which is probably `null` here, what do you think?

---

This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/940)

### DIFF
--- a/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java
+++ b/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java
@@ -1002,13 +1002,27 @@ public class TranslationServiceImpl implements TranslationService {
                         // generate request
                         List<String> oldContents = oldTarget.getContents();
                         ContentState oldState = oldTarget.getState();
+
+                        final EntityType copiedEntityType = oldTarget.getCopiedEntityType();
+                        if (copiedEntityType == null) {
+                            log.error(
+                                    "Attempting to create a TransUnitUpdateRequest but copiedEntityType is null");
+                        }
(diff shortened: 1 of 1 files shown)


definite:
@zanata-jenkins  retest this please

definite:

seanf:
If possible, someone who knows this code should really put Nullable/Nonnull annotations on these values, and trace them back to their source.  If null is okay, we should accept it without logging an error.

I think @aeng wrote the affected lines. @aeng Does it make sense for these values (copiedEntityType, sourceType) to allow null?

seanf:
@davidmason I think JIRA only recognises the branch name if you use all caps, eg ZNTA-640-something.  As an alternative, you could amend the commit comment to "fix(ZNTA-640): etc etc"

huangp:
I just reproduced this. The ```oldTarget.getCopiedEntityType()``` is null when I did a search and replace undo (the translation have null copied entity type). ```oldTarget.getCopiedEntityId()``` is null too but since the argument is type Long not long so it's still okay. We need to be careful as well because if autobox happen somewhere down the track it will also generate NPE.

davidmason:
I'll see if I can reproduce and then fix the search-replace-undo issue.

davidmason:
This fix has since been implemented in a slightly different way. Closing since it is no longer relevant.

PR #947, opened 2015-08-26
fix(ZNTA-637): improve error messages

seanf:
https://zanata.atlassian.net/browse/ZNTA-637


aeng:
:+1:

djansen-redhat:
Should I expect multiple emails with an exception event?

djansen-redhat:
Actually, I may be getting multiple exceptions with the error I'm using to test...
:white_check_mark: Tested.

PR #955, opened 2015-08-31
chore: use Docker builds for Travis, with caching

seanf:


seanf:
@carlosmunoz I think we should backport this to `release` too, so that `release` PR builds will work in Travis.

seanf:
Backported as 3d0ee20d46bb29821a308d7a173558c1dc70b2d7

PR #958, opened 2015-09-04
fix: Make ResourceUtils.getNumPlurals more consistent

seanf:
We should always fall back on locale defaults, if available, whether
the plurals header is absent, empty or unparseable.

### DIFF
--- a/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java
+++ b/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java
@@ -79,6 +79,7 @@ import com.google.common.base.Optional;
 
 import static org.apache.commons.lang.StringUtils.isBlank;
 import static org.apache.commons.lang.StringUtils.isEmpty;
+import static org.apache.commons.lang.StringUtils.isNotEmpty;
 
 @Named("resourceUtils")
 @RequestScoped
@@ -108,6 +109,7 @@ public class ResourceUtils {
     private static final String CONTENT_TYPE_HDR = HeaderFields.KEY_ContentType;
     private static final String PLURAL_FORMS_HDR = "Plural-Forms";
 
+    // TODO we should need only one regex, with a capture group for nplurals
     private static final Pattern PLURAL_FORM_PATTERN =
         Pattern.compile("nplurals=[0-9]+;\\s?plural=*");
     private static final Pattern NPLURALS_TAG_PATTERN = Pattern
@@ -985,7 +987,7 @@ public class ResourceUtils {
     public String getPluralForms(LocaleId localeId, boolean checkDB, boolean useDefault) {
         if(checkDB) {
             String dbPluralForms = getPluralFormsFromDB(localeId);
-            if (StringUtils.isNotEmpty(dbPluralForms)) {
+            if (isNotEmpty(dbPluralForms)) {
                 return dbPluralForms;
             }
         }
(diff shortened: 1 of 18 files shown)


seanf:

seanf:
@carlosmunoz I could use a hand debugging the 404s which happen when the tests try to log start/stop (commit a6fdcca).

seanf:
> @carlosmunoz I could use a hand debugging the 404s which happen when the tests try to log start/stop (commit a6fdcca).

Never mind: I didn't realise I had to register each endpoint class in zanata-test-war's JAX-RS Application class.

zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java
```diff
@@ -1479,4 +1489,11 @@ public static void validateExtensions(Set<String> requestedExt) {
         }
     }
 
+    public static class PluralParseException extends RuntimeException {
```
aeng: Shouldn't this be a class itself?

zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java
```diff
@@ -108,6 +109,7 @@
     private static final String CONTENT_TYPE_HDR = HeaderFields.KEY_ContentType;
     private static final String PLURAL_FORMS_HDR = "Plural-Forms";
 
+    // TODO we should need only one regex, with a capture group for nplurals
```
aeng: Isn't this only 1 pattern?

seanf: There are three patterns in this file, but they didn't change in this PR, so they're not entirely visible here.

zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java
seanf: It is, it just lives inside the ResourceUtils namespace.

I could make it a top-level class, but I don't think it's needed.

aeng:
:+1:

seanf:

seanf:

seanf:

djansen-redhat:
@seanf uploading the files attached (with broken nplurals in the .po header) results in a plurals of 1 for Arabic, where I expected 6.  Could you clarify this for me?
[DJansenTest.zip](https://github.com/zanata/zanata-server/files/183621/DJansenTest.zip)

seanf:
@djansen-redhat I don't suppose you have the server log from that test?

On 22 March 2016 at 16:49, Damian Jansen <notifications@github.com> wrote:

> @seanf <https://github.com/seanf> uploading the files attached (with
> broken nplurals in the .po header) results in a plurals of 1 for Arabic,
> where I expected 6. Could you clarify this for me?
> DJansenTest.zip
> <https://github.com/zanata/zanata-server/files/183621/DJansenTest.zip>
>
> —
> <https://github.com/zanata/zanata-server/pull/958#issuecomment-199665662>
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

seanf:
@djansen-redhat the zanata.xml in that zip had a URL with `https` together with port `8080`. Did you really have Zanata running HTTPS on port 8080?

seanf:
@djansen-redhat I can't reproduce the problem - it uses nplural from the PO if available, otherwise the value specified in the HLocale (if any), otherwise the default. 

Is it possible you forgot that the first first text flow was singular, not plural?

djansen-redhat:
:white_check_mark: Tested

PR #965, opened 2015-09-14
fix: add rich: namespace explicitly

seanf:
https://zanata.atlassian.net/browse/ZNTA-674


seanf:
@carlosmunoz This workaround isn't really needed, as long as the Mojarra 2.1 modules are activated correctly in WildFly (https://zanata.atlassian.net/browse/ZNTA-674?focusedCommentId=13118&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13118).

However, do you think we should make the change anyway?  It wouldn't hurt to be a bit more compatible with Mojarra 2.2, even if we don't plan to use it yet.

carlosmunoz:
@seanf As long as it doesn't break anything, I don't see a problem with this change.
:+1:

djansen-redhat:
The failure for build 4402 can be ignored.

seanf:
Commit e348ace merged to integration/master.  I don't think we need this in release after all.

PR #980, opened 2015-09-24
ZNTA-650 migrate action in pages.xml to jsf pages

huangp:
Migrated commits from #959 

https://zanata.atlassian.net/browse/ZNTA-650

PR #959, opened 2015-09-04
ZNTA-650 pages xml migration

huangp:
https://zanata.atlassian.net/browse/ZNTA-659

PR #981, opened 2015-09-25
ZNTA-699 navigation flow migration

huangp:
https://zanata.atlassian.net/browse/ZNTA-699

Note that this is based on (and it targets) the branch in https://github.com/zanata/zanata-server/pull/980

### DIFF
--- a/zanata-war/src/main/java/org/zanata/action/LoginAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/LoginAction.java
@@ -20,18 +20,25 @@
  */
 package org.zanata.action;
 
+import java.io.IOException;
 import java.io.Serializable;
+
+import javax.faces.context.ExternalContext;
+import javax.faces.context.FacesContext;
 
 import lombok.Getter;
 import lombok.Setter;
+import lombok.extern.slf4j.Slf4j;
 
 import org.jboss.seam.ScopeType;
 import org.jboss.seam.annotations.In;
 import org.jboss.seam.annotations.Name;
 import org.jboss.seam.annotations.Scope;
+import org.jboss.seam.core.Conversation;
 import org.zanata.ApplicationConfiguration;
 import org.zanata.security.AuthenticationManager;
 import org.zanata.security.AuthenticationType;
+import org.zanata.security.UserRedirectBean;
 import org.zanata.security.ZanataCredentials;
 import org.zanata.security.ZanataIdentity;
 import org.zanata.security.openid.FedoraOpenIdProvider;
@@ -48,6 +55,7 @@ import org.zanata.security.openid.YahooOpenIdProvider;
  */
 @Name("loginAction")
 @Scope(ScopeType.PAGE)
+@Slf4j
 public class LoginAction implements Serializable {
     private static final long serialVersionUID = 1L;
 
@@ -74,6 +82,9 @@ public class LoginAction implements Serializable {
     @Getter
     @Setter
     private String openId = "http://";
+
+    @In
+    private UserRedirectBean userRedirect;
 
     public String login() {
         credentials.setUsername(username);
@@ -105,7 +116,33 @@ public class LoginAction implements Serializable {
                     "login() only supports internal, jaas, or kerberos authentication");
         }
 
+        if ("loggedIn".equals(loginResult)) {
+            if (authenticationManager.isAuthenticated() && authenticationManager.isNewUser()) {
+                return "createUser";
+            }
+            if (authenticationManager.isAuthenticated() && !authenticationManager.isNewUser() && userRedirect.shouldRedirectToDashboard()) {
+                return "dashboard";
+            }
+            if (authenticationManager.isAuthenticated() && !authenticationManager.isNewUser() && userRedirect.isRedirect()) {
+                // TODO [CDI] seam will create a conversation when you return view id directly or redirect to external url
+                return continueToPreviousUrl();
+            }
+        } else if ("inactive".equals(loginResult)) {
+            Conversation.instance().end();
+            return "inactive";
+        }
         return loginResult;
+    }
+
+    private String continueToPreviousUrl() {
+        ExternalContext ec = FacesContext.getCurrentInstance().getExternalContext();
+        try {
+            ec.redirect(userRedirect.getUrl());
+            return "continue";
+        } catch (IOException e) {
+            log.warn("error redirecting user to previous url: {}", userRedirect.getUrl(), e);
+            return "dashboard";
+        }
     }
 
     /**
--- a/zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
@@ -29,6 +29,7 @@ import org.zanata.model.HAccount;
 import org.zanata.model.HProject;
 import org.zanata.model.HProjectIteration;
 import org.zanata.ui.CopyAction;
+import org.zanata.util.FacesNavigationUtil;
 
 /**
  * Handles user interaction from merge_trans_modal.xhtml.
@@ -186,9 +187,7 @@ public class MergeTransAction extends CopyAction implements Serializable {
         mergeTranslationsManager.start(sourceProjectSlug,
             sourceVersionSlug, targetProjectSlug, targetVersionSlug,
             !keepExistingTranslation);
-        FacesContext context = FacesContext.getCurrentInstance();
-        NavigationHandler navigationHandler = context.getApplication().getNavigationHandler();
-        navigationHandler.handleNavigation(context, null, "merge-translation");
+        FacesNavigationUtil.handlePageNavigation(null, "merge-translation");
     }
 
     // Check if copy-trans, copy version or merge-trans is running for the
--- a/zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
@@ -5,6 +5,8 @@ import java.util.Collections;
 import java.util.List;
 
 import javax.faces.application.FacesMessage;
+import javax.faces.application.NavigationHandler;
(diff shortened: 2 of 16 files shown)


huangp:
Sean has reviewed it in #959 Need another JSF expert to review again here

zanata-war/src/main/java/org/zanata/action/LoginAction.java
```diff
@@ -105,9 +116,35 @@ public String login() {
...
+                return "dashboard";
+            }
+            if (authenticationManager.isAuthenticated() && !authenticationManager.isNewUser() && userRedirect.isRedirect()) {
+                // TODO [CDI] seam will create a conversation when you return view id directly or redirect to external url
+                return continueToPreviousUrl();
+            }
+        } else if ("inactive".equals(loginResult)) {
+            Conversation.instance().end();
+            return "inactive";
+        }
```
carlosmunoz: I imagine all this is what currently sits on pages.xml, it was simply moved to java.

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
```diff
@@ -184,6 +186,9 @@ public void startMergeTranslations() {
         mergeTranslationsManager.start(sourceProjectSlug,
             sourceVersionSlug, targetProjectSlug, targetVersionSlug,
             !keepExistingTranslation);
+        FacesContext context = FacesContext.getCurrentInstance();
+        NavigationHandler navigationHandler = context.getApplication().getNavigationHandler();
+        navigationHandler.handleNavigation(context, null, "merge-translation");
```
carlosmunoz: Why do we need this? Shouldn't this be going to the same page where the action is triggered by default?
Also, sounds like it's something that would commonly be used so maybe we should put it in some common class.

zanata-war/src/main/java/org/zanata/action/ProjectHome.java
```diff
@@ -286,8 +286,6 @@ public void setSelectedProjectType(String selectedProjectType) {
                 && !selectedProjectType.equals("null")) {
             ProjectType projectType = ProjectType.valueOf(selectedProjectType);
             getInstance().setDefaultProjectType(projectType);
-        } else {
-            getInstance().setDefaultProjectType(null);
```
carlosmunoz: This looks like it doesn't have much to do with navigation. Why did we remove these lines?

zanata-war/src/main/webapp/WEB-INF/faces-config.xml
```diff
@@ -78,4 +78,387 @@
...
+      <to-view-id>#{userRedirect.url}</to-view-id>
+      <redirect/>
+    </navigation-case>
+    <navigation-case>
+      <!-- Default redirection after a log in is to the dashboard -->
+      <if>#{identity.loggedIn}</if>
+      <to-view-id>/dashboard/home.xhtml</to-view-id>
+      <redirect/>
+    </navigation-case>
+  </navigation-rule>
```
carlosmunoz: On a previous comment I noted that one of the actions (`LoginAction`) had new Java code to determine navigation. Seems like this is the same thing but in `faces-config.xml`. Should we standardize on one place?

zanata-war/src/main/webapp/WEB-INF/faces-config.xml
```diff
@@ -78,4 +78,387 @@
...
+      &lt;!&ndash; TODO [CDI] end conversation (it will lose notification message) &ndash;&gt;
+      &lt;!&ndash;<end-conversation />&ndash;&gt;
+      <redirect/>
+    </navigation-case>
+  </navigation-rule>-->
+
+  <navigation-rule>
+    <from-view-id>/account/register.xhtml</from-view-id>
+    <navigation-case>
+      <from-outcome>true</from-outcome>
```
carlosmunoz: Should this be `success`, as is being used on several of the Actions?

zanata-war/src/main/webapp/WEB-INF/faces-config.xml
```diff
@@ -78,4 +78,387 @@
...
+      <to-view-id>/dashboard/home.xhtml</to-view-id>
+      <redirect/>
+    </navigation-case>
+  </navigation-rule>
+
+  <!-- Project -->
+  <navigation-rule>
+    <from-view-id>/project/project.xhtml</from-view-id>
+    <navigation-case>
+      <from-outcome>projectSlugUpdated</from-outcome>
```
carlosmunoz: Some other places are using dashes. What about `project-slug-updated`... or conversely switch everything else to camel case. I think this is the odd one out.

zanata-war/src/main/webapp/WEB-INF/faces-config.xml
```diff
@@ -78,4 +78,387 @@
...
+        </view-param>
+        <view-param>
+          <name>iterationSlug</name>
+          <value>#{versionHome.instance.slug}</value>
+        </view-param>
+      </redirect>
+    </navigation-case>
+    <navigation-case>
+      <from-action>#{versionHome.update}</from-action>
+      <from-outcome>versionSlugUpdated</from-outcome>
```
carlosmunoz: another camel case vs dashed scenario

zanata-war/src/main/webapp/WEB-INF/layout/version/merge_trans_modal.xhtml
```diff
@@ -158,7 +158,7 @@
               action="#{mergeTransAction.startMergeTranslations}"
               value="#{msgs['jsf.iteration.mergeTrans']}"/>
 
-            <ui:fragment rendered="#{mergeTransAction.isCopyActionsRunning(mergeTransAction.targetProjectSlug, mergeTransAction.targetVersionSlug)}">
+            <ui:fragment rendered="#{mergeTransAction.isCopyActionsRunning()}">
```
carlosmunoz: :+1: Good find, but this looks like it's fixing a different issue. Not sure if we should wait for CDI to also fix this.

zanata-war/src/main/webapp/WEB-INF/pages.xml
```diff
@@ -111,14 +14,7 @@
...
 
-  <page view-id="/account/register.xhtml">
-    <navigation from-action="#{identity.isLoggedIn()}">
-      <rule if-outcome="true">
-        <redirect view-id="/dashboard/home.xhtml" />
-      </rule>
-    </navigation>
-  </page>
-
+  <!-- TODO [CDI] uses conversation scope facesMessages. if migrated to faces-config, it will lose notification message -->
```
carlosmunoz: Do we already have a solution for when we are on CDI?

zanata-war/src/main/java/org/zanata/action/LoginAction.java
huangp: Yes. From memory faces-config has less functionality than pages.xml. e.g. it seems only work with string result not just from different EL.

zanata-war/src/main/java/org/zanata/action/MergeTransAction.java
huangp: This is in an AJAX context not a normal form action. I think this is the only place that requires this at the moment.

carlosmunoz: Ahh... so it won't apply the faces-config navigation rule from an ajax request to an action?

zanata-war/src/main/java/org/zanata/action/ProjectHome.java
huangp: When I test it after the change, upon opening up project home, one of the request (I think it's one of the ajax request but hard to verify) keep call this method and set the project type to null. This method is meant to be called only if user changes project type and there is no way they can select null. So I remove it.

zanata-war/src/main/webapp/WEB-INF/faces-config.xml
huangp: This is a result from ```#{identity.isLoggedIn}```

zanata-war/src/main/webapp/WEB-INF/faces-config.xml
huangp: It was using this value in pages.xml. I am just copying it over. I will make a new commit to change the name.

zanata-war/src/main/webapp/WEB-INF/pages.xml
huangp: JSF's flash scope in theory is designed to solve this problem (post -> redirect -> show the message). Sean thinks the deltaspike ViewAccessScope (as a replacement for conversation scope) may work.

carlosmunoz: I think flash scope is a better option. We could do this right now if we wanted to. Something like:
```java
FacesContext.getCurrentInstance().getExternalContext().getFlash()
```

zanata-war/src/main/webapp/WEB-INF/faces-config.xml
carlosmunoz: I think we should change it to a string. The navigation case is too general and makes it hard to understand what happens. I think in general (i.e. whenever possible) we should use the `<from-action>` parameter to be able to clearly understand how it's navigating.

zanata-war/src/main/webapp/WEB-INF/faces-config.xml
carlosmunoz: Ok cool

zanata-war/src/main/webapp/WEB-INF/faces-config.xml
huangp: Actually this case won't work. The action is invoked during preRenderView, I have to use explicit redirect to make it work. Now the navigation rule is removed

huangp:
Recreated in https://github.com/zanata/zanata-server/pull/1001 to target integration/master

PR #982, opened 2015-09-25
refactor - remove various Seam references

huangp:
- [x] TODO: rebase to remove these commits: dfc56c9 39251a1 c120818

### DIFF
--- a/zanata-model/src/main/java/org/zanata/model/HProject.java
+++ b/zanata-model/src/main/java/org/zanata/model/HProject.java
@@ -70,7 +70,6 @@ import org.hibernate.search.annotations.Analyzer;
 import org.hibernate.search.annotations.Field;
 import org.hibernate.search.annotations.Indexed;
 import org.hibernate.validator.constraints.NotEmpty;
-import org.jboss.seam.faces.FacesMessages;
 import org.zanata.annotation.EntityRestrict;
 import org.zanata.common.EntityStatus;
 import org.zanata.common.LocaleId;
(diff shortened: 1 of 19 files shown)


seanf:
Seam's `@Synchronized` doesn't just use Java's `synchronized

huangp:
We will have to develop CDI interceptor then. Have you looked at the code that uses it? Just because Seam offers that feature doesn't mean we indeed need it.

seanf:
This sort of timeout has come into play when openid servers are being slow, for one thing.  I think we do need it.  And not just in the places which currently say `@Synchronized` - I think it's implicit for Seam session-scoped components.

huangp:
It'd be an interceptor for all session scoped CDI beans. But take a look at the classes in this PR. They are all application scoped beans. Can you see any reason we should use timeout?

seanf:
Given that we're going to need a timeout feature anyway, I can't think of a reason why these shouldn't have a timeout.  Better an error message than a hung server.

huangp:
Depending on what we want to achieve and what we want to focus right now. 
The aim of this commit is: we could get a little bit closer to a compile-able code base once we remove Seam. 
The interceptor can be done as post CDI task. The more stuff we leave AFTER sed script migration, the longer that branch will live and harder for us to prevent merge conflicts. The quicker we can get a compile-able CDI version of Zanata, the better chance for us to develop against it. This includes aforementioned interceptor.

seanf:
I want to achieve a refactoring to reduce Seam dependencies, not a functionality change. That way, it's safe to merge the changes sooner.

There's nothing to stop us porting org.jboss.seam.core.SynchronizationInterceptor to CDI before the sed script migration (it could be tested in zanata-cdi-migration). We can probably even merge it into zanata-server ahead of time, since Weld is disabled and won't activate the new interceptor.

seanf:
@huangp I didn't mean to merge origin/master into this branch (commit c2e24b4).  When possible, I'd like to rebase this branch on top of master (which now includes java18), minus the merge.

Also, I think we should try not to do too much in each branch, because it makes it hard to review the changes.

seanf:
Looks good to me (when tests pass).

PR #983, opened 2015-09-25
Add any missing foreign keys

seanf:
In the past, some Zanata environments defaulted to MyISAM for new
tables, and so some tables were MyISAM when we tried to create various
foreign keys. (MySQL silently ignores ADD FOREIGN KEY for MyISAM
tables.) This commit ensures that the affected foreign keys are
created if they don't exist.

### DIFF
--- a/zanata-war/src/main/resources/db/changelogs/db.changelog-3.7.xml
+++ b/zanata-war/src/main/resources/db/changelogs/db.changelog-3.7.xml
@@ -184,6 +184,11 @@
     </createIndex>
   </changeSet>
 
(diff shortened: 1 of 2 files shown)


seanf:
This approach won't work on MySQL 5.x.  ADD FOREIGN KEY... IF NOT EXISTS was only added in MariaDB 10.

seanf:

carlosmunoz:
:+1:

djansen-redhat:
:white_check_mark: Tested, all fks appear to have been given.

PR #984, opened 2015-09-26
fix(ZNTA-634): increase session timeout

carlosmunoz:
For authenticated users. Value is determined by components.xml
Session timeout gets increased after logging in, and reverts back after logging out.

https://zanata.atlassian.net/browse/ZNTA-634

### DIFF
--- a/zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java
+++ b/zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java
@@ -423,12 +423,11 @@ public class ApplicationConfiguration implements Serializable {
                 .getInstance()
                 .getRequest();
         if (request != null) {
+            int timeoutInSecs = max(authenticatedSessionTimeoutMinutes * 60,
+                    defaultAnonymousSessionTimeoutMinutes * 60);
             request
-                    .getSession()
-                    .setMaxInactiveInterval(
-                            max(authenticatedSessionTimeoutMinutes * 60,
-                                    defaultAnonymousSessionTimeoutMinutes *
-                                    60));
+                .getSession()
+                .setMaxInactiveInterval(timeoutInSecs);
         }
     }
 
@@ -439,9 +438,9 @@ public class ApplicationConfiguration implements Serializable {
                 .getRequest();
         if (request != null) {
             request
-                    .getSession()
-                    .setMaxInactiveInterval(
-                            defaultAnonymousSessionTimeoutMinutes * 60);
+                .getSession()
+                .setMaxInactiveInterval(
+                        defaultAnonymousSessionTimeoutMinutes * 60);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


huangp:
:+1:

carlosmunoz:
@seanf Implemented changes.

zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java
```diff
@@ -408,4 +415,33 @@ public String copyrightNotice() {
...
+            @Observes PostAuthenticateEvent payload) {
+        HttpServletRequest request = ServletContexts
+                .getInstance()
+                .getRequest();
+        if (request != null) {
+            request
+                    .getSession()
+                    .setMaxInactiveInterval(
+                            max(authenticatedSessionTimeoutMinutes * 60,
+                                    defaultAnonymousSessionTimeoutMinutes *
```
seanf: I think this would be a little easier to read if you defined a local variable (eg timeout) and calculated it a few lines up.  (Or if the setter methods adjusted authenticatedSessionTimeoutMinutes when crazy values come in.)

zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java
```diff
@@ -408,4 +415,32 @@ public String copyrightNotice() {
...
     }
+
+    @Observer(PostAuthenticateEvent.EVENT_NAME)
+    public void setAuthenticatedSessionTimeout(
+            @Observes PostAuthenticateEvent payload) {
+        HttpServletRequest request = ServletContexts
+                .getInstance()
+                .getRequest();
+        if (request != null) {
+            int timeoutInMins = max(authenticatedSessionTimeoutMinutes * 60,
```
seanf: It looks like timeoutInSeconds to me.

zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java
```diff
@@ -76,6 +80,9 @@
     @Getter
     private static final int defaultMaxFilesPerUpload = 100;
 
+    @Getter
+    private static final int defaultAnonymousSessionTimeoutMinutes = 30;
```
djansen-redhat: Not to be picky (surely not!) but why minutes here and calculated to seconds later?

carlosmunoz: 1. Because there was an existing parameter that measures timeouts in minutes.
2. The Request API requires seconds.
3. If we ever decide to make this a user provided value, makes more sense to define a session timeout in minutes I guess.

djansen-redhat:
:white_check_mark: I'm _reasonably_ sure this works as intended. Time based things are a pain to test.

carlosmunoz:
Merged in 5313a5f8800f6472f0408d5a16d5d4a3030d8ba3

PR #985, opened 2015-09-27
fix(ZNTA-639): handle ViewExpiredException

carlosmunoz:
Handle these type of exceptions both in ajax and non-ajax requests.
For ajax requests, inform the user and ask them to reload.
For non-ajax requests, inform the user and offer to head back to the
original page.

https://zanata.atlassian.net/browse/ZNTA-639

### DIFF
--- a/zanata-war/src/main/java/org/zanata/ZanataInit.java
+++ b/zanata-war/src/main/java/org/zanata/ZanataInit.java
@@ -66,7 +66,6 @@ import org.jboss.seam.annotations.Observer;
 import org.jboss.seam.annotations.Scope;
 import org.jboss.seam.contexts.ServletLifecycle;
 import org.jboss.seam.mail.MailSession;
-import org.jboss.seam.transaction.FacesTransactionEvents;
 import org.zanata.events.ServerStarted;
 import org.zanata.exception.ZanataInitializationException;
 import org.zanata.rest.dto.VersionInfo;
@@ -113,14 +112,8 @@ public class ZanataInit {
     @In
     private EntityManagerFactory entityManagerFactory;
 
-    @In
-    private FacesTransactionEvents facesTransactionEvents;
-
     @Observer("org.jboss.seam.postInitialization")
     public void initZanata() throws Exception {
-        // disable the automatic JSF message when a transaction fails
-        // TODO [CDI] Will not be needed after CDI migration
-        facesTransactionEvents.setTransactionFailedMessageEnabled(false);
         checkAppServerVersion();
         ServletContext servletContext =
                 ServletLifecycle.getCurrentServletContext();
--- a/zanata-war/src/main/java/org/zanata/ZanataInit.java
+++ b/zanata-war/src/main/java/org/zanata/ZanataInit.java
@@ -66,6 +66,7 @@ import org.jboss.seam.annotations.Observer;
 import org.jboss.seam.annotations.Scope;
 import org.jboss.seam.contexts.ServletLifecycle;
 import org.jboss.seam.mail.MailSession;
+import org.jboss.seam.transaction.FacesTransactionEvents;
 import org.zanata.events.ServerStarted;
 import org.zanata.exception.ZanataInitializationException;
 import org.zanata.rest.dto.VersionInfo;
@@ -112,8 +113,13 @@ public class ZanataInit {
     @In
     private EntityManagerFactory entityManagerFactory;
 
+    @In
+    private FacesTransactionEvents facesTransactionEvents;
+
     @Observer("org.jboss.seam.postInitialization")
     public void initZanata() throws Exception {
+        // disable the automatic JSF message when a transaction fails
+        facesTransactionEvents.setTransactionFailedMessageEnabled(false);
         checkAppServerVersion();
         ServletContext servletContext =
                 ServletLifecycle.getCurrentServletContext();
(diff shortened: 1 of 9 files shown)


carlosmunoz:
@huangp Sorry, I think @seanf grabbed the pull request originally assigned to you. Can you take a look at this one instead?

zanata-war/src/main/java/org/zanata/ZanataInit.java
```diff
@@ -112,8 +113,13 @@
     @In
     private EntityManagerFactory entityManagerFactory;
 
+    @In
+    private FacesTransactionEvents facesTransactionEvents;
+
     @Observer("org.jboss.seam.postInitialization")
     public void initZanata() throws Exception {
+        // disable the automatic JSF message when a transaction fails
+        facesTransactionEvents.setTransactionFailedMessageEnabled(false);
```
huangp: Please add a TODO [CDI] tag on this line. It's not needed when migrated to CDI.

huangp: Not sure whether this will work too (add this in component.xml)
```  <component class="org.jboss.seam.transaction.FacesTransactionEvents" installed="false" />```
Trying to reduce the seam stuff in our code base...

huangp:
:+1: other than the comment

zanata-war/src/main/java/org/zanata/ZanataInit.java
carlosmunoz: I think it would :+1: 
Let me approach it that way.

djansen-redhat:
This appears to work as intended. A query though - should the notification be the "error" one, i.e. the one that stays on the page until it's dismissed?

djansen-redhat:
There's also a bit of an odd issue where if the user clicks back, it takes them to the Log In page. Logging in then shows them the same Log In page.

definite:

carlosmunoz:
> This appears to work as intended. A query though - should the notification be the "error" one, i.e. the one that stays on the page until it's dismissed?

My thought is that nothing is "broken", so no need to alarm the user with a red alert. But happy to change it if it makes more sense otherwise.

seanf:
Well, isn't the user's session broken, taking some of the user's state
along with it? It seems like a mild warning might prevent some user
confusion. Those disappearing messages are easy to miss...

On 29 October 2015 at 15:01, Carlos Munoz <notifications@github.com> wrote:

> This appears to work as intended. A query though - should the notification
> be the "error" one, i.e. the one that stays on the page until it's
> dismissed?
>
> My thought is that nothing is "broken", so no need to alarm the user with
> a red alert. But happy to change it if it makes more sense otherwise.
>
> —
> <https://github.com/zanata/zanata-server/pull/985#issuecomment-152076303>.
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

carlosmunoz:
@djansen-redhat 
> There's also a bit of an odd issue where if the user clicks back, it takes them to the Log In page. Logging in then shows them the same Log In page.

Can't reproduce this. How are you producing this situation?

djansen-redhat:
@carlosmunoz I'm not sure now that it's a problem with this pull request.  It seems to happen intermittently with many areas, e.g. clicking "Glossary" after expired view.

carlosmunoz:
@djansen-redhat Ok, so just want to summarize both yours and @seanf 's review comments:
 - [ ] Change the error alert and page colors to 'alert' (Red) @seanf Can you confirm this is the way you want to go? Is it only for the ajax alert? or for the error page as well?

I will ignore the issue you reported as it seems to be unrelated to this pull request per your comment.

djansen-redhat:
_Personally_ I think the blue page is fine, it doesn't alarm the user. But yes, the notification should be persistent until dismissed.  Like me!

seanf:
Agreed. It doesn't need to look too scary, but (like any warning) it
probably shouldn't disappear on a timer.

On 2 November 2015 at 11:13, Damian Jansen <notifications@github.com> wrote:

> *Personally* I think the blue page is fine, it doesn't alarm the user.
> But yes, the notification should be persistent until dismissed. Like me!
>
> —
> <https://github.com/zanata/zanata-server/pull/985#issuecomment-152885386>.
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

carlosmunoz:
@djansen-redhat @seanf I did a bit of research and it seems there is no way to currently have a permanent alert. Wondering if we could split that up into a different pull request.

djansen-redhat:
@carlosmunoz I think that's acceptable.

djansen-redhat:
Oops.
Verified (nothing has changed since issue mentioned).

carlosmunoz:
@seanf Your thoughts?

seanf:
That sounds fine, as long as we don't forget about the "split-out" part.

Could we at least increase the timeout?  Even for info messages, I think
they were a bit too brief, last time I saw one.

On 2 November 2015 at 15:58, Carlos Munoz <notifications@github.com> wrote:

> @seanf <https://github.com/seanf> Your thoughts?
>
> —
> <https://github.com/zanata/zanata-server/pull/985#issuecomment-152921920>.
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

carlosmunoz:
Merged in 4f26d4012e6823d8855e6e6de92a436c59cbe461

There is a requested change in zanata-assets to increase the timeout period of the notifications, but merging this change since it's on a different repo.

PR #989, opened 2015-09-29
Increase password length to 1024

aeng:
https://zanata.atlassian.net/browse/ZNTA-624


huangp:
:+1:

definite:
The max length is set exact at 1024, but a user does input 1025, the password will be truncated, which is hard for a normal user to spot it.

definite:
Test case:
1. Type 'a' in password field.
2. Click outside the field

Expected:
'a' is kept in password field

Actual:
password field is wiped out.

definite:
:+1:

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/52c9c40ed863c9b4ca36728ffea0be159c8e3f91

PR #991, opened 2015-09-30
Add CDI interceptor for Synchronized

seanf:


huangp:
:+1:

seanf:
@djansen-redhat to minimise conflicts, I suggest we merge this now, since it didn't break the tests, and the interceptor will only be active if (a) weld is enabled in standalone.xml (b) we include beans.xml in the war file and (c) we start using the new annotation.  

The only place we use the new annotation is https://github.com/zanata/zanata-server/pull/886 (seam-to-cdi).

djansen-redhat:
I think I can rely on the tests for this one.

PR #997, opened 2015-10-02
refactor: Remove usages of Seam's Work class

carlosmunoz:
Turned the seam class into a functional utility.
This revision still performs the same code as the Seam Work class.


carlosmunoz:
@huangp Care to start taking a look?
For the code inside the new `TransactionUtil.runInTransaction` method, are we leaving that for after CDI migration? If not, any pointers on what code in there I can start changing now (specially around getting a hold of a transaction)?

seanf:
@carlosmunoz Two functional test failures.
@zanata-jenkins test this please.

carlosmunoz:
@huangp functional tests passing now. Could you please review this?
I didn't tackle the replacement for transaction objects. I plan on doing that separately.

zanata-war/src/main/java/org/zanata/service/impl/CopyTransWorkFactory.java
```diff
@@ -57,13 +82,374 @@
     @In
     private VersionStateCache versionStateCacheImpl;
 
-    public CopyTransWork createCopyTransWork(HLocale targetLocale,
+    public Integer runCopyTransInNewTx(HLocale targetLocale,
+            HCopyTransOptions options, HDocument document,
+            boolean requireTranslationReview, List<HTextFlow> copyTargets)
+            throws Exception {
+        return runInTransaction(() -> runCopyTrans(targetLocale, options,
```
huangp: So this is a Runnable/Callable calling an instance method of the outer class? I am not quite sure how the injection work in such condition. The Runnable instance is bound to a stateless scoped bean instance? We also don't make use of the container thread pool, do we? Looks like a ```@Async``` may be better?

huangp: Was reading this http://stackoverflow.com/questions/21284191/cdi-multithreading

huangp: And this https://wmarkito.wordpress.com/2012/07/20/cdi-events-synchronous-x-asynchronous/

carlosmunoz: So, this is not about multithreading (or async). Fortunately all the code runs synchronously and sequentially here. It's just that it needs to start a new transaction around the method. The injection won't be a problem because it's running in the same context all the time (as opposed to async or multi-thread). The Runnable can see everything that the method can, and there is no problem with accessing the injected beans because the bean's invoked method (the one calling the runnable) finishes after the runnable.

The posts you reference are useful when we implement Async. For that, we do need a thread executor (already there) as well as manually starting the CDI contexts (already in the demo project). That is a much harder problem to solve.

zanata-war/src/main/java/org/zanata/transaction/TransactionUtil.java
```diff
@@ -0,0 +1,162 @@
...
+            return null;
+        });
+    }
+
+    protected static boolean
+            isNewTransactionRequired(boolean transactionActive) {
+        return !transactionActive;
+    }
+
+    public static boolean isRollbackRequired(Exception e, boolean isJavaBean) {
```
huangp: Looks like we only call this method with isJavaBean as true. May clean this up to only accept e as parameter

zanata-war/src/main/java/org/zanata/transaction/TransactionUtil.java
```diff
@@ -0,0 +1,162 @@
...
+                (clazz.isAnnotationPresent(ApplicationException.class) && clazz
+                        .getAnnotation(ApplicationException.class).rollback());
+    }
+
+    private static boolean isSystemException(Exception e, boolean isJavaBean,
+            Class<? extends Exception> clazz) {
+        return isJavaBean &&
+                (e instanceof RuntimeException) &&
+                !clazz.isAnnotationPresent(APPLICATION_EXCEPTION) &&
+                !clazz.isAnnotationPresent(ApplicationException.class) &&
```
huangp: Last time I checked, we don't have a case to use ApplicationException or APPLICATION_EXCEPTION, so this two conditions can be removed.

zanata-war/src/main/java/org/zanata/transaction/TransactionUtil.java
```diff
@@ -0,0 +1,162 @@
...
+                // TODO: this is hackish, maybe just turn off RollackInterceptor
+                // for @Converter/@Validator components
+                !JSF.VALIDATOR_EXCEPTION.isInstance(e) &&
+                !JSF.CONVERTER_EXCEPTION.isInstance(e);
```
huangp: Will need to in source these two utility methods from seam

huangp:
:+1:

definite:

definite:
:+1:

PR #999, opened 2015-10-05
Use Nashorn for CommonMark rendering

seanf:


zanata-war/src/main/java/org/zanata/util/CommonMarkRenderer.java
```diff
@@ -21,20 +21,22 @@
 package org.zanata.util;
 
 import com.google.common.base.Throwables;
+import com.google.common.io.Resources;
+import jdk.nashorn.api.scripting.JSObject;
```
carlosmunoz: @seanf Is this an OpenJDK-specific API? If not, is there a fallback?

seanf: Nashorn is part of both Oracle JDK and OpenJDK 1.8.  It's unfortunate that we have to depend on a Nashorn interface here, but there seems to be no way around that if we want the best performance.

seanf: Nashorn is also in Azul Zulu 8 and IBM's JDK 8.

carlosmunoz:
:+1:

seanf:
For testing, please note that we currently use server-side javascript:

* when rendering markdown on the home page
* when rendering markdown on a project About page

djansen-redhat:
:white_check_mark: This looks fine to me

PR #1000, opened 2015-10-06
chore: move zanatasearchindex under target dir in tests

seanf:
... instead of creating `zanatasearchindex` in current directory, which prevents `mvn clean` from doing its job (eg when Lucene lock files are lying around)


carlosmunoz:
:+1:

PR #1001, opened 2015-10-06
ZNTA-699 navigation flow migration

huangp:
https://zanata.atlassian.net/browse/ZNTA-699

Note the code is already reviewed in https://github.com/zanata/zanata-server/pull/981


carlosmunoz:
@huangp Is this still valid since we have already removed Seam from the CDI branch?

huangp:
Yes this is still valid. We merged this branch into seam-to-cdi. We could merge this into master earlier if we want to. But I think I've encounter issues when after signin the redirection sometimes will trigger exception.

seanf:
@carlosmunoz @huangp Since this code is part of the CDI branch, I suggest we close this PR.

PR #1004, opened 2015-10-09
test: bind EAP to localhost only

seanf:
See also https://github.com/zanata/zanata-server/pull/1003

### DIFF
--- a/functional-test/src/test/resources/conf/standalone.xml
+++ b/functional-test/src/test/resources/conf/standalone.xml
@@ -446,10 +446,10 @@
   </profile>
   <interfaces>
     <interface name="management">
-      <inet-address value="${jboss.bind.address.management:0.0.0.0}" />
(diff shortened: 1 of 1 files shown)


carlosmunoz:
This is what should be done for any Jboss server. so :+1: 
Just curious... what are we seeing here? and why is it just raising a red flag?

seanf:
Why just now, you mean?  

After https://github.com/zanata/zanata-server/pull/1003 I went looking through the build logs to see if we'd had any other instances of "BindException: Address already in use" besides SMTP_PORT, and found one with the JBOSS_HTTP_PORT. Then I realised it was the same problem as with SMTP, but in the EAP configuration.

With both fixes in, I'm hoping we won't see "BindException: Address already in use" in our builds unless we fail to use allocate-jboss-ports, or if standalone.xml fails to use the allocated ports (environment vars).

carlosmunoz:
:+1:

PR #1005, opened 2015-10-12
Add more checks for bad/stale TM results

seanf:


carlosmunoz:
I only have a question about the result of using the `PassThroughTokenizer` when the analyzed field is not a lower case one. What's the intended effect there?

seanf:
Well, the /intention/ is to parse the query (or index the field) without
change - don't split into tokens; don't change the case.

Do you think the intention might be wrong, or the implementation?

On 13 October 2015 at 09:30, Carlos Munoz <notifications@github.com> wrote:

> I only have a question about the result of using the PassThroughTokenizer
> when the analyzed field is not a lower case one. What's the intended effect
> there?
>
> —
> <https://github.com/zanata/zanata-server/pull/1005#issuecomment-147548504>
> .
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

carlosmunoz:
So, in the case of a non-lower case field we don't apply the `StandardFilter`?

djansen-redhat:
The failure at build 4688 (setRegisterURLTest) can be ignored

djansen-redhat:
:white_check_mark: Reasonably sure this doesn't break anything

PR #1006, opened 2015-10-13
fix(ZNTA-721): pre-sort TM by Lucene score

seanf:
Previously, our custom sort caused newer matches always to be favoured over highly-scored matches. Now we sort first by score, then by date.

See https://zanata.atlassian.net/browse/ZNTA-721

### DIFF
--- a/zanata-model/src/main/java/org/zanata/hibernate/search/DateBridge.java
+++ b/zanata-model/src/main/java/org/zanata/hibernate/search/DateBridge.java
@@ -28,6 +28,7 @@ import org.hibernate.search.bridge.TwoWayStringBridge;
 
 public class DateBridge implements TwoWayStringBridge {
 
+    // TODO include milliseconds for more precision (".SSS")
     private final static String DATE_FORMAT = "yyyy-MM-dd hh:mm:ss";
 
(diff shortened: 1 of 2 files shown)


djansen-redhat:
The failure in build 4695 (unsetTermsOfUseURL) can be ignored

carlosmunoz:
:+1:

definite:
:+1: 
Note that after deploy and before reindex, the automatic search might still mix up.
However, after reindex, the automatic search result behave as expected.

seanf:
Thanks. Note that this code change does not affect indexing.  A re-index
should only be needed if the index is out of date for other reasons (eg
database restored from backup, or a Zanata upgrade where indexing was
changed).

On 13 October 2015 at 16:17, Ding-Yi Chen <notifications@github.com> wrote:

> [image: :+1:]
> Note that after deploy and before reindex, the automatic search might
> still mix up.
> However, after reindex, the automatic search result behave as expected.
>
> —
> <https://github.com/zanata/zanata-server/pull/1006#issuecomment-147616426>
> .
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

PR #1012, opened 2015-10-17
Internal "user request management" and tracking

aeng:
https://zanata.atlassian.net/browse/ZNTA-555


zanata-model/src/main/java/org/zanata/model/LanguageRequest.java
```diff
@@ -0,0 +1,53 @@
...
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.OneToOne;
+import java.io.Serializable;
+
+/**
+ * @author Alex Eng <a href="aeng@redhat.com">aeng@redhat.com</a>
```
seanf: Could you please add a description?

zanata-model/src/main/java/org/zanata/model/LanguageRequest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @ManyToOne
+    @JoinColumn(name = "localeId", nullable = false)
+    private HLocale locale;
+
+    private Boolean coordinator;
+
+    private Boolean reviewer;
+
+    private Boolean translator;
```
seanf: The fields in the class could use javadocs.  Any chance we could use native `boolean`s here?

zanata-model/src/main/java/org/zanata/model/Request.java
```diff
@@ -0,0 +1,126 @@
...
+import javax.persistence.EntityListeners;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.PrePersist;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+import java.util.Date;
+
+/**
+ * @author Alex Eng <a href="aeng@redhat.com">aeng@redhat.com</a>
```
seanf: This could use a javadoc too.

zanata-model/src/main/java/org/zanata/model/Request.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * Return new request with updated state.
+     * Expire this request (set validTo)
+     *
+     * @param actor
+     * @param state
+     * @param comment
+     */
+    public Request update(HAccount actor, RequestState state, String comment) {
```
seanf: This is a bit like a `with` method (because it returns a clone with different field values).  Except for the important fact that it also expires the old record.

`update` is probably fine, but we could also consider `updateWith`, `replaceWith`, `expireWithReplace`...

zanata-model/src/main/java/org/zanata/model/TimeEntityBase.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.persistence.Id;
+import javax.persistence.MappedSuperclass;
+import javax.persistence.Temporal;
+import javax.persistence.TemporalType;
+import javax.validation.constraints.NotNull;
+import java.io.Serializable;
+import java.util.Date;
+
+/**
+ * @author Alex Eng <a href="aeng@redhat.com">aeng@redhat.com</a>
```
seanf: This one is definitely going to need a javadoc.  And does the Request's EntityListener belong here?

zanata-model/src/main/java/org/zanata/model/type/RequestStateType.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    public String objectToSQLString(RequestState value, Dialect dialect)
+            throws Exception {
+        return "\'" + toString(value) + "\'";
+    }
+
+    public RequestState stringToObject(String xml) throws Exception {
+        if (xml.length() != 1) {
+            throw new MappingException(
+                    "multiple or zero characters found parsing string");
```
seanf: If you're going to that much trouble, you might as well include the string xml in the exception message.

zanata-model/src/main/java/org/zanata/model/type/RequestStateType.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+        return RequestState.valueOf(xml.charAt(0));
+    }
+
+    public RequestState fromStringValue(String xml) {
+        return RequestState.valueOf(xml.charAt(0));
```
seanf: And then this method just assumes that charAt(0) is all we care about.  Should we check the length here too? (Maybe just an assert?)

zanata-model/src/main/java/org/zanata/model/type/RequestStateTypeDescriptor.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    RequestState valueOf(String string) {
+        if (string.length() == 1) {
+            return RequestState.valueOf(string.charAt(0));
+        }
+        throw new IllegalArgumentException(
+            "String to look up a ProjectRole must be exactly 1 character. " +
+            "Received \"" + string + "\"");
```
seanf: The formatter should have double-indented this.

zanata-model/src/main/java/org/zanata/model/type/RequestStateTypeDescriptor.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    char getIdentifyingChar(RequestState value) {
+        return value.getInitial();
+    }
+
+    @Override
+    RequestState valueOf(String string) {
+        if (string.length() == 1) {
+            return RequestState.valueOf(string.charAt(0));
```
seanf: How many different places is "convert a char to RequestState" implemented?

zanata-model/src/main/java/org/zanata/model/type/RequestTypeType.java
```diff
@@ -0,0 +1,68 @@
...
+        throws Exception {
+        return "\'" + toString(value) + "\'";
+    }
+
+    public RequestType stringToObject(String xml) throws Exception {
+        if (xml.length() < 1) {
+            throw new MappingException(
+                "multiple or zero characters found parsing string");
+        }
+        return RequestType.getValueOf(xml);
```
seanf: *Another* converter?

zanata-war/src/main/java/org/zanata/action/LanguageAction.java
```diff
@@ -128,6 +134,13 @@
 
     private List<SelectablePerson> searchResults;
 
+    private Map<LocaleRole, Boolean> roles = new HashMap<LocaleRole, Boolean>();
+    {
```
seanf: I don't mind instance initialisers, but we should have a blank line to separate it from the Map declaration. This way looks a bit misleading.

zanata-war/src/main/java/org/zanata/action/LanguageJoinAction.java
```diff
@@ -97,65 +92,164 @@
...
-    public void bindRole(String role, boolean checked) {
-        if (role.equals("translator")) {
-            requestAsTranslator = checked;
-        } else if (role.equals("reviewer")) {
-            requestAsReviewer = checked;
-        } else if (role.equals("coordinator")) {
-            requestAsCoordinator = checked;
-        }
+    public String getRequestedRole(LanguageRequest request) {
+        return Joiner.on(", ")
```
seanf: Now that we're on Java 8, you might be able to use String.join if you prefer.

zanata-war/src/main/java/org/zanata/action/LanguageJoinAction.java
```diff
@@ -97,65 +92,164 @@
...
+
+    @In
+    private LanguageTeamService languageTeamServiceImpl;
+
     @In(value = ZanataJpaIdentityStore.AUTHENTICATED_USER, required = false)
     private HAccount authenticatedAccount;
 
-    public boolean hasSelectedRole() {
-        return requestAsTranslator || requestAsReviewer || requestAsCoordinator;
+    public String getMyRequestedRole() {
```
seanf: What are we doing with this string? It seems a bit non-typesafe. 

Even if it is for immediate display to the user, we should consider internationalisation of role names (eg the `Messages` class).

zanata-war/src/main/java/org/zanata/action/LanguageJoinAction.java
```diff
@@ -97,65 +92,164 @@
...
+            boolean isRequestAsCoordinator) {
+        try {
+            requestServiceImpl
+                    .createLanguageRequest(authenticatedAccount, getLocale(),
+                            isRequestAsCoordinator,
+                            isRequestAsReviewer, isRequestAsTranslator);
+            sendRequestEmail(isRequestAsCoordinator, isRequestAsReviewer,
+                isRequestAsTranslator);
+        } catch (RequestExistException e) {
+            log.warn("Request already exist for {0} in language {1}.",
```
seanf: "exist" should be "exists".  But shouldn't we tell the user about this problem, instead of logging a warning?

zanata-war/src/main/java/org/zanata/action/LanguageJoinAction.java
```diff
@@ -97,65 +92,164 @@
...
+                "Request cancelled by requester {"
+                        + authenticatedAccount.getUsername() + "}";
+        requestServiceImpl.updateLanguageRequest(languageRequest.getId(),
+                authenticatedAccount, RequestState.CANCELLED, comment);
+
+        facesMessages.addGlobal(msgs.format("jsf.language.request.cancelled",
+            authenticatedAccount.getUsername()));
+    }
+
+    public String getMyRoles() {
```
seanf: See https://github.com/zanata/zanata-server/pull/1012/files#r42337956

zanata-war/src/main/java/org/zanata/dao/LanguageRequestDAO.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public LanguageRequest findRequesterPendingRequests(HAccount requester,
+        LocaleId localeId) {
+        String query =
+            "from LanguageRequest req where req.locale.localeId = :localeId and req.request.requester.id = :requesterId and req.request.validTo is null";
+        Query q = getSession().createQuery(query)
+            .setParameter("requesterId", requester.getId())
+            .setParameter("localeId", localeId)
+            .setCacheable(true).setComment(
+                "requestDAO.findRequestInLocale");
```
seanf: wrong comment

zanata-war/src/main/java/org/zanata/dao/LanguageRequestDAO.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public LanguageRequestDAO(Session session) {
+        super(LanguageRequest.class, session);
+    }
+
+    public LanguageRequest findRequesterPendingRequests(HAccount requester,
+        LocaleId localeId) {
+        String query =
+            "from LanguageRequest req where req.locale.localeId = :localeId and req.request.requester.id = :requesterId and req.request.validTo is null";
```
seanf: Please split up the string so that it can be line-wrapped before 80 chars.

zanata-war/src/main/java/org/zanata/dao/LanguageRequestDAO.java
```diff
@@ -0,0 +1,53 @@
...
+            .setParameter("localeId", localeId)
+            .setCacheable(true).setComment(
+                "requestDAO.findRequestInLocale");
+        return (LanguageRequest) q.uniqueResult();
+    }
+
+    public List<LanguageRequest> findPendingRequests(
+        List<LocaleId> localeIds) {
+        String query =
+            "from LanguageRequest req where req.request.validTo is null and req.locale.localeId in (:localeIds)";
```
seanf: This should be split too, thanks.

zanata-war/src/main/java/org/zanata/dao/RequestDAO.java
```diff
@@ -0,0 +1,53 @@
...
+        String query =
+            "from Request req where req.entityId= :entityId order by req.validFrom";
+        Query q = getSession().createQuery(query)
+            .setParameter("entityId", entityId)
+            .setCacheable(true).setComment(
+                "requestDAO.getByEntityId");
+        return q.list();
+    }
+
+    public Request getPendingRequestByEntityId(String entityId) {
```
seanf: I think the most common case for these time-based entities would be "get me the latest version of entity", so I think it would be better to name this one "getByEntityId", and name the one which returns a list "getHistoryByEntityId".

Also, perhaps these methods could be declared in an AbstractTimeBasedDAO to help ensure consistency - even if all the methods are abstract for now.

zanata-war/src/main/java/org/zanata/email/RequestToJoinLanguageEmailStrategy.java
```diff
@@ -68,14 +66,12 @@ public String getSubject(Messages msgs) {
...
         PersistentMap<String, Object> context = super.makeContext(
                 genericContext, toAddresses);
-        String safeHTML = HtmlUtil.SANITIZER.sanitize(htmlMessage);
         return context
                 .insert("fromLoginName", fromLoginName)
                 .insert("fromName", fromName)
                 .insert("replyEmail", replyEmail)
                 .insert("localeId", localeId)
                 .insert("localeNativeName", localeNativeName)
-                .insert("htmlMessage", safeHTML)
```
seanf: Are we not going to send an email for these requests, any more?

zanata-war/src/main/java/org/zanata/events/RequestUpdatedEvent.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Id for request subclass. e.g {@link org.zanata.model.LanguageRequest}
+     */
+    Long requestId;
+
+    /**
+     * Id for actor for this request. HAccount id
+     */
+    Long actorId;
```
seanf: Perhaps we should use native `long`s for these event fields, or Optional<Long> if they are optional (to be explicit).

zanata-war/src/main/java/org/zanata/exception/RequestExistException.java
```diff
@@ -0,0 +1,11 @@
+package org.zanata.exception;
+
+/**
+ * @author Alex Eng <a href="aeng@redhat.com">aeng@redhat.com</a>
+ */
+public class RequestExistException extends ZanataServiceException {
```
seanf: I can't tell what this is (javadoc please), but RequestExistsException might be better.

zanata-war/src/main/java/org/zanata/service/RequestService.java
```diff
@@ -0,0 +1,105 @@
...
+            HLocale locale, boolean isRequestAsCoordinator,
+            boolean isRequestAsReviewer,
+            boolean isRequestAsTranslator) throws RequestExistException;
+
+    /**
+     * Check if request already exist in the locale by this user
+     * @param requester
+     * @param locale
+     */
+    boolean isLanguageRequestExist(HAccount requester, HLocale locale);
```
seanf: Please change is -> does.

zanata-war/src/main/java/org/zanata/service/RequestService.java
```diff
@@ -0,0 +1,105 @@
...
+     * @throws EntityNotFoundException
+     */
+    void updateLanguageRequest(Long requestId, HAccount actor,
+        RequestState state, String comment) throws EntityNotFoundException;
+
+    /**
+     * Get language request by id
+     * @param id
+     */
+    LanguageRequest getLanguageRequest(Long id);
```
seanf: Use `long` if you can.  And please change `id` to `requestId` or whatever is appropriate.

zanata-war/src/main/java/org/zanata/service/RequestService.java
```diff
@@ -0,0 +1,105 @@
...
+     * Return pending request by entityId
+     * @param entityId
+     */
+    Request getPendingRequestByEntityId(String entityId);
+
+    /**
+     * Return request including history by entityId
+     * @param entityId
+     */
+    List<Request> getRequestByEntityId(String entityId);
```
seanf: perhaps getRequestHistoryByEntityId?

zanata-war/src/main/java/org/zanata/service/impl/RequestServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+        LanguageRequest languageRequest =
+            languageRequestDAO.findRequesterPendingRequests(requester,
+                locale.getLocaleId());
+
+        if (languageRequest != null) {
+            throw new RequestExistException(
+                RequestType.LOCALE + ", " + requester.getUsername() + ", " +
+                    ", " + locale.getDisplayName());
+        }
+        String entityId = UUID.randomUUID().toString();
```
seanf: Please extract this to a method, eg `requestDAO.generateEntityId`, in case we want to change the way we generate these IDs.

zanata-war/src/main/java/org/zanata/service/impl/RequestServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+        Request request = new Request(RequestType.LOCALE, requester, entityId);
+
+        languageRequest =
+            new LanguageRequest(request, locale, isRequestAsCoordinator,
+                isRequestAsReviewer, isRequestAsTranslator);
```
seanf: It would be good to give LanguageRequest a builder, because I can see it being *extremely* easy to mix up the order of the booleans isRequestAsCoordinator, isRequestAsReviewer and isRequestAsTranslator.

zanata-war/src/main/java/org/zanata/service/impl/RequestServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Override
+    public void updateLanguageRequest(Long requestId, HAccount actor,
+        RequestState state, String comment) throws EntityNotFoundException {
+        LanguageRequest languageRequest = languageRequestDAO.findById(requestId);
+        if (languageRequest != null) {
+            Request request = languageRequest.getRequest();
+
+            Request newRequest = request.update(actor, state, comment);
```
seanf: We should pass in the current time, and then request.update can ensure that oldRequest.validTo is equal to newRequest.validFrom.

zanata-war/src/main/java/org/zanata/service/impl/RequestServiceImpl.java
```diff
@@ -0,0 +1,172 @@
...
+            requestDAO.makePersistent(request);
+
+            //Need to add 1 second to request validTo date to avoid unique index constraint
+            newRequest.setValidTo(addOneSecond(request.getValidTo()));
```
seanf: If we set newRequest.validFrom to oldRequest.validTo, and leave newRequest.validTo as null, we shouldn't need to do this.

seanf:
@aeng I've finished reviewing for now.

zanata-model/src/main/java/org/zanata/model/type/RequestStateTypeDescriptor.java
aeng: Only in `RequestState.valueOf`

zanata-war/src/main/java/org/zanata/email/RequestToJoinLanguageEmailStrategy.java
aeng: We are sending email, but no html message.

zanata-model/src/main/java/org/zanata/model/type/RequestStateTypeDescriptor.java
seanf: Right, sorry. But some of the knowledge about this conversion does seem to be quite spread out. 

There should also only be one place which converts a String to RequestState. We have several places which call `charAt(0)` and then call `valueOf(char)`. I suggest adding a `valueOf(String)` method to RequestState, and using it everywhere instead of `RequestState.valueOf(string.charAt(0))`.

seanf:
- [ ] @aeng Do you think Request's EntityListener might really belong in TimeEntityBase?

zanata-war/src/main/java/org/zanata/email/RequestToJoinLanguageEmailStrategy.java
seanf: I originally misunderstood the intention of this change because of my badly named variable.  

- [ ] However, I think we agreed that there is no need to discard "user message" as a feature.

seanf:
@aeng @lukebrooker this branch has been broken by the jersey changes in api. It still works if you build https://github.com/zanata/zanata-api/commit/21c0158a2d414ede43c2544b3555b637ced495e4 locally.

This branch probably needs to be rebased (or master merged in), but perhaps we should wait until review is finished.

djansen-redhat:
- [ ] Request button should become Cancel Request button without refresh (causes UI issues if reused)
- [ ] Accepting a request does not update the members list
- [ ] Request acceptance notification is too short (milliseconds)
- [ ] Limit "Additional Information" field to something reasonable (500, or 1000 chars)

zanata-war/src/main/resources/messages.properties
```diff
@@ -843,7 +843,14 @@ jsf.MemberOfTeam=You are now a member of the {0} language team
 jsf.LeftTeam=You have left the {0} language team
 jsf.AddedAPermission={0} has been made a {1}
 jsf.RemovedAPermission={0} has been removed as {1}
-
+jsf.language.myRoles=You are <span class="label">{0}</span> in this language team.
+jsf.language.cancelMyRequest=Cancel my pending request
+jsf.language.request.cancelled=Request cancelled by {0}
+jsf.language.requestedRole=You have pending request for <span class="label">{0}</span>
+jsf.language.request.role=Requested role: {0}
+jsf.language.request.empty=No pending request
```
djansen-redhat: Should be 'requests'

lukebrooker:
## Display issues

- [ ] Clearly display the language code and native name for each language page
- [ ] "Your message had been sent…" should be "You request has been sent"
- [ ] Show only available roles next to each name, not ticks and crosses
- [ ] Add "Cancel my request" under button
- [ ] Remove icons from request & leave buttons
- [ ] Approve / decline buttons should have icons
- [ ] Remove icon from language name

## Other issues

- [ ] Change button to "Pending request" in disabled button after submitting request
- [ ] Only show one request/leave button
- [ ] When checking "co-ordinator" it should disable translator and reviewer
- [ ] On request approval, refresh is needed to add user to the members list
- [ ] Request to join team should submit the action immediately, not require modal
- [ ] Refresh is needed to show pending request
- [ ] After a request is approved/declined, when you switch to another tab, clear the requests list
- [ ] On approval, buttons should turn into checkboxes, allowing changing of roles

zanata-model/src/main/java/org/zanata/model/type/RequestState.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Getter
+    private char initial;
+
+    RequestState(char initial) {
+        this.initial = initial;
+    }
+
+    public static RequestState getEnum(String string) {
+        char initial = string.charAt(0);
```
seanf: - [ ] Should check for length=1

zanata-war/src/main/java/org/zanata/action/LanguageJoinAction.java
```diff
@@ -156,6 +203,60 @@ public void send() {
...
+        }
+
+        if(localeMember.isReviewer()) {
+            roles.add(StringUtils.lowerCase(msgs.get("jsf.Reviewer")));
+        }
+
+        if(roles.size() == 1) {
+            return msgs.format("jsf.language.myRole", Joiner.on(" and ").join(roles));
+        }
+        return msgs.format("jsf.language.myRoles", Joiner.on(" and ").join(roles));
```
seanf: This isn't translation safe because it assumes English plural rules, plus it concatenates hard-coded the English word "and" according to English grammar.

I suggest something like 

    jsf.language.myRoles:You have the following role/roles in this team: {0}

where {0} would be [admin] or [translator, coordinator].

Using "role/roles" is ugly, but at least it's consistently ugly in many languages. To fix it properly, we would need to use a different i18n approach, one which understands plurals. I think that's out of scope here.

zanata-war/src/main/java/org/zanata/action/LanguageJoinAction.java
```diff
@@ -75,54 +85,91 @@
...
 
-    public void bindRole(String role, boolean checked) {
-        if (role.equals("translator")) {
-            requestAsTranslator = checked;
-        } else if (role.equals("reviewer")) {
-            requestAsReviewer = checked;
-        } else if (role.equals("coordinator")) {
-            requestAsCoordinator = checked;
-        }
+    public String getRequestedRole(LanguageRequest request) {
```
seanf: - [ ] SF: looks like this should be getRequestedRoles or getLocalisedRequestedRoles, and/or it needs a javadoc

zanata-war/src/main/java/org/zanata/action/LanguageAction.java
```diff
@@ -265,6 +292,26 @@ private void redirectToLanguageHome() {
         return localeMemberDAO.findAllByLocale(new LocaleId(language));
     }
 
+    public String getMemberRoles(HLocaleMember member) {
```
seanf: - [ ] SF: getLocalisedMemberRoles?

zanata-war/src/main/java/org/zanata/action/LanguageJoinAction.java
seanf: - [ ] SF: getMyLocalisedRoles() ?

zanata-war/src/main/resources/messages.properties
```diff
@@ -1231,10 +1249,10 @@ jsf.AlreadyInGroup=Already in Group
 jsf.email.joingrouprequest.AdditionalInfoMessage=To ensure your request is processed without delay, please provide any additional information that will help the group maintainers to process your request.
 !FIXME make key similar to related keys
 !{0} = locale id. e.g. en-us, {1} language display name. e.g. English
-jsf.email.RequestToJoinLanguageTeam.title=Request to join {0} [{1}] language team
+jsf.email.RequestToJoinLanguageTeam.title=Request to join {0} language team
```
seanf: - [ ] SF: Did we really mean to remove the {1} language display name? It's still listed in the comment above.

zanata-war/src/main/resources/messages.properties
```diff
@@ -1231,10 +1249,10 @@ jsf.AlreadyInGroup=Already in Group
...
 jsf.email.JoinGroupRequest.Subject=Request to join group "{0}"
 jsf.language.email.joinrequest.Subject=User "{0}" request to join [{1}] language team
-jsf.language.email.AdditionalInfoMessage=Please provide any additional information that will help the team coordinators identify you and process your request.
+jsf.language.email.AdditionalInfoMessage=Provide some details about your experiance and how you would like to contribute
```
seanf: - [ ] SF: experiance -> experience

seanf:
@aeng I've reviewed it again, and added some new comments. Also, some of my old comments were not discussed or addressed. You can find them all if you search for "SF:" in the TODO list (http://userscripts-mirror.org/scripts/show/455500).

zanata-war/src/main/java/org/zanata/action/LanguageJoinAction.java
aeng: Yes, need to be public, also used in requests_tab.xhtml

djansen-redhat:

zanata-war/src/main/java/org/zanata/action/LanguageJoinAction.java
seanf: - [ ] SF: Where is this used? Seems like a better name would be `getMyRequestedRoles()` (plural).

seanf:
@aeng Could you please let me know about changes with a comment on the pull request? It's pretty easy to miss an IRC message.

You still haven't answered several of my SF questions.

aeng:
@seanf I can only see `SF: Where is this used? Seems like a better name would be getMyRequestedRoles() (plural).`  - Changed.

`SF: Also, perhaps these methods could be declared in an AbstractTimeBasedDAO to help ensure consistency - even if all the methods are abstract for now.` - Can't see which method you were talking about.

zanata-war/src/main/java/org/zanata/action/LanguageJoinAction.java
aeng: Its been removed.

zanata-war/src/main/java/org/zanata/dao/RequestDAO.java
aeng: Not sure if we need another abstract class for these method which are clearly for Request entity.

djansen-redhat:
- [ ] Members tab does not work the same as the Requests tab, in terms of checkbox clicking.
- [ ] User has to hard refresh the Members tab to see proper checkboxes if the Coorrdinator box was clicked

zanata-war/src/main/java/org/zanata/dao/RequestDAO.java
seanf: Okay, I guess it's not hard to change if and when we add more time-based entities.

aeng:
`Should a user be able to be a member of a team, but with no roles?`

I think its best to separate this logic out of this feature. We might need some data migration if we implement this.

djansen-redhat:

djansen-redhat:
- [ ] 'Request declined' reason should be emailed to the user

djansen-redhat:
- [ ] Accepting a request with reviewer will remove the user's "translator" role (and vice versa)

djansen-redhat:
:white_check_mark: Phew! Tested, with minor issues to be addressed later.
Waiting on Jenkins' green light.

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/8d41c54c011b7baf7a23e9a7c1e6121582c90127

PR #1011, opened 2015-10-13
update zanata-spa to commit 68a0859e97288ab494b5e01af6d06db2d8cd0dfa

davidmason:
This version has the header using React instead of Angular, but functionality
is not changed.

PR #1013, opened 2015-10-18
refactor: update zanata-spa to commit 68a0859e97288ab494b5e01af6d06db2d8cd0dfa

davidmason:
This version has the header using React instead of Angular, but functionality
is not changed.

Was originally targeting the wrong branch as #1011


djansen-redhat:
The test failure (build 4741) can be ignored.

PR #1014, opened 2015-10-19
fix(test): Replace return this with return new Page

djansen-redhat:
Return this is page functions bypasses the AbstractPage waits/checks.


djansen-redhat:
This fixes a few of the text input failures

PR #1018, opened 2015-10-22
Fix lock message at startup

seanf:


carlosmunoz:
um... :+1: ?

zanata-war/src/main/java/org/zanata/ZanataInit.java
```diff
@@ -257,7 +257,7 @@ private void checkLuceneLocks(File indexDir)
             // we don't trust Lucene's NativeFSLockFactory for NFS locks
             String docURL = "http://docs.jboss.org/hibernate/search/4.4/reference/en-US/html/search-configuration.html#search-configuration-directory-lockfactories";
             log.warn("The Hibernate Search index dir '{}' might be NFS. " +
-                    "NativeFSLockFactory might not be safe: {}" +
+                    "NativeFSLockFactory might not be safe: See {}",
                     indexDir, docURL);
```
seanf: @carlosmunoz the important change was replacing '+' with ','.  It was printing the arguments out of order, and it was really confusing until I got to the bottom of it.

djansen-redhat:
The test failure in build 4794 (ServerSettingsTest.setServerURLTest) can be ignored.

PR #1023, opened 2015-10-26
Work around path problem for JavaMelody

seanf:
On some platforms, the MonitoringFilter has no init parameters
by the time requests come in, which causes JavaMelody to appear
under /monitoring instead of /admin/monitoring. This change uses
a servlet context parameter instead of a filter parameter.


aeng:
:+1:  :+1:

djansen-redhat:
Woohoo! He fixed the thing!
:white_check_mark: Tested

PR #1026, opened 2015-10-28
Fix email message generated when user joining team

seanf:
It was:
> You can add user with username "Some User" to the "xx" team as translator using the "Add Team Member" action on the language team page.

whereas it should have been:

> You can add user with username "some_username" to the "xx" team as translator using the "Add Team Member" action on the language team page.


djansen-redhat:
:white_check_mark: Tested

PR #1044, opened 2015-11-18
fix(revert translation): fix null pointer exception when revert translation

aeng:
https://zanata.atlassian.net/browse/ZNTA-738


zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java
```diff
@@ -722,8 +722,8 @@ private String validateTranslations(ContentState newState,
                 runInTransaction(() -> {
                     for (HTextFlowTarget target : removedTargets) {
                         target =
-                                textFlowTargetDAO.findById(target.getId(),
-                                        true); // need to refresh from
+                            textFlowTargetDAO.findById(target.getId(),
+                                true); // need to refresh from
```
huangp: Looks like your formatter is not consistent to our standard one. The continuous indentation should be 8 spaces.

huangp:
other than the formatting :+1:

definite:
:+1:

PR #1045, opened 2015-11-22
fix(activation): display activation message

aeng:
Display activation message after redirect

https://zanata.atlassian.net/browse/ZNTA-720


huangp:
:+1:

definite:
:+1:

PR #1053, opened 2015-11-27
fix(url):handle whitespace and slash in doc id url

aeng:
https://zanata.atlassian.net/browse/ZNTA-767


carlosmunoz:
:+1: looks good

aeng:

definite:
Tested with Firefox 42.0
It uploaded with out problem, but after I clicked the document to translate that.
1. It shows **No document selected**
2. No entries was listed, and **(No Content)**  is shown.

definite:
Seems the error in my machine seems to be a configuration issue, not the scope of this PR. I will file another bug for that.
This bug should be verified.

PR #1056, opened 2015-11-27
Eliminate JS errors from functional tests

seanf:
Based on revised code from http://stackoverflow.com/a/4411986/14379

Also removing `setTimeout()` support: the `timeoutCounter` is not used, and
  the wrapper code is causing problems.

This should eliminate the server error during `EditHomePageTest.editPageCode` and `EditProjectAboutTest.addAboutPageDetails`: `Uncaught TypeError: func is not a function`

### DIFF
--- a/zanata-war/src/main/java/org/zanata/action/AjaxCounterBean.java
+++ b/zanata-war/src/main/java/org/zanata/action/AjaxCounterBean.java
@@ -34,40 +34,27 @@ import javax.servlet.http.HttpServletRequest;
 @Name("ajaxCounter")
 @Slf4j
 public class AjaxCounterBean {
+    // NB this doesn't handle setInterval or setTimeout
+    // TODO handle XHR.abort (NB: readystatechange listener triggers on Chrome 45, despite MDN docs)
     // http://stackoverflow.com/questions/4410218/trying-to-keep-track-of-number-of-outstanding-ajax-requests-in-firefox
-    private static final String AJAX_COUNTER_SCRIPT = "<script type=\"application/javascript\">" +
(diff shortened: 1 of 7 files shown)


zanata-war/src/main/java/org/zanata/action/AjaxCounterBean.java
```diff
@@ -34,40 +34,27 @@
...
-            "            }\n" +
-            "        };\n" +
-            "        this._send.apply(this, arguments);\n" +
+            "    if (xhr.active === undefined) {\n" +
+            "        xhr.active = 0;\n" +
+            "        var pt = xhr.prototype;\n" +
+            "        var _send = pt.send;\n" +
+            "        pt.send = function() {\n" +
+            "            xhr.active++;\n" +
+            "            this.addEventListener('readystatechange', function(e) {\n" +
```
carlosmunoz: Adding the event listener here could mean we add a listener for every fired request, so each request will decrement the counter as many times as requests have been fired in total.

carlosmunoz:
:+1:

seanf:
@carlosmunoz Turns out we have a lot of JS warnings I hadn't noticed. (Some of them probably because RichFaces is using jQuery 2 by accident.) I've changed the test failure trigger to errors, not warnings.

seanf:
@carlosmunoz Mind looking this over again?

carlosmunoz:
:+1:

djansen-redhat:

seanf:
@djansen-redhat Looks like this passed.

djansen-redhat:
:white_check_mark: Thank you Jenkins

seanf:
Correction: this pull request does not in fact abort tests in case of JS errors, because `logLogs()` is called with `false`.

PR #1057, opened 2015-11-27
Try to eliminate "Uncaught ReferenceError: zanata is not defined"

seanf:
We get even more of these in https://github.com/zanata/zanata-server/pull/886


seanf:
Incorporated into https://github.com/zanata/zanata-server/pull/1056

PR #1061, opened 2015-11-30
fix(login redirect): prevent redirect to login page after login

aeng:
https://zanata.atlassian.net/browse/ZNTA-815


carlosmunoz:
@aeng wrong target branch

zanata-war/src/main/java/org/zanata/security/UserRedirectBean.java
```diff
@@ -166,9 +167,13 @@ public boolean isRedirectToRegister() {
         return isRedirectTo(REGISTER_URL);
     }
 
+    public boolean isRedirectLoginPage() {
```
carlosmunoz: `isRedirectToLoginPage` to keep it consistent with the other similar method names.

aeng:
Moved to https://github.com/zanata/zanata-server/pull/1063.

PR #1070, opened 2015-12-14
Suppress FindBugs warnings in Lombok code

seanf:


zanata-test-war/src/main/java/org/zanata/rest/SampleDataResourceImpl.java
```diff
@@ -28,14 +26,16 @@
...
+
 /**
  * @author Patrick Huang <a
  *         href="mailto:pahuang@redhat.com">pahuang@redhat.com</a>
  */
 @RequestScoped
 @Path("/data/sample")
-@Slf4j
 public class SampleDataResourceImpl implements SampleDataResource {
+    private static final Logger log = getLogger(SampleDataResourceImpl.class);
```
carlosmunoz: Why are we avoiding lombok here?

seanf: > Why are we avoiding lombok here?
￼
Because it was triggering a compiler bug. Plus I've come to the conclusion that Lombok is a somewhat nasty compiler hack, so I like to eliminate Lombok when it's easy to do so.

carlosmunoz:
I think I had already reviewed this but forgot to mark it.
:+1:

PR #1076, opened 2016-01-04
Update statistics json + review statistic

aeng:
https://zanata.atlassian.net/browse/ZNTA-744

Depending on: https://github.com/zanata/zanata-api/pull/40

sample output:

```
{
  "username": "admin",
  "contributions": [
    {
      "locale": "zh-Hans",
      "translation-stats": {
        "approved": 0,
        "rejected": 0,
        "translated": 105,
        "needReview": 0
      }
    },
    {
      "locale": "fr",
      "translation-stats": {
        "approved": 65,
        "rejected": 0,
        "translated": 6,
        "needReview": 0
      },
      "review-stats": {
        "approved": 65,
        "rejected": 0
      }
    }
  ]
}
```


davidmason:
@aeng the output would be a bit easier to work with if "contributions" were an object, depending what a user it trying to do with it. To get the French statistics out of that sample output as it is now (I will call it `output`):

```javascript
import { findWhere } from 'lodash'

var frenchContributions = findWhere(output.contributions, { locale: 'fr' })
```

If we made it an object keyed by localeId, we could just do:

```javascript
var frenchContributions = output.contributions['fr']
```

On the other hand, the case of iterating over an object is a bit more error-prone than iterating over an array (when doing so by hand at least), so maybe it doesn't matter.

definite:

zanata-war/src/main/java/org/zanata/dao/TextFlowTargetHistoryDAO.java
```diff
@@ -75,54 +84,109 @@ public TextFlowTargetHistoryDAO(Session session) {
...
+                personId, fromDate, toDate, automatedEntry);
+        query.setComment("textFlowTargetHistoryDAO.getUserReviewStatisticInVersion");
+        return query.list();
+    }
+
+    private Query buildContributionStatisticQuery(boolean translations,
+            Long versionId, Long personId, Date fromDate, Date toDate,
+            boolean automatedEntry) {
+        String lastModifiedColumn =
+                translations ? "translated_by_id" : "reviewed_by_id";
```
huangp: why not just pass in the string value as parameter instead of using a boolean. It will be easier to read too

huangp:
:+1:  only a preference thing

zanata-war/src/main/java/org/zanata/dao/TextFlowTargetHistoryDAO.java
aeng: I'm using the boolean to do conditional `query.setParameterList`.

definite:
:+1:

aeng:
@zanata-jenkins please retest this

aeng:
Merged into integration/master. https://github.com/zanata/zanata-server/commit/7b4e1d6ec537a9e034d3e64c7f3035b25d32a7f4

Build job success:#5215

PR #1082, opened 2016-01-20
feat(user email): admin can configure to display user email

aeng:
https://zanata.atlassian.net/browse/ZNTA-879

API
- project contributors list - when configuration for display user set to true
- user service (private api for editor, /user/{username}) -  when configuration for display user set to true

UI
- user profile page - when logged in and configuration for display user set to true

### DIFF
--- a/zanata-war/src/main/java/org/zanata/rest/editor/service/UserService.java
+++ b/zanata-war/src/main/java/org/zanata/rest/editor/service/UserService.java
@@ -10,6 +10,7 @@ import javax.ws.rs.core.Response;
 import javax.inject.Named;
 import org.apache.deltaspike.jpa.api.transaction.Transactional;
 import org.apache.commons.lang.StringUtils;
+import org.zanata.ApplicationConfiguration;
 import org.zanata.dao.AccountDAO;
 import org.zanata.dao.PersonDAO;
 import org.zanata.model.HAccount;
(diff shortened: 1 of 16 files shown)


djansen-redhat:
Can I suggest making it look more like the option "Email log" below it? e.g. large text and help text visible

aeng:
@djansen-redhat done.

djansen-redhat:
The failure in build 5279 can be ignored.

zanata-war/src/main/java/org/zanata/rest/editor/service/resource/UserResource.java
```diff
@@ -59,7 +59,7 @@
      *
      * @param username
      */
-    public User generateUser(String username);
+    public User generateUser(String username, boolean includeEmail);
```
huangp: I know this is already there but this method and the other one is not REST methods. They can stay private in the implementation class

zanata-war/src/main/java/org/zanata/rest/editor/service/UserService.java
```diff
@@ -69,7 +74,7 @@ public Response getUserInfo(String username) {
     }
 
     @Override
-    public User generateUser(String username) {
+    public User generateUser(String username, boolean includeEmail) {
```
huangp: Why not just leave the method unchanged and use applicationConfiguration.isDisplayUserEmail() to get the boolean value again.

huangp:
:+1:

djansen-redhat:
"transferToUser(org.zanata.model.HAccount,boolean) has private access in org.zanata.rest.editor.service.UserService"

djansen-redhat:
These seems to work fine - can I just confirm that the maintainer list placement of email was removed from the spec?

djansen-redhat:
:white_check_mark: Tested

PR #1083, opened 2016-01-21
chore: use deltaspike for data access.

carlosmunoz:
Includes a first instance of a Deltaspike repository class, as well as
replacing some uses of DAOs by this new class. This class shows how to
use some of the features in repositories.

---

This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1083)

### DIFF
--- a/zanata-war/src/main/java/org/zanata/repository/LocaleRepository.java
+++ b/zanata-war/src/main/java/org/zanata/repository/LocaleRepository.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2016, Red Hat, Inc. and individual contributors as indicated by the
+ * @author tags. See the copyright.txt file in the distribution for a full
+ * listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation; either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * This software is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this software; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA, or see the FSF
+ * site: http://www.fsf.org.
+ */
+package org.zanata.repository;
+
+import org.apache.deltaspike.data.api.AbstractEntityRepository;
+import org.apache.deltaspike.data.api.Repository;
+import org.zanata.common.LocaleId;
+import org.zanata.model.HLocale;
+
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Root;
+import java.io.Serializable;
+import java.util.List;
+
+/**
+ * Repository for the HLocale entity.
+ *
+ * @author Carlos Munoz <a href="mailto:camunoz@redhat.com">camunoz@redhat.com</a>
+ */
+@Repository
+public abstract class LocaleRepository
+        extends AbstractEntityRepository<HLocale, Long>
+        implements Serializable {
+
+    public abstract List<HLocale> findByActiveEqual(boolean active);
+
+    public abstract HLocale findOptionalByLocaleId(LocaleId localeId);
+
+    public List<HLocale> findByActive() {
+        return findByActiveEqual(true);
+    }
+
+    public HLocale findByLocaleId(LocaleId localeId) {
+        return findOptionalByLocaleId(localeId);
+    }
+
+    public List<HLocale> findAllActiveAndEnabledByDefault() {
+        // TODO To get rid of some of this boilerplate code and/or use
+        // DeltaSpike's criteria API, we might need to activate JPA metamodel
+        // generation
+        CriteriaQuery<HLocale> criteria = criteriaQuery();
+        CriteriaBuilder cb =
+                entityManager().getEntityManagerFactory()
+                        .getCriteriaBuilder();
+        Root<HLocale> root = criteria.from(entityClass());
+        criteria.where(
+                cb.and(
+                        cb.equal(root.get("active"), true),
+                        cb.equal(root.get("enabledByDefault"), true)
+                )
+        );
+
+        return entityManager().createQuery(criteria).getResultList();
+    }
+}
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     <!-- Applies to surefire and failsafe (if enabled) -->
     <failIfNoTests>true</failIfNoTests>
 
-    <deltaspike.version>1.5.1</deltaspike.version>
+    <deltaspike.version>1.5.2</deltaspike.version>
     <cdi.runtime>runtime</cdi.runtime>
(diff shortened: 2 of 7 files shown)


carlosmunoz:
@zanata-jenkins please retest this

carlosmunoz:

seanf:
@carlosmunoz did you see the compilation error? 

zanata-war/src/test/java/org/zanata/service/impl/LocaleServiceImplTest.java:[72,35] cannot find symbol
  symbol:   method setLocaleDAO(org.zanata.dao.LocaleDAO)
  location: variable testLocaleServiceImpl of type org.zanata.service.impl.LocaleServiceImpl

zanata-war/src/main/java/org/zanata/repository/LocaleRepository.java
```diff
@@ -0,0 +1,74 @@
...
+@Repository
+public abstract class LocaleRepository
+        extends AbstractEntityRepository<HLocale, Long>
+        implements Serializable {
+
+    public abstract List<HLocale> findByActiveEqual(boolean active);
+
+    public abstract HLocale findOptionalByLocaleId(LocaleId localeId);
+
+    public List<HLocale> findByActive() {
```
seanf: Could we call this `findActive()`?

carlosmunoz: Unfortunately no. The magic happens based on the name of the method, and it must start with `findBy` and follow certain rules (see https://deltaspike.apache.org/documentation/data.html#QueryMethodExpressions)

zanata-war/src/main/java/org/zanata/repository/LocaleRepository.java
```diff
@@ -0,0 +1,74 @@
...
+        criteria.where(
+                cb.and(
+                        cb.equal(root.get("active"), true),
+                        cb.equal(root.get("enabledByDefault"), true)
+                )
+        );
+
+        return entityManager().createQuery(criteria).getResultList();
+    }
+}
```
seanf: It's a shame this method is a lot longer than the LocaleDAO version:

    public List<HLocale> findAllActiveAndEnabledByDefault() {
        return findByCriteria(Restrictions.eq("active", true),
                Restrictions.eq("enabledByDefault", true));
    }

This class is also missing a more complex method:

    public List<HLocale> findBySimilarLocaleId(LocaleId localeId) {
        return (List<HLocale>) getSession()
                .createQuery("from HLocale l where lower(l.localeId) = :id ")
                .setString("id", localeId.getId().toLowerCase())
                .setComment("LocaleDAO.findBySimilarLocaleId").list();
    }

Otherwise we might at least be able to get rid of LocaleDAO.

seanf:
:+1: It looks fine otherwise, although there doesn't seem to be any code saving yet.

zanata-war/src/main/java/org/zanata/repository/LocaleRepository.java
carlosmunoz: So, the more complex method is because the generic JPA criteria API is really complex. I think we can wither encapsulate some of the boilerplate code, or use some kind of wrapper (like Deltaspike's one). We will need to activate JPA's metamodel generation which basically gives you compiled classes with all your tables and fields.

We can also completely replace the DAO by moving this method into the repository. Since the repository is using entity manager, I would like to keep it that way, but there's nothing stopping us from continuing to use hibernate sessions.

carlosmunoz:
It's more of a first attempt at using the data framework. Still not sure if we want to continue to tightly couple the code with Deltaspike.

zanata-war/src/main/java/org/zanata/repository/LocaleRepository.java
seanf: Even when you're apparently writing the implementation yourself, like this
method?

On 11 February 2016 at 15:22, Carlos Munoz <notifications@github.com> wrote:

> In zanata-war/src/main/java/org/zanata/repository/LocaleRepository.java
> <https://github.com/zanata/zanata-server/pull/1083#discussion_r52565309>:
>
> > +
> > +/**
> > + * Repository for the HLocale entity.
> > + *
> > + * @author Carlos Munoz <a href="mailto:camunoz@redhat.com">camunoz@redhat.com</a>
> > + */
> > +@Repository
> > +public abstract class LocaleRepository
> > +        extends AbstractEntityRepository<HLocale, Long>
> > +        implements Serializable {
> > +
> > +    public abstract List<HLocale> findByActiveEqual(boolean active);
> > +
> > +    public abstract HLocale findOptionalByLocaleId(LocaleId localeId);
> > +
> > +    public List<HLocale> findByActive() {
>
> Unfortunately no. The magic happens based on the name of the method, and
> it must start with findBy and follow certain rules (see
> https://deltaspike.apache.org/documentation/data.html#QueryMethodExpressions
> )
>
> —
> <https://github.com/zanata/zanata-server/pull/1083/files#r52565309>.
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

zanata-war/src/main/java/org/zanata/repository/LocaleRepository.java
seanf: On 11 February 2016 at 15:31, Carlos Munoz <notifications@github.com> wrote:

> In zanata-war/src/main/java/org/zanata/repository/LocaleRepository.java
> <https://github.com/zanata/zanata-server/pull/1083#discussion_r52565616>:
>
> > +        // generation
> > +        CriteriaQuery<HLocale> criteria = criteriaQuery();
> > +        CriteriaBuilder cb =
> > +                entityManager().getEntityManagerFactory()
> > +                        .getCriteriaBuilder();
> > +        Root<HLocale> root = criteria.from(entityClass());
> > +        criteria.where(
> > +                cb.and(
> > +                        cb.equal(root.get("active"), true),
> > +                        cb.equal(root.get("enabledByDefault"), true)
> > +                )
> > +        );
> > +
> > +        return entityManager().createQuery(criteria).getResultList();
> > +    }
> > +}
>
> So, the more complex method is because the generic JPA criteria API is
> really complex.
>
That sounds like another argument for Zanata to use Hibernate and not pure
JPA, where there's an option.

I think we can either encapsulate some of the boilerplate code, or use some
> kind of wrapper (like Deltaspike's one). We will need to activate JPA's
> metamodel generation which basically gives you compiled classes with all
> your tables and fields.
>

It might be time we used Querydsl too, which even gets a mention here:
https://deltaspike.apache.org/documentation/data.html#QueryDelegates


> We can also completely replace the DAO by moving this method into the
> repository. Since the repository is using entity manager, I would like to
> keep it that way, but there's nothing stopping us from continuing to use
> hibernate sessions.
>


-- 
Sean Flanigan

Principal Software Engineer
Red Hat

carlosmunoz:
@seanf I just realized this will imply a big change in tests. We currently use Autowire to inject the DAOs. With the new Deltaspike repos we can't do that as there are some auto-provided functions.

One option is to change some of these tests to use CDIunit, but then this becomes a bigger change. Do you think it's worth going down that path at this stage?

seanf:
> @seanf I just realized this will imply a big change in tests. We currently use Autowire to inject the DAOs. With the new Deltaspike repos we can't do that as there are some auto-provided functions.

> One option is to change some of these tests to use CDIunit, but then this becomes a bigger change. Do you think it's worth going down that path at this stage?

It doesn't seem like DS repos will get us much just yet. I'm assuming CDIAutowire won't be able to supply the repository implementations either (requires a DeltaSpike extension, am I right?)

Perhaps we should put this on hold until we're ready to use CDI-Unit for all these tests (including parameterised tests).

definite:

carlosmunoz:
Closing this PR as everything has moved to the zanata-platform repo. Will revisit this if it's needed in the future.

PR #1084, opened 2016-01-21
fix(applicationConfig): update dao scope to RequestScope

aeng:
https://zanata.atlassian.net/browse/ZNTA-882

### DIFF
--- a/zanata-war/src/main/java/org/zanata/dao/ApplicationConfigurationDAO.java
+++ b/zanata-war/src/main/java/org/zanata/dao/ApplicationConfigurationDAO.java
@@ -1,12 +1,14 @@
 package org.zanata.dao;
 
 import org.hibernate.Session;
+
+import javax.enterprise.context.RequestScoped;
 import javax.inject.Named;
 import org.zanata.model.HApplicationConfiguration;
 
 @Named("applicationConfigurationDAO")
 
-@javax.enterprise.context.Dependent
+@RequestScoped
 public class ApplicationConfigurationDAO extends
         AbstractDAOImpl<HApplicationConfiguration, Long> {
 


zanata-war/src/main/java/org/zanata/dao/ApplicationConfigurationDAO.java
```diff
@@ -1,12 +1,14 @@
...
 import org.hibernate.Session;
+
+import javax.enterprise.context.RequestScoped;
 import javax.inject.Named;
 import org.zanata.model.HApplicationConfiguration;
 
 @Named("applicationConfigurationDAO")
 
-@javax.enterprise.context.Dependent
+@RequestScoped
```
carlosmunoz: Not sure about this change alex. Maybe I'm being overly cautious, but DAOs I think should remain dependent scoped as they should not have any state, hence any instance of them should work regardless of the context. Could you elaborate on how this fixes the problem?

aeng:
All the DAO is using `RequestScoped` and I can't reproduce the issue with this change.

aeng:
Correction: Cannot reproduce issue without any changes.

zanata-war/src/main/java/org/zanata/dao/ApplicationConfigurationDAO.java
carlosmunoz: I think you're right... all other DAOs are request scoped. In light of this I remove my objection :wink: 

@seanf @huangp  Do you remember why we had to make them Request scoped instead of dependent? Proxies perhaps?

huangp: Because EntityManager and Session are Request scope. DAO makes sense to
stay in the same scope.

On Fri, Jan 22, 2016 at 10:17 AM, Carlos Munoz <notifications@github.com>
wrote:

> In
> zanata-war/src/main/java/org/zanata/dao/ApplicationConfigurationDAO.java
> <https://github.com/zanata/zanata-server/pull/1084#discussion_r50486499>:
>
> >  import javax.inject.Named;
> >  import org.zanata.model.HApplicationConfiguration;
> >
> >  @Named("applicationConfigurationDAO")
> >
> > -@javax.enterprise.context.Dependent
> > +@RequestScoped
>
> I think you're right... all other DAOs are request scoped. In light of
> this I remove my objection [image: :wink:]
>
> @seanf <https://github.com/seanf> @huangp <https://github.com/huangp> Do
> you remember why we had to make them Request scoped instead of dependent?
> Proxies perhaps?
>
> —
> <https://github.com/zanata/zanata-server/pull/1084/files#r50486499>.
>



-- 
Patrick Huang
Senior Software Engineer
Engineering - Internationalisation
Red Hat, Asia-Pacific Pty Ltd
Level 1, 193 North Quay
Brisbane 4000
Office: +61 7 3514 8278
Fax: +61 7 3514 8199
IRC: pahuang
github: github.com/huangp
Website: www.redhat.com

djansen-redhat:
I gave this a cursory check to make sure it doesn't break anything, but that's about it.

zanata-war/src/main/java/org/zanata/dao/ApplicationConfigurationDAO.java
seanf: Yes, I think @Dependent is best used only for classes which *have* to be
Dependent, like String - the lack of proxies is a subtle danger we don't
need to put up with for ordinary beans.

PR #1089, opened 2016-02-02
Add fedora zanata

definite:


definite:
Seem like I should not change the branch when create the pull request.

PR #1090, opened 2016-02-02
doc(user-guide): Add fedora.zanata.org

definite:


djansen-redhat:
:x: Incorrect target branch

PR #1094, opened 2016-02-15
fix(translation memory): remove 0% matching TM entry

aeng:
https://zanata.atlassian.net/browse/ZNTA-905

### DIFF
--- a/zanata-war/src/main/java/org/zanata/service/impl/TranslationMemoryServiceImpl.java
+++ b/zanata-war/src/main/java/org/zanata/service/impl/TranslationMemoryServiceImpl.java
@@ -123,6 +123,8 @@ public class TranslationMemoryServiceImpl implements TranslationMemoryService {
 //    private static final float BOOST_PROJITERSLUG = SysProperties.getFloat(
 //            SysProperties.TM_BOOST_PROJITERSLUG, 1.5f);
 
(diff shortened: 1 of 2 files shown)


zanata-war/src/main/java/org/zanata/service/impl/TranslationMemoryServiceImpl.java
```diff
@@ -359,9 +359,16 @@ private void processIndexMatch(TransMemoryQuery transMemoryQuery,
...
             TransMemoryResultItem.MatchType matchType =
                     fromContentState(textFlowTarget.getState());
-            TransMemoryResultItem item = createOrGetResultItem(transMemoryQuery, matchesMap, match, matchType,
-                    textFlowContents, targetContents);
-            addTextFlowTargetToResultMatches(textFlowTarget, item);
+
+            double percent =
+                calculateSimilarityPercentage(transMemoryQuery,
+                    textFlowContents);
+            if (percent != 0) {
```
huangp: I think it include 0% TM is because the actual result is of type double. So a 0.005 (0.5%) match is either truncated or rounded to 0%. Here you may need to check if the percent is too small instead. e.g. less than 0.01 or 0.05

huangp:
:+1:

definite:

definite:

definite:
:+1:

PR #1095, opened 2016-02-19
fix: kerberos ticket-based authentication

carlosmunoz:
With the move to CDI, using reflection to modify properties on CDI
proxied beans is not recommended. This change addresses said issue when
propagating the Kerberos authentication context onto the Zanata identity
bean.

### DIFF
--- a/zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
+++ b/zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
@@ -185,6 +185,9 @@ public class ZanataIdentity implements Identity, Serializable {
 
     /**
      * Accepts an external subject and principal. (Use with caution)
+     * This method is used to propagate an authentication context. For
+     * example when spawing a new thread for an async task, or when
+     * authenticating externally through Kerberos.
      */
     public void acceptExternalSubjectAndPpal(Subject subject,
             Principal principal) {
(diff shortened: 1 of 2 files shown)


zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
```diff
@@ -185,6 +185,9 @@ public void acceptExternallyAuthenticatedPrincipal(Principal principal) {
 
     /**
      * Accepts an external subject and principal. (Use with caution)
+     * This method is used to propagate an authentication context. For
+     * example when spawing a new thread for an async task, or when
+     * authenticating externally through Kerberos.
      */
     public void acceptExternalSubjectAndPpal(Subject subject,
```
seanf: So the method already existed? Perhaps if it had had a less "optimised" name it would have been easier to find.

seanf:
> With the move to CDI, using reflection to modify properties on CDI proxied beans is not recommended. 

I think the problem with CDI proxies would happen with any direct field access, via reflection or not.

seanf:
:+1:

zanata-war/src/main/java/org/zanata/security/ZanataIdentity.java
carlosmunoz: Yeah, method was there.

carlosmunoz:
@seanf Can I mark this as reviewed and send to QA?

seanf:
Sorry, yes, that's what I meant by the :+1:

definite:
:+1:

PR #1072, opened 2015-12-16
Clean up DeltaSpike scopes, windowIds and test logging

seanf:
https://zanata.atlassian.net/browse/ZNTA-833

PR #1099, opened 2016-02-25
Add DswidParamChecker to detect dswid changes in tests

seanf:


seanf:
This change is expected to cause test failures at least until https://github.com/zanata/zanata-server/pull/1072 goes in (since it cleans up our dswid parameters quite a bit).

djansen-redhat:
Build 5384 seems to have broken for other reasons - Caused by: java.net.MalformedURLException: unknown protocol: data

seanf:
No, I think that was my new URL code. Thanks @djansen-redhat

djansen-redhat:
:+1: This looks good to me, does it need a manual test you think @seanf ?
What about instances where we would send the browser to a url directly - from my understanding of the code that would be handled by !dswid.isPresent()?

seanf:
On 26 February 2016 at 14:47, Damian Jansen <notifications@github.com>
wrote:

> [image: :+1:] This looks good to me, does it need a manual test you think
> @seanf <https://github.com/seanf> ?
>

No, shouldn't.


> What about instances where we would send the browser to a url directly -
> from my understanding of the code that would be handled by
> !dswid.isPresent()?
>
It will just log a warning, at present, if the new URL lacks a dswid.  What
sort of URL would it be? Just the home page?


> —
> <https://github.com/zanata/zanata-server/pull/1099#issuecomment-189114242>
> .
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

djansen-redhat:
@seanf oh - true, we do it all the time with goToHome(). Never mind!

seanf:
Would there be others? I would actually prefer to make "losing the dswid"
trigger test failures, if possible. But first I suppose I need to find a
way to ignore direct navigation.

On 26 February 2016 at 15:53, Damian Jansen <notifications@github.com>
wrote:

> @seanf <https://github.com/seanf> oh - true, we do it all the time with
> goToHome(). Never mind!
>
> —
> <https://github.com/zanata/zanata-server/pull/1099#issuecomment-189124015>
> .
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

djansen-redhat:
Not sure, off the top of my head. Something to introduce later with more consideration?

seanf:
On 26 February 2016 at 16:21, Damian Jansen <notifications@github.com>
wrote:

> Not sure, off the top of my head. Something to introduce later with more
> consideration?
>

Certainly. That enhancement can wait.

But you're right, this thing does need a test. Also, I need to reset its
state between tests, so that it can never break the wrong test.


> —
> <https://github.com/zanata/zanata-server/pull/1099#issuecomment-189128994>
> .
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

seanf:
@djansen-redhat I've added a reset between tests. 

On second thought, I don't think it's worth writing a unit test for DswidParamChecker (deliberately changing the dswid to trigger a failure) - this test will probably just fail when we implement "a way to ignore direct navigation". I didn't mean it needed another manual test - I have seen it fail when the dswid changed.

djansen-redhat:
The build failure 9393 'JoinLanguageTeamTest.translatorJoinsLanguageTeam' can be ignored

djansen-redhat:
:+1: Looks good

PR #1101, opened 2016-02-29
feat(webhook): event for translation update

aeng:
https://zanata.atlassian.net/browse/ZNTA-938

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -34,6 +34,11 @@
 
 -----------------------
 
+<h5>New Features</h5>
+* [ZNTA-938](https://zanata.atlassian.net/browse/ZNTA-938) - Webhook event for translation update by user.
+
+-----------------------
+
 ## 3.8.2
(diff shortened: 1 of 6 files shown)


seanf:
I think this approach will prove too expensive.

1. I don't think we can afford to post a webhook event every time a single translation is changed. (We should batch them up, say once a minute.)
2. The processWebHookEvent call looks expensive (even ignoring the webhook I/O) and seems to be synchronous, which means it will slow down translation saves.
3. *If* we really needed to have an event which is fired on every single translation, I don't think webhooks are efficient enough - we'd be better off with an efficient event bus, perhaps JMS-based, or perhaps websockets.

Ultimately, I think translation/document-level events like this ought to tie in with the work @carlosmunoz is doing for event-driven statistics/status changes.

seanf:
@aeng see the branch https://github.com/zanata/zanata-server/tree/stats-refactor for the WIP by @carlosmunoz

aeng:
Moved to new PR - https://github.com/zanata/zanata-server/pull/1137

PR #1104, opened 2016-03-02
fix(tm action): fix TranslationMemoryAction.lastTaskResult

aeng:
https://zanata.atlassian.net/browse/ZNTA-793


huangp:
:+1:

djansen-redhat:
Checked it didn't break async process (using TM clear)

PR #1107, opened 2016-03-04
convert tests to cdiunit

carlosmunoz:
Get rid of Autowire if possible and replace it with cdi-unit based tests.

When building cdiunit tests keep these in mind:

You must run with the CdiUnitRunner by annotating the test class with:

```java
@RunWith(CdiUnitRunner.class)
```

If you want to mock any injected dependencies, you can do so like this:

```java
@Produces @Mock Serializable mock;
```

The cdiunit runner will mock the dependencies, so there is no need to manually call mockito.

You can also provide concrete dependencies:

```java
@Produces
public EntityManager getEntityManager() {
    // create the dependency here
}
```

Weld will complain if there are unsatisfied dependencies, and all dependency injection rules will apply to the test.

Additionally, the test will load a very limited set of dependency classes. If you wish for the test to consider more classes you need to annotate the test class with:

```java
@AdditionalClasses({ Class1.class, Class2.class })
```

Not all scopes will be activated by default when running a cdiunit test, to activate both session and/or request scope, you have to annotate the test methid with:

```java
@InRequestScope
```

If you encounter ambiguous dependencies because of your test producers, you can annotate your producer to make it the default one like this:

```java
@Produces @ProducesAlternative
public Serializable myProducer() {
...
}
```

To be able to use the ServiceLocator, as well as any other core DeltaSpike artifacts, the test class needs to be annotated with:

```java
@SupportDeltaspikeCore
```

Also adding JUnit rules to replace the old inheritance-based JUnit boilerplate code. There is a JUnit rule now called `JpaRule` which will initiate a JPA entity manager factory **and** entity manager before each test. Use is as follows:

```java
@ClassRule
@Rule
public static JpaRule jpaRule = new JpaRule();
```

To run DBUnit test change sets, you can use the new `DBUnitDataSetRunner` class as follows:

```java
DBUnitDataSetRunner runner =
                new DBUnitDataSetRunner(jpaRule.getEntityManager());
        runner.runDataSetOperations(
                new DataSetOperation(
                        "org/zanata/test/model/ClearAllTables.dbunit.xml",
                        DatabaseOperation.CLEAN_INSERT),
                new DataSetOperation(
                        "org/zanata/test/model/LocalesData.dbunit.xml",
                        DatabaseOperation.CLEAN_INSERT),
                new DataSetOperation(
                        "org/zanata/test/model/AccountData.dbunit.xml",
                        DatabaseOperation.CLEAN_INSERT),
                new DataSetOperation(
                        "org/zanata/test/model/CopyTransTestData.dbunit.xml",
                        DatabaseOperation.CLEAN_INSERT)
        );
```


carlosmunoz:
@seanf Do you mind reviewing? The diff is large so Github is omitting some changes, but the tests are so different now you might be better off looking at their end state.

zanata-war/src/main/java/org/zanata/webtrans/server/rpc/AbstractActionHandler.java
```diff
@@ -15,7 +15,9 @@
         implements ActionHandler<A, R> {
 
     protected AbstractActionHandler() {
-        assert getActionType() != null;
+        // TODO This is a sanity check only. It has been disabled to allow for
+        // CDIUnit tests
+        //assert getActionType() != null;
```
seanf: Was this meant to be permanently disabled?

zanata-war/src/test/java/org/zanata/rest/service/ResourceUtilsTest.java
```diff
@@ -60,13 +68,9 @@ public static void logMemoryForTests() {
...
-    public void initializeResourceUtils() {
-        MockitoAnnotations.initMocks(this);
-        resourceUtils =
-                SeamAutowire.instance().reset()
-                    .use("entityManager", mockEm)
-                    .use("localeDAO", mockLocaleDAO)
-                    .autowire(ResourceUtils.class);
+    public void initializeRequestScope() {
+        // NB: this is easier than adding @InRequestScope to all test methods
+        contextController.openRequest();
```
seanf: Should we be doing this in other places?

zanata-war/src/test/java/org/zanata/service/impl/MergeTranslationsServiceImplTest.java
```diff
@@ -196,10 +227,12 @@ public void testMergeTranslations1() {
...
         //entity in preMergeData should be updated after merge process
 
-        verify(spyService).mergeTranslationBatch(Matchers.eq(
+        // TODO This verification is not correctly detecting that the method
+        // has been invoked
+        /*verify(spyService).mergeTranslationBatch(Matchers.eq(
                 expectedSourceVersion), Matchers.eq(expectedTargetVersion),
                 Matchers.anyList(), Matchers.eq(useNewerTranslation),
-                Matchers.anyInt(), Matchers.anyInt());
+                Matchers.anyInt(), Matchers.anyInt());*/
```
seanf: We might need to complete this TODO.

zanata-war/src/test/java/org/zanata/service/impl/RequestServiceImplTest.java
```diff
@@ -132,7 +132,8 @@ public void testUpdateLanguageRequest() {
         service.updateLanguageRequest(languageRequest.getId(), requester,
             RequestState.ACCEPTED, comment);
 
-        verify(requestUpdatedEvent).fire(isA(RequestUpdatedEvent.class));
+        // TODO Mock events to do this kind of check
+        //verify(requestUpdatedEvent).fire(isA(RequestUpdatedEvent.class));
```
seanf: We might need to complete this TODO.

zanata-war/src/test/java/org/zanata/test/rule/DataSetOperation.java
```diff
@@ -0,0 +1,154 @@
...
+    public void prepare(boolean replaceNull) {
+        if (dataSet == null)
+            return;
+
+        if (replaceNull) {
+            dataSet.addReplacementObject("[NULL]", null);
+        }
+        // TODO This method doesn't implement the functionality of its
+        // original form  (see ZanataDbUnitJpaTest) with regard to the binary
+        // dir replacement.
```
seanf: I don't understand this comment.

zanata-war/src/test/java/org/zanata/test/DBUnitDataSetRunner.java
```diff
@@ -0,0 +1,105 @@
...
+
+import javax.persistence.EntityManager;
+
+/**
+ * Utility test class to run DBUnit data sets using a provided entity manager.
+ * Use this instead of extending {@link org.zanata.ZanataDbunitJpaTest}.
+ *
+ * @author Carlos Munoz <a href="mailto:camunoz@redhat.com">camunoz@redhat.com</a>
+ */
+public class DBUnitDataSetRunner {
```
seanf: Should we deprecate ZanataDbunitJpaTest? Or implement it using this class?

zanata-war/src/test/java/org/zanata/test/rule/FunctionalTestRule.java
```diff
@@ -0,0 +1,117 @@
...
+
+    // For reentrant rules
+    protected int counter = 0;
+
+    @Override
+    public Statement apply(Statement base, Description description) {
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                boolean isClassLevel = description.getMethodName() == null;
```
seanf: I've been wanting a way for JUnit Rules to ensure that they are used in the appropriate way (class vs method). Hmm.

zanata-war/src/main/java/org/zanata/webtrans/server/rpc/AbstractActionHandler.java
carlosmunoz: It was, but I actually found a way to re-enable it.

Turns out I just needed to decorate the `@ActionHandlerFor` annotation with `@Inherited` in order for proxies to get the annotation as well.

zanata-war/src/test/java/org/zanata/rest/service/ResourceUtilsTest.java
carlosmunoz: You can either do this, or annotate the test methods like the comment says. I think either way is fine, but the annotation might give you more fine-grained control if you want it.

zanata-war/src/test/java/org/zanata/test/rule/DataSetOperation.java
carlosmunoz: This is about some functionality present in `ZanataDBUnitJpaTest` which is not ported over to this new class, and which replaces the string `[binaryDir]` with the current locale directory. I did not bring it over as I don't think it's being used.

zanata-war/src/test/java/org/zanata/test/DBUnitDataSetRunner.java
carlosmunoz: Deprecate I think. I'd rather use Junit rules than class inheritance, but either way is fine by me. Outside of the scope for this PR.

carlosmunoz:
@seanf Can you please have a second look?

zanata-war/src/test/java/org/zanata/test/rule/DataSetOperation.java
seanf: Okay. In that case, I'd say get rid of the TODO comment, since we don't
plan to do it.

On 23 March 2016 at 14:44, Carlos Munoz <notifications@github.com> wrote:

> In zanata-war/src/test/java/org/zanata/test/rule/DataSetOperation.java
> <https://github.com/zanata/zanata-server/pull/1107#discussion_r57109370>:
>
> > +    }
> > +
> > +    public DatabaseOperation getOperation() {
> > +        return operation;
> > +    }
> > +
> > +    public void prepare(boolean replaceNull) {
> > +        if (dataSet == null)
> > +            return;
> > +
> > +        if (replaceNull) {
> > +            dataSet.addReplacementObject("[NULL]", null);
> > +        }
> > +        // TODO This method doesn't implement the functionality of its
> > +        // original form  (see ZanataDbUnitJpaTest) with regard to the binary
> > +        // dir replacement.
>
> This is about some functionality present in ZanataDBUnitJpaTest which is
> not ported over to this new class, and which replaces the string
> [binaryDir] with the current locale directory. I did not bring it over as
> I don't think it's being used.
>
> —
> <https://github.com/zanata/zanata-server/pull/1107/files/04678b5bd3f9626bca1d1dd92c7e25136727f9c7#r57109370>
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

zanata-war/src/test/java/org/zanata/service/impl/RequestServiceImplTest.java
```diff
@@ -133,7 +138,9 @@ public void testUpdateLanguageRequest() {
             RequestState.ACCEPTED, comment);
 
         // TODO Mock events to do this kind of check
-        //verify(requestUpdatedEvent).fire(isA(RequestUpdatedEvent.class));
+        List<RequestUpdatedEvent> firedEvents =
```
seanf: Does the TODO comment still apply?

zanata-war/src/test/java/org/zanata/test/DBUnitDataSetRunner.java
seanf: I think we should deprecate ZanataDbunitJpaTest or at least add a comment there suggesting this approach instead.

seanf:

zanata-war/src/test/java/org/zanata/test/DBUnitDataSetRunner.java
carlosmunoz: I will add a comment suggesting the approach.

djansen-redhat:
:white_check_mark: Tested by Jenkins

seanf:
:+1: Just need to get rid of the obsolete comment "TODO Mock events to do this kind of check"

carlosmunoz:
Merged in 98016cd142303592a16348007ceee2bd8aced817

PR #1108, opened 2016-03-07
Add better handling for server errors

seanf:


davidmason:
:+1:  Looks fine to me. The only question is that in one place the condition for logging a stack trace is `== 500`, and in another place the condition for logging is `>= 500` - is the difference on purpose?

It shouldn't make any practical difference since the non-500 codes all seem to be things that happen outside the application (e.g. network issues, throttling, etc.).

seanf:
@davidmason I hadn't noticed that. You're right that it probably doesn't matter, but I've changed it to be more consistent.

djansen-redhat:
:white_check_mark: Looks good to me

PR #1109, opened 2016-03-08
Fix warning: JSF1091: No mime type could be found for file zanata.woff

seanf:


djansen-redhat:
:white_check_mark: Verified by eyeballing the log

PR #1110, opened 2016-03-10
chore(dependency): Update dependencies to what Fedora 23 has

definite:
Update xom:xom to com.io7m.xom

See also:
https://github.com/zanata/zanata-parent/pull/52
https://github.com/zanata/zanata-common/pull/20

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1341,11 +1341,6 @@
         <artifactId>functionaljava</artifactId>
         <version>3.1</version>
       </dependency>
-      <!--
-        Fedora now has com.io7m.xom:xom 1.2.10,
-        so just use the one that defined in parent
-      -->
-     
       <dependency>
         <!-- New group Id -->
         <groupId>com.io7m.xom</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
   <parent>
     <groupId>org.zanata</groupId>
     <artifactId>zanata-parent</artifactId>
-    <version>26-SNAPSHOT</version>
+    <version>27-SNAPSHOT</version>
     <relativePath>../parent</relativePath>
   </parent>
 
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -1342,19 +1342,15 @@
         <version>3.1</version>
       </dependency>
       <!--
-        In the server, we can use a newer version than Fedora has.
-        This one also has a source jar!
+        Fedora now has com.io7m.xom:xom 1.2.10,
+        so just use the one that defined in parent
```
seanf: @definite you say that, but then you specify the version a few lines down anyway. We should be making server less dependent on zanata-parent anyway, so I suggest removing this comment.

seanf:
@definite you can run checkstyle before pushing with this: `mvn checkstyle:check -DstaticAnalysis`

zanata-model/pom.xml
```diff
@@ -320,8 +320,10 @@
     </dependency>
 
     <dependency>
-      <groupId>xom</groupId>
+      <!-- New group Id -->
+      <groupId>com.io7m.xom</groupId>
       <artifactId>xom</artifactId>
+      <version>1.2.10</version>
```
seanf: If you have the version in server/pom.xml's dependencyManagement, you don't need to repeat it here.

zanata-war/pom.xml
```diff
@@ -2394,8 +2394,10 @@
       <version>1.4</version>
     </dependency>
     <dependency>
-      <groupId>xom</groupId>
+      <!-- New group Id -->
+      <groupId>com.io7m.xom</groupId>
       <artifactId>xom</artifactId>
+      <version>1.2.10</version>
```
seanf: If you have the version in server/pom.xml's dependencyManagement, you don't need to repeat it here.

pom.xml
seanf: @definite This comment should be removed.

seanf:

PR #1112, opened 2016-03-11
Merge zanata ui

lukebrooker:
Updated to remove zanata-ui dependency.

Haven't been able to check with a running server yet.


aeng:
Merged into https://github.com/zanata/zanata-server/tree/replace-fixed-data-table, https://github.com/zanata/zanata-server/commit/faddb795c30b0b5271df96264a4fbd8f8ad6d817

PR #1113, opened 2016-03-13
fix(version): delete version

aeng:
https://zanata.atlassian.net/browse/ZNTA-953

### DIFF
--- a/zanata-war/src/main/webapp/WEB-INF/layout/project/languages-list.xhtml
+++ b/zanata-war/src/main/webapp/WEB-INF/layout/project/languages-list.xhtml
@@ -68,16 +68,16 @@
   <!-- Placed at the end of the list content so that it runs when all the
        content from ajax is present in the DOM -->
   <script type="text/javascript">
-  //<![CDATA[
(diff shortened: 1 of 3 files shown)


zanata-war/src/main/webapp/WEB-INF/layout/project/languages-list.xhtml
```diff
@@ -70,7 +70,9 @@
   <script type="text/javascript">
   //<![CDATA[
     // refresh checkbox facades on language lists
-    zanata.form.appendCheckboxes(jQuery('##{id}').element);
+    if (typeof zanata != 'undefined') {
```
carlosmunoz: why would this happen? and is it ok to ignore it?

aeng: `zanata.form.appendCheckboxes` only needed after the first load, which `zanata` will be available by then.

carlosmunoz:
:+1:

djansen-redhat:
:white_check_mark: Tested

PR #1117, opened 2016-03-21
Pr parent 27

definite:


seanf:
@definite If this is only to pick up https://github.com/zanata/zanata-common/pull/23 for use in tests, I don't think we should make such a big change in `zanata-server`'s `release` branch.

definite:
This is no longer relevent

PR #1120, opened 2016-03-23
chore(profile): disable doclint for Java 8 when building with release

definite:
Need from building with  -Prelease and zanata-parent-25


seanf:
:+1:

PR #1121, opened 2016-03-24
fix(npm): update dom-event package in frontend

aeng:


seanf:
:+1: Looks fine, but do we need to add any more packages? 

eg: `concat left-pad map one property right-pad strip subscribe`

which were all reported by this script: https://gist.github.com/cwonrails/962c6503a12821cd161f

aeng:
`left-pad` is fixed. Someone tookover the repo. I managed to build locally with this fix.

seanf:
@aeng Looks like the build succeeded.

definite:
:+1:

PR #1126, opened 2016-04-01
Fix bug where merge-type=import sets state to New

seanf:
https://zanata.atlassian.net/browse/ZNTA-987

### DIFF
--- a/zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
+++ b/zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
@@ -205,6 +205,10 @@ public class HTextFlowTarget extends ModelEntityBase implements HasContents,
         return state;
     }
 
+    public void setState(@Nonnull ContentState newState) {
+        state = newState;
+    }
+
     @NotNull
     @Column(name = "tf_revision")
     @Override
@@ -446,7 +450,7 @@ public class HTextFlowTarget extends ModelEntityBase implements HasContents,
 
     @Transient
     public void clear() {
-        setContents();
+        setContents(null, null, null, null, null, null);
         setState(ContentState.New);
         setComment(null);
         setLastModifiedBy(null);
--- a/zanata-war/src/main/java/org/zanata/service/impl/TranslationMergeFirstTran.java
+++ b/zanata-war/src/main/java/org/zanata/service/impl/TranslationMergeFirstTran.java
@@ -14,6 +14,8 @@ import org.zanata.transformer.TargetTransformer;
 import com.google.common.base.Preconditions;
 
 /**
+ * This strategy is used when HTextFlowTarget does not exist yet, regardless
+ * of the selected merge-type.
  * @author Patrick Huang <a
  *         href="mailto:pahuang@redhat.com">pahuang@redhat.com</a>
  */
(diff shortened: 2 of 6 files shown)


seanf:
@carlosmunoz I think this area deserves a complete overhaul, but not in `release`.

zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
```diff
@@ -446,7 +450,7 @@ public String toString() {
 
     @Transient
     public void clear() {
-        setContents();
+        setContents(null, null, null, null, null, null);
```
carlosmunoz: So, I take it this is what was causing the bug?

zanata-war/src/main/java/org/zanata/service/impl/TranslationMergeFirstTran.java
```diff
@@ -14,6 +14,8 @@
 import com.google.common.base.Preconditions;
 
 /**
+ * This strategy is used when HTextFlowTarget does not exist yet, regardless
+ * of the selected merge-type.
```
carlosmunoz: :+1: for documentation

zanata-war/src/main/java/org/zanata/service/impl/TranslationMergeFirstTran.java
```diff
@@ -34,21 +36,17 @@ public boolean merge(TextFlowTarget incomingTarget,
...
-        targetChanged |=
-                new TargetTransformer(extensions).transform(incomingTarget,
-                        hTarget);
+        new TargetTransformer(extensions).transform(incomingTarget,
+                hTarget);
         if (incomingTarget.getState().isTranslated()) {
             hTarget.setState(ContentState.Translated);
         }
-        return targetChanged;
+        return true;
```
carlosmunoz: Was this always returning `true` and you just made it specific here?

zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java
```diff
@@ -681,8 +681,8 @@ private String validateTranslations(ContentState newState,
         }
 
         // NB: removedTargets only applies for MergeType.IMPORT
-        final Collection<HTextFlowTarget> removedTargets =
-                new HashSet<HTextFlowTarget>();
+        final Collection<Long> removedTextFlowTargetIds =
+                new HashSet<>();
```
carlosmunoz: This should improve our memory usage, if just slightly

seanf:
Oops, forgot to merge the commits into one.

zanata-war/src/main/java/org/zanata/service/impl/TranslationMergeFirstTran.java
seanf: Well, it should have been returning `true`. If not, I just fixed a bug.

zanata-war/src/main/java/org/zanata/service/impl/TranslationMergeFirstTran.java
seanf: :+1: for my own sanity.

zanata-model/src/main/java/org/zanata/model/HTextFlowTarget.java
seanf: No the main bug happened because, when we were checking each HTextFlowTarget against removedTargets, equals was returning false (due to other fields being different). This lead to calling `clear()` but I only discovered that via the debugger, because `clear()` didn't work as intended.

carlosmunoz:
:+1:

zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java
seanf: Yes, from a memory point of view, this wouldn't work nicely with batches either (since it starts out with a complete set of HTextFlowTargets in the collection). In fact, the separate transactions we added for batches might be the reason equality wasn't working.

seanf:
I have squashed the two reviewed commits into https://github.com/zanata/zanata-server/pull/1126/commits/63a69395e19ecf53f9a5d4b525c8926f5309547f.

djansen-redhat:
:white_check_mark: Looks good to me

PR #1132, opened 2016-04-06
fix(merge translation): fix merge translation exception with new translation

aeng:
https://zanata.atlassian.net/browse/ZNTA-1013

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -1,3 +1,7 @@
+## 3.8.4
+<h5>Bug fixes</h5>
+* [ZNTA-1013](https://zanata.atlassian.net/browse/ZNTA-1013) - Fix Merge translation exception
+
 ## 3.8.3
 <h5>Bug fixes</h5>
 * [ZNTA-953](https://zanata.atlassian.net/browse/ZNTA-953) - Fix delete version
--- a/zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java
+++ b/zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java
@@ -21,6 +21,7 @@
 package org.zanata.service.impl;
 
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Future;
 
 import javax.annotation.Nonnull;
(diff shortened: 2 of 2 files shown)


docs/release-notes.md
```diff
@@ -1,6 +1,7 @@
 ## 3.8.3
 <h5>Bug fixes</h5>
 * [ZNTA-953](https://zanata.atlassian.net/browse/ZNTA-953) - Fix delete version
+* [ZNTA-1013](https://zanata.atlassian.net/browse/ZNTA-1013) - Fix Merge translation exception
```
carlosmunoz: would this be 3.8.4?

zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java
```diff
@@ -231,6 +238,27 @@ private Integer mergeTranslations(
...
+                                        targetVersionId,
+                                        targetTf.getDocument().getId(),
+                                        targetTf.getId(),
+                                        updatedTarget.getLocale().getLocaleId(),
+                                        updatedTarget.getId(),
+                                        updatedTarget.getState(),
+                                        entry.getValue());
+
+                        Events.instance().raiseTransactionSuccessEvent(
+                                TextFlowTargetStateEvent.EVENT_NAME, event);
```
carlosmunoz: I was thinking more of raising one big event with the information of all the changed textflow targets. That way we gain some performance improvements.
This is no worse than what's happening at the moment though, which is a single event per updated text flow. But we can improve it.

aeng: That will be more change - new Event type to take in list of informations which I tried to minimise.

carlosmunoz: OK, Lets then at least put a TODO on it so we can refactor in the future.

aeng: Done.

carlosmunoz:
:+1:

djansen-redhat:
:white_check_mark: I *believe* this works as expected

PR #1137, opened 2016-04-11
feat(webhook): event for translation update

aeng:
https://zanata.atlassian.net/browse/ZNTA-938

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1137)


seanf:
@aeng How does this version address the points I raised in https://github.com/zanata/zanata-server/pull/1101#issuecomment-190069230 ? Batching, finding something more efficient than webhooks for frequent events, that sort of thing.

aeng:
2 improvements as of now
- using jms queue for webhook io
- 1 queue per update (we fired single event * webooks previously)

seanf:
> 1 queue per update (we fired single event * webooks previously)

Can you explain that one please? That sounds like a **lot** of JMS queues.

> Todo: batching up event on time based

The main thing is probably to take advantage of the natural batching which already happens when a document is uploaded. 

Each database transaction/batch could trigger an aggregate event, or we could even wait until the entire document is finished, and generate a single aggregate event.

aeng:
> 1 queue per update (we fired single event * webooks previously)
We support multiple webhook url per project(lets say 10 url configured), so total webhook event = (1 * 10). Previously we process 10 of them synchronously and fire the webhook, now we batch them up into 1 single asynchronous event.

> The main thing is probably to take advantage of the natural batching which already happens when a document is uploaded.
The new webhook is for per translation update, document upload is only part of it. We need to listen to update from editor too. Not sure how can we do time based batch here.

zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java
```diff
@@ -0,0 +1,58 @@
...
+@EqualsAndHashCode
+public class DocStatsEvent {
+    private final DocumentLocaleKey key;
+
+    private final Long projectVersionId;
+
+    /**
+     * Updated content states with word counts
+     */
+    private final Map<ContentState, Integer> contentStates;
```
seanf: - [ ] I would be tempted to call this something like wordDeltasByState. The ContentState bit is part of the type, but it's not automatically obvious what the Integers are for.

zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java
```diff
@@ -0,0 +1,58 @@
...
+
+    private final Long lastModifiedTargetId;
+
+    public static Map<ContentState, Integer> updateContentState(
+        @Nullable Map<ContentState, Integer> contentStates, ContentState newState,
+        ContentState previousState, Long longWordCount) {
```
seanf: If we're going to accept Long here, we should have a Map to Long. Given that we've chosen to use Long for HTextFlow.wordCount, we should probably be consistent. Do we need to accept nullable Longs, or could we change it to long?

zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java
```diff
@@ -0,0 +1,58 @@
...
+
+        Map<ContentState, Integer> newContentStates =
+            contentStates == null ? new HashMap<ContentState, Integer>() :
+                contentStates;
+        int wordCount = longWordCount == null ? 0 : longWordCount.intValue();
+
+        Integer previousStateCount = newContentStates.get(previousState);
+        if (previousStateCount == null) {
+            previousStateCount = wordCount * -1;
```
seanf: `previousStateCount = - wordCount` would be a bit cleaner.

zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java
```diff
@@ -0,0 +1,58 @@
...
+    public static Map<ContentState, Integer> updateContentState(
+        @Nullable Map<ContentState, Integer> contentStates, ContentState newState,
+        ContentState previousState, Long longWordCount) {
+
+        Map<ContentState, Integer> newContentStates =
+            contentStates == null ? new HashMap<ContentState, Integer>() :
+                contentStates;
+        int wordCount = longWordCount == null ? 0 : longWordCount.intValue();
+
+        Integer previousStateCount = newContentStates.get(previousState);
```
seanf: Using HashMap.getOrDefault(*, 0) would lead to simpler code, here and further down.

zanata-war/src/main/java/org/zanata/events/webhook/DocumentMilestoneEvent.java
```diff
@@ -18,7 +18,7 @@
  * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA, or see the FSF
  * site: http://www.fsf.org.
  */
-package org.zanata.events;
+package org.zanata.events.webhook;
```
seanf: I think we should keep the webhook payloads in a completely separate package from our CDI events `org.zanata.events`. Maybe `org.zanata.webhook.events`?

zanata-war/src/main/java/org/zanata/events/webhook/DocumentMilestoneEvent.java
```diff
@@ -77,4 +78,8 @@
      */
     private String editorDocumentUrl;
 
+    @Override
+    public String getType() {
+        return this.getClass().getName();
```
seanf: This 'type' property is for JSON, is it? If so, I think we should have a constant, rather than depending on the class name. eg, if we refactor the package names, our JSON serialisation will change.  What about a simple constant "DocumentMilestoneEvent"? Would that do?

zanata-war/src/main/java/org/zanata/events/webhook/DocumentStatsEvent.java
```diff
@@ -0,0 +1,88 @@
...
+    private final LocaleId locale;
+
+    /**
+     * Updated content states with word counts
+     */
+    private final Map<ContentState, Integer> contentStates;
+
+    @Override
+    public String getType() {
+        return this.getClass().getName();
```
seanf: Again, a constant would be safer.

zanata-war/src/main/java/org/zanata/service/TranslationStateCache.java
```diff
@@ -90,4 +91,6 @@ WordStatistic getDocumentStatistics(Long documentId,
      * @param localeId
      */
     void clearDocumentStatistics(Long documentId, LocaleId localeId);
+
+    void docStatsUpdated(DocStatsEvent event);
```
seanf: Is this just declaring one of the `@Observes` event handler methods? I don't really feel comfortable with exposing these event handlers in the interface signature, because the methods are supposed to be invoked by CDI as an implementation detail, not directly by code as part of the public interface. Does this one need to be here?

zanata-war/src/main/java/org/zanata/service/VersionStateCache.java
```diff
@@ -38,7 +38,7 @@
      * locale. (It's really a Text Flow Target state)
```
seanf: This comment seems outdated now.

zanata-war/src/main/java/org/zanata/webhook/events/DocumentMilestoneEvent.java
```diff
@@ -39,9 +40,12 @@
 @Setter
 @AllArgsConstructor
 @NoArgsConstructor
-@JsonPropertyOrder({ "project", "version", "docId", "locale", "editorDocumentUrl", "milestone"})
+@JsonPropertyOrder({"project", "version", "docId", "locale", "editorDocumentUrl", "milestone", "type"})
 @EqualsAndHashCode
 public class DocumentMilestoneEvent extends WebhookEventType {
+
+    private static final String type = "DocumentMilestoneEvent";
```
seanf: - [ ] String constants should be upper case. Perhaps EVENT_TYPE?

zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java
seanf: @aeng The Integer vs Long problem still needs to be cleaned up.

zanata-war/src/main/java/org/zanata/webhook/events/DocumentStatsEvent.java
```diff
@@ -0,0 +1,89 @@
...
+ * @author Alex Eng <a href="mailto:aeng@redhat.com">aeng@redhat.com</a>
+ */
+@Getter
+@Setter
+@AllArgsConstructor
+@JsonPropertyOrder({"user", "project", "version", "docId", "locale", "contentStates", "type"})
+@EqualsAndHashCode
+public class DocumentStatsEvent extends WebhookEventType {
+
+    private static final String type = "DocumentStatsEvent";
```
seanf: - [ ] EVENT_TYPE?

zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java
aeng: @seanf it's a nullable Long from HTextFlow but I don't want to have Null wordcount in the webhook event. Perhaps need to accept Integer in the method. What do you think?

seanf: Alex, HTextFlow.getWordCount() is marked NotNull. I think the value can
only be null during initialisation. Once the HTextFlow has been
initialised, wordCount will not be null.

I would switch it all to Long/long rather than Integer if possible.


On 5 May 2016 at 07:47, Alex Eng <notifications@github.com> wrote:

> In zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java
> <https://github.com/zanata/zanata-server/pull/1137#discussion_r62120533>:
>
> > +@EqualsAndHashCode
> > +public class DocStatsEvent {
> > +    private final DocumentLocaleKey key;
> > +
> > +    private final Long projectVersionId;
> > +
> > +    /**
> > +     * Updated content states with word counts
> > +     */
> > +    private final Map<ContentState, Integer> contentStates;
> > +
> > +    private final Long lastModifiedTargetId;
> > +
> > +    public static Map<ContentState, Integer> updateContentState(
> > +        @Nullable Map<ContentState, Integer> contentStates, ContentState newState,
> > +        ContentState previousState, Long longWordCount) {
>
> @seanf <https://github.com/seanf> it's a nullable Long from HTextFlow but
> I don't want to have Null wordcount in the webhook event. Perhaps need to
> accept Integer in the method. What do you think?
>
> —
> <https://github.com/zanata/zanata-server/pull/1137/files/fdee8c13c56ff20a1ffa399056bb04cb6e19832c#r62120533>
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java
```diff
@@ -25,25 +26,25 @@
...
 
     private final Long lastModifiedTargetId;
 
-    public static Map<ContentState, Integer> updateContentState(
-        @Nullable Map<ContentState, Integer> contentStates, ContentState newState,
-        ContentState previousState, @Nullable Long longWordCount) {
+    public static Map<ContentState, Long> updateContentState(
+            @Nullable Map<ContentState, Long> contentStates,
+            ContentState newState,
+            ContentState previousState, @Nonnull Long wordCount) {
```
seanf: I think you could replace `@Nonnull Long` with `long`.

zanata-war/src/main/java/org/zanata/service/impl/DocumentServiceImpl.java
```diff
@@ -265,8 +265,10 @@ private void processWebHookDocumentMilestoneEvent(
             WordStatistic oldStats = StatisticsUtil.copyWordStatistic(stats);
             if(oldStats != null) {
 
-                for(Map.Entry<ContentState, Integer> entry: event.getWordDeltasByState().entrySet()) {
-                    oldStats.increment(entry.getKey(), entry.getValue());
+                for (Map.Entry<ContentState, Long> entry : event
+                        .getWordDeltasByState().entrySet()) {
+                    oldStats.increment(entry.getKey(),
+                            entry.getValue().intValue());
```
seanf: Perhaps we don't need to push the int->long conversion into AbstractStatistic, but we should have a note about the down-conversion here (calling `intValue()` on a Long).

zanata-war/src/test/java/org/zanata/service/impl/DocumentServiceImplTest.java
```diff
@@ -247,12 +247,12 @@ private void runDocumentStatisticUpdatedTest(
...
         DocumentLocaleKey key = new DocumentLocaleKey(docId, localeId);
-        Map<ContentState, Integer> map = new HashMap<>();
+        Map<ContentState, Long> map = new HashMap<>();
 
-        int oldStatCount = wordCount * -1;
+        Long oldStatCount = wordCount * -1L;
         map.put(oldState, oldStatCount);
 
-        int newStatCount = wordCount + (map.get(newState) != null ? map.get(newState) : 0);
+        Long newStatCount = wordCount + (map.get(newState) != null ? map.get(newState) : 0);
```
seanf: I think newStatCount would be better as a `long`.

zanata-war/src/test/java/org/zanata/service/impl/TranslationUpdatedManagerTest.java
```diff
@@ -100,7 +100,7 @@
     private LocaleId localeId = LocaleId.DE;
     private String versionSlug = "versionSlug";
     private String projectSlug = "projectSlug";
-    private int wordCount = 10;
+    private Long wordCount = 10L;
```
seanf: If you have to call intValue() on it all the time, perhaps wordCount should stay an int (or a long).

zanata-war/src/main/java/org/zanata/service/impl/DocumentServiceImpl.java
aeng: @seanf Not sure what you mean by that.

seanf: I mean, ideally we would use long/Long consistently, instead of downgrading long to int anywhere. But if we're not going to do that, we should at least have a comment making it clear that we are "casting" long to int here.

zanata-war/src/test/java/org/zanata/service/impl/TranslationUpdatedManagerTest.java
seanf: > If you have to call intValue() on it all the time, perhaps wordCount should stay an int (or a long).
 
After all, this is just a test class.

aeng:
@seanf Any updates on this since my last commit?

seanf:
Review status: 0 of 19 files reviewed at latest revision, 16 unresolved discussions.

---

*[docs/release-notes.md, line 10 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHbf6S8DW8Nc7OWlpVI:-KHbf6S9KWUx457nfsgZ:-725579254) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/docs/release-notes.md#L10)):*
> ```Markdown
> ## 3.9
> <h5>New Features</h5>
> * [ZNTA-938](https://zanata.atlassian.net/browse/ZNTA-938) - Webhook event for translation update by user.
> ```

This should appear under 4.0, not 3.9

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

seanf:
Review status: 0 of 19 files reviewed at latest revision, 14 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java, line 42 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHbtQfzcW6P86a1BJpX:-KHbtQfzcW6P86a1BJpY:348386338) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java#L42)):*
> ```Java
>                         : contentStates;
> 
>         Long previousStateCount =
> ```

This should be `long` now, not `Long`.

---

*[zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java, line 47 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHbtXXC2AaW5tb_8P6N:-KHbtXXC2AaW5tb_8P6O:348386338) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java#L47)):*
> ```Java
>         newContentStates.put(previousState, previousStateCount);
> 
>         Long newStateCount = newContentStates.getOrDefault(newState, 0L);
> ```

This should be `long` now, not `Long`.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 199 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHbutAQC7d-IuyZ4Ipd:-KHbutAQC7d-IuyZ4Ipe:-204913143) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L199)):*
> ```Java
> 
>         List<TextFlowTargetState> states = Lists.newArrayList();
>         Map<ContentState, Long> contentStates = Maps.newHashMap();
> ```

I suggest `contentStateDeltas`.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 261 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHbvU1u6LqPCkdy6rBY:-KHbvU1u6LqPCkdy6rBZ:-678159528) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L261)):*
> ```Java
>                                         hTextFlowTarget.getState(),
>                                         currentState);
>                         states.add(state);
> ```

To help distinguish them from contentState, I suggest we should call these variables tftState[s] or targetState[s].

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 263 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHbv6n99RnNvZJ2_qUu:-KHbv6n99RnNvZJ2_qUv:1279558911) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L263)):*
> ```Java
>                         states.add(state);
>                         contentStates = DocStatsEvent
>                             .updateContentState(contentStates,
> ```

I suggest something like `updateContentStateDeltas`.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 806 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHbvwGpNgCcSOoZWU4u:-KHbvwGqZeuCfm0HNCub:51183183) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L806)):*
> ```Java
>                 ImmutableList.builder();
> 
>         Map<ContentState, Long> contentStates = Maps.newHashMap();
> ```

I suggest `contentStateDeltas` again. This is looking a bit familiar...

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 900 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHbw2gc_vMY191rGa_-:-KHbw2gc_vMY191rGa_0:339013139) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L900)):*
> ```Java
>                             hTarget.getId(), hTarget.getState(), currentState);
> 
>                     states.add(state);
> ```

Again tftState[s] or similar. Very familiar...  There must be some way we can share some of this code (in `saveBatch`) with `translate`.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 911 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHbwe-zFCubo4wA3Aq9:-KHbwe-zFCubo4wA3AqA:185764188) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L911)):*
> ```Java
>             }
>         }
>         ImmutableList<TextFlowTargetState> list = states.build();
> ```

How about `list` -> `tftStates`?

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationStateCacheImpl.java, line 207 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHbxBrcXTgLTg97Sc9V:-KHbxBrcXTgLTg97Sc9W:497839758) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationStateCacheImpl.java#L207)):*
> ```Java
>     }
> 
>     @Async
> ```

Why are these two event listeners Async now?

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java, line 72 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHc1iNYKxOjry5dZ6cF:-KHc1iNZxKYjHyHRJDKD:-996843409) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java#L72)):*
> ```Java
>     }
> 
>     void processWebHookEvent(DocStatsEvent event) {
> ```

Should this be private, or VisibleForTesting?

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java, line 89 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHc-hUmnvp1oORUdlMP:-KHc-hUn5BWQYSeUGjHM:1793348268) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java#L89)):*
> ```Java
>         LocaleId localeId = event.getKey().getLocaleId();
> 
>         User user = userService.transferToUser(person.getAccount(),
> ```

I know it's existing code, but I had to load the source code in my IDE to work out what `userService.transferToUser(HAccount)` actually does. It needs a more descriptive name, maybe `getUserInfo(HAccount)`. 

If you don't feel like renaming it now (the method has 8 references BTW), fair enough, but perhaps you could add a "TODO rename to getUserInfo(HAccount)" or similar (in UserService).

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java, line 99 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHc1pgkVDOqDErLM0vd:-KHc1pgkVDOqDErLM0ve:-1844688748) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java#L99)):*
> ```Java
>     }
> 
>     public void publishWebhookEvent(List<WebHook> webHooks,
> ```

Does this need to be public?

---

*[zanata-war/src/main/java/org/zanata/webhook/events/DocumentStatsEvent.java, line 83 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHc3eAio14k1GLpiiw-:-KHc3eAjkRFjvvYFp2IM:174728685) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/webhook/events/DocumentStatsEvent.java#L83)):*
> ```Java
>      * Updated content states with word counts
>      */
>     private final Map<ContentState, Long> contentStates;
> ```

This needs a more descriptive name, perhaps wordDeltasByState? Now that I think about it, contentStateDeltas could be ambiguous. 

We could also include stringDeltasByState (or msgDeltasByState), but perhaps we should stick to word counts for now. Our definition of a string or message might change when we implement server-side segmentation. Right @carlosmunoz ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

seanf:
Reviewed 5 of 22 files at r1, 6 of 13 files at r2, 1 of 2 files at r3, 4 of 9 files at r4, 3 of 3 files at r5, 3 of 3 files at r6.
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/service/impl/VersionStateCacheImpl.java, line 104 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHc5n11spN1lNZPK3wP:-KHc5n11spN1lNZPK3wQ:128406825) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/VersionStateCacheImpl.java#L104)):*
> ```Java
>             for (Map.Entry<ContentState, Long> entry : event
>                     .getWordDeltasByState().entrySet()) {
>                 stats.increment(entry.getKey(), entry.getValue().intValue());
> ```

I think you should be able to use `longValue()` here, which should be safer.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

seanf:
Reviewed 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

aeng:
Review status: 13 of 23 files reviewed at latest revision, 14 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java, line 99 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHc1pgkVDOqDErLM0vd:-KHsk2cm9GnXVwrOwmU6:1946805374) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java#L99)):*
> Does this need to be public?</details>
@VisibleForTesting

---

*[zanata-war/src/main/java/org/zanata/service/impl/VersionStateCacheImpl.java, line 104 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHc5n11spN1lNZPK3wP:-KHskMAq-cdfe4BGg7xs:1672807256) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/VersionStateCacheImpl.java#L104)):*
> I think you should be able to use `longValue()` here, which should be safer.</details>
AbstractStatistic accepts only int. I think this refactor should be a separate PR.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

seanf:
Reviewed 1 of 13 files at r2, 12 of 12 files at r8.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 900 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHbw2gc_vMY191rGa_-:-KHwLCV8qS6AhiuD9Ujd:16183136) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L900)):*
> Again tftState[s] or similar. Very familiar...  There must be some way we can share some of this code (in `saveBatch`) with `translate`.</details>
IntelliJ's Analyse/Locate Duplicates shows that this section (from `TextFlowTargetState state = ...` down to `contentStateDeltas = DocStatsEvent.updateContentStateDeltas...` is duplicated in one of the `translate` methods. 

The duplicate code should be extracted to another method, eg `aggregateChanges(textFlow, tfTarget, targetStates, contentStateDeltas)`.

Also, for the sake of performance, I think the method `DocStatsEvent.updateContentStateDeltas` should mutate the supplied `contentStateDeltas` map (don't make it `@Nullable`), instead of copying it and returning a new one. That will also help with the `aggregateChanges` method.

---

*[zanata-war/src/main/java/org/zanata/service/impl/VersionStateCacheImpl.java, line 104 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHc5n11spN1lNZPK3wP:-KHwAXohiozkwWVQ6i8v:-1520566393) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/VersionStateCacheImpl.java#L104)):*
<details><summary><i>Previously, aeng (Alex Eng) wrote…</i></summary>
> AbstractStatistic accepts only int. I think this refactor should be a separate PR.</details>
Oh right, I was looking at remainingHours (double), not the ints in the superclass. 

The Java 8 method `Math.toIntExact` looks like a good way of handling cases like this:  http://stackoverflow.com/a/32869210/14379

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

aeng:
Review status: 20 of 23 files reviewed at latest revision, 3 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 900 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHbw2gc_vMY191rGa_-:-KI0H9SbI8X2ORyCJWD2:-719267080) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L900)):*
> IntelliJ's Analyse/Locate Duplicates shows that this section (from `TextFlowTargetState state = ...` down to `contentStateDeltas = DocStatsEvent.updateContentStateDeltas...` is duplicated in one of the `translate` methods. 
> 
> The duplicate code should be extracted to another method, eg `aggregateChanges(textFlow, tfTarget, targetStates, contentStateDeltas)`.
> 
> Also, for the sake of performance, I think the method `DocStatsEvent.updateContentStateDeltas` should mutate the supplied `contentStateDeltas` map (don't make it `@Nullable`), instead of copying it and returning a new one. That will also help with the `aggregateChanges` method.
</details>
Updated.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java, line 72 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHc1iNYKxOjry5dZ6cF:-KI0Gq_CBF7Na_5E973T:-502311804) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java#L72)):*
> Should this be private, or VisibleForTesting?
</details>
Should be visibleForTesting

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

aeng:
Review status: 20 of 23 files reviewed at latest revision, 2 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/service/impl/VersionStateCacheImpl.java, line 104 \[r6\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KHc5n11spN1lNZPK3wP:-KI0H67QyAAG6kmXrCox:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/bee9298bc205d265b8c58e1cdd66a26db809969b/zanata-war/src/main/java/org/zanata/service/impl/VersionStateCacheImpl.java#L104)):*
> Oh right, I was looking at remainingHours (double), not the ints in the superclass. 
> 
> The Java 8 method `Math.toIntExact` looks like a good way of handling cases like this:  http://stackoverflow.com/a/32869210/14379
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

seanf:
Reviewed 3 of 3 files at r9.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java, line 42 \[r9\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI0mFsgZEV7meAOg5IJ:-KI0mFsgZEV7meAOg5IK:-517175082) ([raw file](https://github.com/zanata/zanata-server/blob/3d5125192a0857aefe1494e76b7051bab3e4f057/zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java#L42)):*
> ```Java
>         Map<ContentState, Long> newContentStates =
>                 contentStates == null ? Maps.newHashMap()
>                         : Maps.newHashMap(contentStates);
> ```

Isn't contentStates Nonnull now? Anyway, I think you could just operate on contentStates, instead of copying to a new HashMap.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 931 \[r9\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI0maqkLB5PkX3hmhVF:-KI0maqlA8aMwjqsM-6s:1265767113) ([raw file](https://github.com/zanata/zanata-server/blob/3d5125192a0857aefe1494e76b7051bab3e4f057/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L931)):*
> ```Java
>                 .updateContentStateDeltas(contentStateDeltas,
>                         state.getNewState(),
>                         state.getPreviousState(), textFlow.getWordCount());
> ```

Again, I think we just want to modify the existing map, which means we don't need to return a value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

aeng:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java, line 42 \[r9\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI0mFsgZEV7meAOg5IJ:-KI0vPw0iJ3Bw2Ouyf54:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/3d5125192a0857aefe1494e76b7051bab3e4f057/zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java#L42)):*
> Isn't contentStates Nonnull now? Anyway, I think you could just operate on contentStates, instead of copying to a new HashMap.
</details>
Done.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 931 \[r9\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI0maqkLB5PkX3hmhVF:-KI0vNTuC1VEMXmDwgt0:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/3d5125192a0857aefe1494e76b7051bab3e4f057/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L931)):*
> Again, I think we just want to modify the existing map, which means we don't need to return a value.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

seanf:
Reviewed 4 of 4 files at r10.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

djansen-redhat:
The logging of webhook firing (line 70 of WebHooksPublisher.java) should be debug, not info, else there will be a log flood every time someone uploads a .po file

aeng:
@djansen-redhat good point.

seanf:
Some of @djansen-redhat 's questions made me realise that I need to do a bit more high-level review, just to double check the performance implications.

seanf:
Reviewed 1 of 1 files at r11.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java, line 36 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6XSTSHpjL80B7qW3F:-KI6XSTSHpjL80B7qW3G:817439733) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java#L36)):*
> ```Java
> 
>     public static void updateContentStateDeltas(
>             @Nonnull Map<ContentState, Long> contentStates,
> ```

We should probably name this param wordDeltasByState for consistency.

---

*[zanata-war/src/main/java/org/zanata/service/impl/DocumentServiceImpl.java, line 271 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6YXXrB9dNr46zTVvH:-KI6YXXsnxmxeZeYUVIO:-1443056374) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/DocumentServiceImpl.java#L271)):*
> ```Java
>                         .getWordDeltasByState().entrySet()) {
>                     //casting Long to int for statistics calculation
>                     int count = entry.getValue().intValue();
> ```

We should use Math.toIntExact here (instead of the comment).

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 841 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6ZbJdlDKdDK2u9Z4z:-KI6ZbJdlDKdDK2u9Z5-:1464488292) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L841)):*
> ```Java
>                 // we have eagerly loaded all targets upfront
>                 HTextFlowTarget hTarget = textFlow.getTargets().get(locale.getId());
>                 ContentState currentState = ContentState.New;
> ```

We should call this oldState.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 919 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6ZezbeyrRr18akbi-:-KI6ZezbeyrRr18akbi0:142020604) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L919)):*
> ```Java
> 
>     private void aggregateChanges(HTextFlow textFlow, HTextFlowTarget hTarget,
>             ContentState currentState, List<TextFlowTargetState> targetStates,
> ```

We should call currentState oldState instead.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 923 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6_fudlG7kE4hKDfnZ:-KI6_fudlG7kE4hKDfn_:1700570662) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L923)):*
> ```Java
>         TextFlowTargetState state =
>                 new TextFlowTargetState(textFlow.getId(),
>                         hTarget.getId(), hTarget.getState(), currentState);
> ```

We might want to change TextFlowTargetState to TextFlowTargetStateChange, since that's what it is really capturing.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java, line 74 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6aeJGkPyWNQ2S42-D:-KI6aeJGkPyWNQ2S42-E:-1032112630) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java#L74)):*
> ```Java
>     protected void processWebHookEvent(DocStatsEvent event) {
>         HTextFlowTarget target =
>                 textFlowTargetDAO.findById(event.getLastModifiedTargetId());
> ```

Just for a bit more efficiency, we should move this line below the return statements, since it isn't always needed.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java, line 81 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6anyn_MAQ40lIucNp:-KI6anyn_MAQ40lIucNq:1273445521) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java#L81)):*
> ```Java
>         HDocument document = documentDAO.findById(event.getKey().getDocumentId());
>         String docId = document.getDocId();
>         String versionSlug = document.getProjectIteration().getSlug();
> ```

We should also move this line below the return statement.

---

*[zanata-war/src/main/java/org/zanata/service/impl/VersionStateCacheImpl.java, line 106 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6bE7yKvdf5v67wD4D:-KI6bE7yKvdf5v67wD4E:1018714821) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/VersionStateCacheImpl.java#L106)):*
> ```Java
>                 stats.increment(entry.getKey(),
>                         Math.toIntExact(entry.getValue()));
>             }
> ```

You might like to use Java 8 streams for this sort of thing in future.

---

*[zanata-war/src/main/java/org/zanata/service/impl/WebHooksPublisher.java, line 70 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6bXyPSZa-pFmJxsP3:-KI6bXyPSZa-pFmJxsP4:-267517597) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/WebHooksPublisher.java#L70)):*
> ```Java
>                 postBuilder.header(WEBHOOK_HEADER, sha);
>             }
>             log.debug("firing webhook: {}:{}", callbackURL, data);
> ```

It's probably worth saying async in the log message, even if it is just a debug now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

aeng:
Review status: 16 of 26 files reviewed at latest revision, 8 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java, line 74 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6aeJGkPyWNQ2S42-D:-KI6dpqUIY198yCFwHoA:2131239401) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java#L74)):*
> Just for a bit more efficiency, we should move this line below the return statements, since it isn't always needed.
</details>
target is needed in next line.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

seanf:
Regarding the performance implications, I think the main points are:

1. The batch size: the bigger it is, the fewer events will need to be fired. Perhaps we should put something about this in the docs and/or release notes.
2. Some sysadmins may wish to disable the TFT-level webhooks for performance reasons. It might be worth having a global off switch.

Otherwise, everything looks good. 

> Some of @djansen-redhat 's questions made me realise that I need to do a bit more high-level review, just to double check the performance implications.
</details>


---

Reviewed 10 of 10 files at r12.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java, line 74 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6aeJGkPyWNQ2S42-D:-KI6fou4oUPlAD2Dwa4l:335332973) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java#L74)):*
<details><summary><i>Previously, aeng (Alex Eng) wrote…</i></summary>
> target is needed in next line.
</details>
Oops, sorry.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-:-KI6g3Z4qHwc-_05nbE_:124263232)*

seanf:
The javadocs are good, but I was thinking of user-visible docs - something sys admins might see. But I forgot that these numbers are currently hard-coded, which means sys admins can't use them to manage system performance.  Now that the batch size is more important, we might need to make it configurable via system property.

I just realised that all projects which have enabled webhooks will now generate DocStatsEvents as well as DocumentMilestoneEvent - this is too much. 

Only some projects will care about this level of detail, so we should let project maintainers control which events they want to receive (and DocStatsEvents should be off by default). Then if the DocStatsEvent is not enabled, we can skip quite a bit of processing.

> Regarding the performance implications, I think the main points are:
> 
> 1. The batch size: the bigger it is, the fewer events will need to be fired. Perhaps we should put something about this in the docs and/or release notes.
> 2. Some sysadmins may wish to disable the TFT-level webhooks for performance reasons. It might be worth having a global off switch.
> 
> Otherwise, everything looks good. 
> 
</details>


---

Reviewed 2 of 2 files at r13.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java, line 116 \[r13\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6qmIVUHlmGW_yIwBd:-KI6qmIWNYx-Q1lNf3RY:425288361) ([raw file](https://github.com/zanata/zanata-server/blob/f5362e637169e94cda3558c5e7d2d63824fa58b4/zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java#L116)):*
> ```Java
>      * related translated states.
>      */
>     private final static int TRANSLATION_BATCH_SIZE = 10;
> ```

@aeng Do you think we should bump up this batch size? Is there a reason it is only 10? Or perhaps we should simply be using the same constant here and in TranslationService.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-:-KI6rEGmc_iyFaDgeHc7:1757095233)*

seanf:
@aeng Let me know when you want me to review again.

aeng:
I've added webhook Type that will attached to each webhook url. So project maintainer can set the type of webhook they want in the url.

> @aeng Let me know when you want me to review again.
</details>


---

Review status: 20 of 37 files reviewed at latest revision, 7 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java, line 116 \[r13\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6qmIVUHlmGW_yIwBd:-KIBFaug6U7HaoRIQKe-:-1359265462) ([raw file](https://github.com/zanata/zanata-server/blob/f5362e637169e94cda3558c5e7d2d63824fa58b4/zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java#L116)):*
> @aeng Do you think we should bump up this batch size? Is there a reason it is only 10? Or perhaps we should simply be using the same constant here and in TranslationService.
</details>
The batch size for this is different from TranslationService. This is batch size to search matching TextFlow, and each textflow might have 10/20/100 target to process. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-:-KIBFnZksfOjpXyMg86x:1912303621)*

aeng:
Review status: 20 of 37 files reviewed at latest revision, 7 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java, line 36 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6XSTSHpjL80B7qW3F:-KIBHH4C46x73Kc2KIle:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/events/DocStatsEvent.java#L36)):*
> We should probably name this param wordDeltasByState for consistency.
</details>
Done.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 841 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6ZbJdlDKdDK2u9Z4z:-KIBHIZRd5lwY09E0VpS:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L841)):*
> We should call this oldState.
</details>
Done.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 919 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6ZezbeyrRr18akbi-:-KIBHJFONyEJB65r45kC:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L919)):*
> We should call currentState oldState instead.
</details>
Done.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 923 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6_fudlG7kE4hKDfnZ:-KIBHJv_3-6n_KpPXRss:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L923)):*
> We might want to change TextFlowTargetState to TextFlowTargetStateChange, since that's what it is really capturing.
</details>
Done.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java, line 81 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6anyn_MAQ40lIucNp:-KIBHKlNH8mJgaMZrttm:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java#L81)):*
> We should also move this line below the return statement.
</details>
Done.

---

*[zanata-war/src/main/java/org/zanata/service/impl/WebHooksPublisher.java, line 70 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6bXyPSZa-pFmJxsP3:-KIBHLyhYBGLIc7bLFsi:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/WebHooksPublisher.java#L70)):*
> It's probably worth saying async in the log message, even if it is just a debug now.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

seanf:
Reviewed 1 of 22 files at r1, 2 of 13 files at r2, 8 of 12 files at r14, 7 of 7 files at r15, 2 of 2 files at r16.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[zanata-model/src/main/java/org/zanata/model/type/WebhookType.java, line 21 \[r16\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KIBv8H6KynGDmAZtAwj:-KIBv8H6KynGDmAZtAwk:121697912) ([raw file](https://github.com/zanata/zanata-server/blob/ba11356b24394edcdb22131646f5b19dfce6babb/zanata-model/src/main/java/org/zanata/model/type/WebhookType.java#L21)):*
> ```Java
>     }
> 
>     public static WebhookType getValueOf(String abbr) {
> ```

Perhaps it would be easier (less maintenance) to name the enum values DOC_MILESTONE and DOC_STATS, so that we can just use the auto-generated valueOf method. 

Alternatively, we could just use the enum names (DocumentMilestoneEvent and DocumentStatsEvent) in the database, and not bother with the abbreviations.

---

*[zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java, line 116 \[r13\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6qmIVUHlmGW_yIwBd:-KIBw_wnwgGwMqomY69L:-1377296411) ([raw file](https://github.com/zanata/zanata-server/blob/f5362e637169e94cda3558c5e7d2d63824fa58b4/zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java#L116)):*
<details><summary><i>Previously, aeng (Alex Eng) wrote…</i></summary>
> The batch size for this is different from TranslationService. This is batch size to search matching TextFlow, and each textflow might have 10/20/100 target to process. 
</details>
Okay, thanks for the explanation. Just for clarity, perhaps we should rename it, eg TEXTFLOWS_PER_BATCH?

---

*[zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java, line 110 \[r16\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KIBwAfr_eIwsQ1wgbiO:-KIBwAfr_eIwsQ1wgbiP:-1464495027) ([raw file](https://github.com/zanata/zanata-server/blob/ba11356b24394edcdb22131646f5b19dfce6babb/zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java#L110)):*
> ```Java
> 
>     /**
>      * Batch size for find matching HTextFlow to process merging of translations.
> ```

:+1:  Perhaps add this too? "(Each TextFlow may lead to changes in multiple TextFlowTargets (up to one per locale)"

---

*[zanata-war/src/main/resources/messages.properties, line 318 \[r16\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KIBxDMbPSn7AaaXy8_z:-KIBxDMbPSn7AaaXy8a-:143673790) ([raw file](https://github.com/zanata/zanata-server/blob/ba11356b24394edcdb22131646f5b19dfce6babb/zanata-war/src/main/resources/messages.properties#L318)):*
> ```INI
> jsf.webhook.response.state={0}% {1}
> jsf.webhookType.DocumentMilestoneEvent.desc=Trigger when a document is 100% translated or approved
> jsf.webhookType.DocumentStatsEvent.desc=Trigger when translations are updated
> ```

How about something like "Trigger when translations are updated (singly or in a batch)" 

(Just to make it clearer that Zanata tries not to send an event for every single translation.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

aeng:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[zanata-model/src/main/java/org/zanata/model/type/WebhookType.java, line 21 \[r16\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KIBv8H6KynGDmAZtAwj:-KIQrmRJdBs2i62747NO:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/ba11356b24394edcdb22131646f5b19dfce6babb/zanata-model/src/main/java/org/zanata/model/type/WebhookType.java#L21)):*
> Perhaps it would be easier (less maintenance) to name the enum values DOC_MILESTONE and DOC_STATS, so that we can just use the auto-generated valueOf method. 
> 
> Alternatively, we could just use the enum names (DocumentMilestoneEvent and DocumentStatsEvent) in the database, and not bother with the abbreviations.
</details>
Done.

---

*[zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java, line 116 \[r13\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6qmIVUHlmGW_yIwBd:-KIQrosdf6KcyQR0Sn2t:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/f5362e637169e94cda3558c5e7d2d63824fa58b4/zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java#L116)):*
> Okay, thanks for the explanation. Just for clarity, perhaps we should rename it, eg TEXTFLOWS_PER_BATCH?
</details>
Done.

---

*[zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java, line 110 \[r16\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KIBwAfr_eIwsQ1wgbiO:-KIQrnlLo4JDBSgJ0GW4:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/ba11356b24394edcdb22131646f5b19dfce6babb/zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java#L110)):*
> :+1:  Perhaps add this too? "(Each TextFlow may lead to changes in multiple TextFlowTargets (up to one per locale)"
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

seanf:
:lgtm:

<details><summary><i>Previously, aeng (Alex Eng) wrote…</i></summary>
> I've added webhook Type that will attached to each webhook url. So project maintainer can set the type of webhook they want in the url.
</details>


---

Reviewed 7 of 7 files at r17.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-:-KIQycdjVAt0WuEjMwHf:1417088721)*

djansen-redhat:
Test:
- Options
- [ ] Toggle webhook for milestone
- [ ] Toggle webhook for translation
- [ ] Both are fired at 100% translated, when both are enabled
- Translation merge batches at 10, so
- [ ] Document with < 10 textflows fires webhook
- ...

djansen-redhat:
*[zanata-war/src/main/java/org/zanata/service/TranslationService.java, line 41 \[r18\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KIV4IMMh5Q0kqL6scFG:-KIV4IMNJ5UYFMcfasYJ:-236794020) ([raw file](https://github.com/zanata/zanata-server/blob/d1b29b987b72ea7da6aa5405c8b3aca7f6042a9c/zanata-war/src/main/java/org/zanata/service/TranslationService.java#L41)):*
> ```Java
>      * Batch size processing for large file upload
>      *
>      * This will determine how many DocStatsEvent will be trigger as part of webhook event.
> ```

I'm a bit confused by this comment, it effectively means how many are collected before firing, yes?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

aeng:
Review status: 29 of 37 files reviewed at latest revision, 4 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/service/TranslationService.java, line 41 \[r18\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KIV4IMMh5Q0kqL6scFG:-KIV6tAKO-lddlCXherd:160394146) ([raw file](https://github.com/zanata/zanata-server/blob/d1b29b987b72ea7da6aa5405c8b3aca7f6042a9c/zanata-war/src/main/java/org/zanata/service/TranslationService.java#L41)):*
<details><summary><i>Previously, djansen-redhat (Damian Jansen) wrote…</i></summary>
> I'm a bit confused by this comment, it effectively means how many are collected before firing, yes?
</details>
Yes

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137)*

seanf:
:lgtm:

<details><summary><i>Previously, djansen-redhat (Damian Jansen) wrote…</i></summary>
> Test:
> 
> 
> 
> Options
> 
> - [ ] Create/remove webhook for milestone
> 
> - [ ] Create/remove webhook for translation
> 
> - [ ] Both are fired at 100% translated, when both exist
> 
> 
> 
> Translation merge batches at 10, so
> 
> - [ ] Document with < 10 textflows still fires webhook
> 
> - [ ] Many textflows batches
> 
> 
> 
> Translation upload batches at 100, so
> 
> - [ ] Check translation of 99 and 199 items (1 and 2 webhooks respectively)
> 
> - [ ] Performance check of large upload
</details>


---

Reviewed 1 of 13 files at r2, 4 of 4 files at r18, 4 of 4 files at r19.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 841 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6ZbJdlDKdDK2u9Z4z:-KIW6rT1TyuLyJ3ktGBj:5173665) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L841)):*
<details><summary><i>Previously, aeng (Alex Eng) wrote…</i></summary>
> Done.
</details>
OK

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 919 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6ZezbeyrRr18akbi-:-KIW6wy-iktzg7BgtcLV:5174689) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L919)):*
<details><summary><i>Previously, aeng (Alex Eng) wrote…</i></summary>
> Done.
</details>
ok

---

*[zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java, line 923 \[r11\]](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-KI6_fudlG7kE4hKDfnZ:-KIW6zREWUmM_CyHAHTp:5174689) ([raw file](https://github.com/zanata/zanata-server/blob/cc16716756e7edd9ef9e9cc30a39b17a50c4b391/zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java#L923)):*
<details><summary><i>Previously, aeng (Alex Eng) wrote…</i></summary>
> Done.
</details>
ok

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1137#-:-KIW7LVAwqMMEDlzKXTp:1417088721)*

definite:

djansen-redhat:
:white_check_mark: Works well. Tested with multiple and varying hook targets

PR #1138, opened 2016-04-11
refactor(ZNTA-1017): translation updates event to handle batch

aeng:
https://zanata.atlassian.net/browse/ZNTA-1017

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -31,8 +31,6 @@
 * [ZNTA-412](https://zanata.atlassian.net/browse/ZNTA-412) - Profile link to project maintainers, language members, and version group maintainers
 * [ZNTA-905](https://zanata.atlassian.net/browse/ZNTA-905) - Remove 0% matching translation memory entry
 * [ZNTA-928](https://zanata.atlassian.net/browse/ZNTA-928) - Readonly project doesn't have "lock" icon in UI
-* [ZNTA-793](https://zanata.atlassian.net/browse/ZNTA-793) - Remove unused method in TranslationMemoryAction
-* [ZNTA-1017](https://zanata.atlassian.net/browse/ZNTA-1017) - Refactor translation update event 
 * [ZNTA-54](https://zanata.atlassian.net/browse/ZNTA-54) - Allow delete language
 
 <h5>Infrastructure Changes</h5>
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -32,6 +32,7 @@
 * [ZNTA-905](https://zanata.atlassian.net/browse/ZNTA-905) - Remove 0% matching translation memory entry
 * [ZNTA-928](https://zanata.atlassian.net/browse/ZNTA-928) - Readonly project doesn't have "lock" icon in UI
 * [ZNTA-793](https://zanata.atlassian.net/browse/ZNTA-793) - Remove unused method in TranslationMemoryAction
+* [ZNTA-1017](https://zanata.atlassian.net/browse/ZNTA-1017) - Refactor translation update event 
 
 <h5>Infrastructure Changes</h5>
 * Recommended platform: JBoss EAP 6 (6.4.6.GA or later).
--- a/zanata-war/src/main/java/org/zanata/events/DocumentLocaleKey.java
+++ b/zanata-war/src/main/java/org/zanata/events/DocumentLocaleKey.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2016, Red Hat, Inc. and individual contributors as indicated by the
+ * @author tags. See the copyright.txt file in the distribution for a full
+ * listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation; either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * This software is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this software; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA, or see the FSF
+ * site: http://www.fsf.org.
+ */
+
+package org.zanata.events;
+
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import org.zanata.common.LocaleId;
+
+import javax.annotation.Nullable;
+import java.io.Serializable;
+
+/**
+ * @author Alex Eng<a href="mailto:aeng@redhat.com">aeng@redhat.com</a>
+ */
+@AllArgsConstructor
+@EqualsAndHashCode
+@Getter
+public class DocumentLocaleKey implements Serializable{
+
+    // this may be null in the case of document uploads
+    private final @Nullable Long actorId;
+    private final Long projectIterationId;
+    private final Long documentId;
+    private final LocaleId localeId;
+}
--- a/zanata-war/src/main/java/org/zanata/events/DocumentLocaleKey.java
+++ b/zanata-war/src/main/java/org/zanata/events/DocumentLocaleKey.java
@@ -36,7 +36,6 @@ import java.io.Serializable;
 @Getter
 public class DocumentLocaleKey implements Serializable {
 
-    private final Long projectIterationId;
     private final Long documentId;
     private final LocaleId localeId;
 }
--- a/zanata-war/src/main/java/org/zanata/events/DocumentLocaleKey.java
+++ b/zanata-war/src/main/java/org/zanata/events/DocumentLocaleKey.java
@@ -26,7 +26,6 @@ import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import org.zanata.common.LocaleId;
 
-import javax.annotation.Nullable;
 import java.io.Serializable;
 
 /**
@@ -35,10 +34,8 @@ import java.io.Serializable;
 @AllArgsConstructor
 @EqualsAndHashCode
 @Getter
-public class DocumentLocaleKey implements Serializable{
+public class DocumentLocaleKey implements Serializable {
 
-    // this may be null in the case of document uploads
-    private final @Nullable Long actorId;
     private final Long projectIterationId;
     private final Long documentId;
     private final LocaleId localeId;
--- a/zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
+++ b/zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
@@ -195,9 +195,9 @@ public class ActivityServiceImpl implements ActivityService {
                 HDocument document =
                     documentDAO.getById(event.getKey().getDocumentId());
 
-                for (TextFlowTargetStateEvent.TextFlowTargetState state : event
-                    .getStates()) {
-                    transactionUtil.run(() -> {
+                transactionUtil.run(() -> {
+                    for (TextFlowTargetStateEvent.TextFlowTargetState state : event
+                        .getStates()) {
                         HTextFlowTarget target =
                             textFlowTargetDAO.findById(
                                 state.getTextFlowTargetId(), false);
@@ -211,8 +211,9 @@ public class ActivityServiceImpl implements ActivityService {
                             activityType,
                             target.getTextFlow().getWordCount()
                                 .intValue());
-                    });
-                }
+                    }
+                });
+
             } catch (Exception e) {
                 Throwables.propagate(e);
             } finally {
--- a/zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
+++ b/zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
@@ -186,16 +186,16 @@ public class ActivityServiceImpl implements ActivityService {
         // workaround for https://issues.jboss.org/browse/WELD-2019
         final TextFlowTargetStateEvent event = event_;
 
-        Long actorId = event.getKey().getActorId();
+        Long actorId = event.getActorId();
 
         if (actorId != null) {
             Lock lock = activityLockManager.getLock(actorId);
             lock.lock();
             try {
-                HDocument document =
-                    documentDAO.getById(event.getKey().getDocumentId());
-
                 transactionUtil.run(() -> {
+                    HDocument document =
+                        documentDAO.getById(event.getKey().getDocumentId());
+
                     for (TextFlowTargetStateEvent.TextFlowTargetState state : event
                         .getStates()) {
                         HTextFlowTarget target =
--- a/zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
+++ b/zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
@@ -201,8 +201,8 @@ public class ActivityServiceImpl implements ActivityService {
                     HTextFlowTarget lastReviewedTarget = null;
                     HTextFlowTarget lastTranslatedTarget = null;
 
-                    int totalReviewedWord = 0;
-                    int totalTranslatedWord = 0;
+                    int totalReviewedWords = 0;
+                    int totalTranslatedWords = 0;
 
                     for (TextFlowTargetStateEvent.TextFlowTargetState state : event
                         .getStates()) {
@@ -211,12 +211,12 @@ public class ActivityServiceImpl implements ActivityService {
                                 state.getTextFlowTargetId(), false);
                         if (state.getNewState().isReviewed()) {
                             lastReviewedTarget = target;
-                            totalReviewedWord +=
+                            totalReviewedWords +=
                                 target.getTextFlow().getWordCount()
                                     .intValue();
                         } else {
                             lastTranslatedTarget = target;
-                            totalTranslatedWord +=
+                            totalTranslatedWords +=
                                 target.getTextFlow().getWordCount()
                                     .intValue();
                         }
@@ -225,7 +225,7 @@ public class ActivityServiceImpl implements ActivityService {
                         logActivityAlreadyLocked(actorId,
                             document.getProjectIteration(), lastReviewedTarget,
                             ActivityType.REVIEWED_TRANSLATION,
-                            totalReviewedWord);
+                            totalReviewedWords);
                     }
 
                     if (lastTranslatedTarget != null) {
@@ -233,10 +233,9 @@ public class ActivityServiceImpl implements ActivityService {
                             document.getProjectIteration(),
                             lastTranslatedTarget,
                             ActivityType.UPDATE_TRANSLATION,
-                            totalTranslatedWord);
+                            totalTranslatedWords);
                     }
                 });
-
             } catch (Exception e) {
                 Throwables.propagate(e);
             } finally {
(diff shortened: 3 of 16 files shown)


aeng:

seanf:
@aeng There was a NullPointerException in TranslationUpdatedManagerTest.onTranslationUpdateTest

zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
```diff
@@ -184,24 +185,34 @@ public Object getEntity(EntityType entityType, long entityId) {
...
-                    logActivityAlreadyLocked(actorId,
-                            document.getProjectIteration(), target, activityType,
-                            target.getTextFlow().getWordCount().intValue());
-                });
+                HDocument document =
+                    documentDAO.getById(event.getKey().getDocumentId());
+
+                for (TextFlowTargetStateEvent.TextFlowTargetState state : event
+                    .getStates()) {
+                    transactionUtil.run(() -> {
```
seanf: What would happen if we put this transaction around the for loop, instead of inside it?

zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
```diff
@@ -184,24 +185,34 @@ public Object getEntity(EntityType entityType, long entityId) {
...
+                        HTextFlowTarget target =
+                            textFlowTargetDAO.findById(
+                                state.getTextFlowTargetId(), false);
+                        ActivityType activityType =
+                            state.getNewState().isReviewed()
+                                ? ActivityType.REVIEWED_TRANSLATION
+                                : ActivityType.UPDATE_TRANSLATION;
+
+                        logActivityAlreadyLocked(actorId,
```
seanf: I think we should be able to roll up all the activity for a single document (ie a single TextFlowTargetStateEvent), thus generating a single Activity outright instead of generating an Activity, then updating it repeatedly. Should reduce DB activity quite a bit.

zanata-war/src/main/java/org/zanata/service/impl/MergeTranslationsServiceImpl.java
```diff
@@ -237,28 +243,36 @@ private Integer mergeTranslations(
...
-                                        targetVersionId,
-                                        targetTf.getDocument().getId(),
-                                        targetTf.getId(),
-                                        updatedTarget.getLocale().getLocaleId(),
-                                        updatedTarget.getId(),
-                                        updatedTarget.getState(),
-                                        entry.getValue());
-
-                        textFlowTargetStateEventEvent.fire(event);
+                        DocumentLocaleKey key = new DocumentLocaleKey(null,
```
seanf: Why the null? We should know which user triggered the merge.

zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java
```diff
@@ -256,10 +271,21 @@
             result.translatedTextFlowTarget = hTextFlowTarget;
             results.add(result);
         }
-
+        fireTargetUpdateEvent(eventMap);
         return results;
     }
 
+    private void fireTargetUpdateEvent(
```
seanf: `fireTargetUpdateEvents` would be better.

zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
aeng: The activity type is `ActivityType.UPDATE_TRANSLATION` which expect single translation. If we want to refactor this, then it should be another pull request since it will affect the dashboard display.

zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java
```diff
@@ -186,6 +190,9 @@
         validateReviewPermissionIfApplicable(translationRequests,
                 projectIteration, hLocale);
 
+        Map<DocumentLocaleKey, List<TextFlowTargetState>> eventMap =
+            Maps.newHashMap();
```
seanf: I think the code relating to eventMap would be more readable if you were to use a MultiMap.

zanata-war/src/main/java/org/zanata/events/DocumentLocaleKey.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @author Alex Eng<a href="mailto:aeng@redhat.com">aeng@redhat.com</a>
+ */
+@AllArgsConstructor
+@EqualsAndHashCode
+@Getter
+public class DocumentLocaleKey implements Serializable{
+
+    // this may be null in the case of document uploads
+    private final @Nullable Long actorId;
```
seanf: This field doesn't fit with the name of the class, but rather than rename the class, would it be possible to factor it out of here? A single TextFlowTargetStateEvent should only be for a single actor.

zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
seanf: But don't we already roll up all Activity records for a single document within a period of time (15 minutes or similar)? The end result is a single Activity: x translations were updated, in this period of time, by this user, ending with this translation. With the rolled up events, we should be able to generate that lot in one go.

zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java
seanf: In fact, we shouldn't even need a map here. Within this method, every value in DocumentLocaleKey is a constant, so we should only need a single TextFlowTargetStateEvent.

zanata-war/src/main/java/org/zanata/service/impl/TranslationStateCacheImpl.java
```diff
@@ -191,18 +191,22 @@ public Boolean textFlowTargetHasWarningOrError(Long targetId,
...
-        // update document status information
-        updateDocStatusCache(key, event.getTextFlowTargetId());
+        for(TextFlowTargetStateEvent.TextFlowTargetState state: event.getStates()) {
+            Long tftId = state.getTextFlowTargetId();
 
-        // invalidate target validation
-        targetValidationCache.remove(event.getTextFlowTargetId());
+            // update document status information
+            updateDocStatusCache(new DocumentLocaleKey(documentId,
+                localeId), tftId);
```
seanf: We should only need to update docStatusCache once per TextFlowTargetStateEvent, much the same as we should be able to generate a rolled-up Activity directly from TextFlowTargetStateEvent.

zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java
```diff
@@ -55,23 +61,28 @@ public void textFlowStateUpdated(
...
-                            event.getProjectIterationId(),
-                            event.getDocumentId(), event.getLocaleId(),
-                            wordCount,
-                            event.getPreviousState(), event.getNewState())
-            );
+            for(TextFlowTargetState state: event.getStates()) {
+                int wordCount =
+                    textFlowDAO.getWordCount(state.getTextFlowId());
+                documentStatisticUpdatedEvent
+                    .fire(new DocumentStatisticUpdatedEvent(
```
seanf: We should be able to reduce the number of DocumentStatisticUpdatedEvents now that we have aggregation. We should put a todo here, perhaps something like:

    // TODO turn DocumentStatisticUpdatedEvent into an aggregate event like TextFlowTargetStateEvent
    // TODO generate one DocumentStatisticUpdatedEvent per TextFlowTargetStateEvent

zanata-war/src/main/java/org/zanata/service/impl/VersionStateCacheImpl.java
```diff
@@ -90,19 +93,21 @@ public void create() {
...
+    public void textFlowStateUpdated(
+        @Observes(during = TransactionPhase.AFTER_SUCCESS)
+            TextFlowTargetStateEvent event) {
         VersionLocaleKey key =
-                new VersionLocaleKey(event.getProjectIterationId(),
-                        event.getLocaleId());
+            new VersionLocaleKey(event.getKey().getProjectIterationId(),
+                event.getKey().getLocaleId());
         WordStatistic stats = versionStatisticCache.get(key);
-        if (stats != null) {
```
seanf: Is it still possible for stats to be null?

zanata-war/src/main/java/org/zanata/service/impl/VersionStateCacheImpl.java
```diff
@@ -90,19 +93,21 @@ public void create() {
...
-                    textFlow.getWordCount().intValue());
+        TextFlowDAO textFlowDAO = serviceLocator.getInstance(TextFlowDAO.class);
+
+        for (TextFlowTargetStateEvent.TextFlowTargetState state : event
+            .getStates()) {
+            int wordCount =
+                textFlowDAO.getWordCount(state.getTextFlowId());
+            stats.decrement(state.getPreviousState(), wordCount);
+            stats.increment(state.getNewState(), wordCount);
             versionStatisticCache.put(key, stats);
```
seanf: We don't need to do this cache.put inside the for loop.

zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
```diff
@@ -195,9 +195,9 @@ public void logTextFlowStateUpdate(@Observes(during = TransactionPhase.AFTER_SUC
                 HDocument document =
                     documentDAO.getById(event.getKey().getDocumentId());
```
seanf: To be on the safe side, I would move this back inside the transaction the way it used to be.

zanata-war/src/main/java/org/zanata/service/impl/VersionStateCacheImpl.java
aeng: possible if version page is not loaded before.

zanata-war/src/main/java/org/zanata/service/impl/TranslationStateCacheImpl.java
aeng: We need to pass in tftId into updateDocStatusCache. The cache needs to update `lastModifiedBy` information for that document by using target.getLastmodified.

zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
aeng: We do rollup (Every hour) if its the same EntityId which can be - HDoc, HProjectIter and HTFT.
For this case, the tft is all different.

seanf: @aeng The TFT is different each time, but ultimately we only store one Activity with (a) the total number of words translated and (b) a pointer to the last TFT. Don't we? That was the intention. 

If we're really generating one Activity for every TextFlowTarget (rolling up only when the TFT is modified within the hour), it's no wonder Activity is such a performance problem for us. How do all these Activities appear in the dashboard, eg when a translator translates 20 TFTs in the same document in an hour? Is that shown as 20 lines in the dashboard?

zanata-war/src/main/java/org/zanata/service/impl/TranslationStateCacheImpl.java
seanf: So why not pass in the last TFT details from the ImmutableList, and use that to update docStatusCache? All the TFTs should have the same timestamp and account if they're part of a batch, but using the last one's metadata should give the desired effect.

zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
aeng: > If we're really generating one Activity for every TextFlowTarget (rolling up only when the TFT is modified within the hour),

No. The entry in activity table will be created/updated based on version id, type of activity (eg. update translation) and rolled up within the hour.

>  How do all these Activities appear in the dashboard, eg when a translator translates 20 TFTs in the same document in an hour? Is that shown as 20 lines in the dashboard?

If you constantly update translations in the same version, then its only 1 entry in the activity table.

aeng: But we do need the targetId to record against that entry as the last translated string.

aeng: I think we can use the last TFT details in what you suggested in `TranslationStateCacheImpl`. What do you think?

aeng: But then `translation` and `review` is 2 different activity type. We can try to aggregate these 2 types of activity, total up the word counts and fire 2 event.

seanf: Sounds good.

On 15 April 2016 at 15:27, Alex Eng <notifications@github.com> wrote:

> In
> zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
> <https://github.com/zanata/zanata-server/pull/1138#discussion_r59828360>:
>
> > -                });
> > +                HDocument document =
> > +                    documentDAO.getById(event.getKey().getDocumentId());
> > +
> > +                for (TextFlowTargetStateEvent.TextFlowTargetState state : event
> > +                    .getStates()) {
> > +                    transactionUtil.run(() -> {
> > +                        HTextFlowTarget target =
> > +                            textFlowTargetDAO.findById(
> > +                                state.getTextFlowTargetId(), false);
> > +                        ActivityType activityType =
> > +                            state.getNewState().isReviewed()
> > +                                ? ActivityType.REVIEWED_TRANSLATION
> > +                                : ActivityType.UPDATE_TRANSLATION;
> > +
> > +                        logActivityAlreadyLocked(actorId,
>
> But then translation and review is 2 different activity type. We can try
> to aggregate these 2 types of activity, total up the word counts and fire 2
> event.
>
> —
> You are receiving this because you were assigned.
> <https://github.com/zanata/zanata-server/pull/1138/files/462a372aa89c14a002268b729deeb53d52a19f33#r59828360>
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

zanata-war/src/main/java/org/zanata/service/impl/TranslationUpdatedManager.java
```diff
@@ -55,23 +61,30 @@ public void textFlowStateUpdated(
...
-                            event.getProjectIterationId(),
-                            event.getDocumentId(), event.getLocaleId(),
-                            wordCount,
-                            event.getPreviousState(), event.getNewState())
-            );
+            for(TextFlowTargetState state: event.getStates()) {
+                int wordCount =
+                    textFlowDAO.getWordCount(state.getTextFlowId());
+                // TODO: turn DocumentStatisticUpdatedEvent into an aggregate event like TextFlowTargetStateEvent
+                // TODO: generate one DocumentStatisticUpdatedEvent per TextFlowTargetStateEvent
```
seanf: Sorry, can you make that TODO PERF? It looks like DocumentServiceImpl.documentStatisticUpdated is the only observer for DocumentStatisticUpdatedEvent, and it would really benefit from batching, especially for projects which have webhooks enabled.

// TODO PERF: generate one DocumentStatisticUpdatedEvent per TextFlowTargetStateEvent. See DocumentServiceImpl.documentStatisticUpdated

zanata-war/src/main/java/org/zanata/service/impl/ActivityServiceImpl.java
```diff
@@ -184,24 +187,56 @@ public Object getEntity(EntityType entityType, long entityId) {
...
-                            document.getProjectIteration(), target, activityType,
-                            target.getTextFlow().getWordCount().intValue());
+                    HDocument document =
+                        documentDAO.getById(event.getKey().getDocumentId());
+
+                    HTextFlowTarget lastReviewedTarget = null;
+                    HTextFlowTarget lastTranslatedTarget = null;
+
+                    int totalReviewedWord = 0;
+                    int totalTranslatedWord = 0;
```
seanf: Plurals (words) would be better. Thanks.

zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java
```diff
@@ -859,31 +878,36 @@ public String apply(TextFlowTarget input) {
...
-
             if (handleOp.isPresent()) {
                 handleOp.get().increaseProgress(1);
             }
         }
+        Long actorId =
+            assignCreditToUploader ? authenticatedAccount.getPerson().getId() :
+                null;
+        for (Map.Entry<DocumentLocaleKey, Collection<TextFlowTargetState>> entry : eventMap
+            .asMap().entrySet()) {
```
seanf: This method (saveBatch) works on a single HLocale and HDocument, so there should only be a single TextFlowTargetStateEvent. We shouldn't need the Map/MultiMap here either.

zanata-war/src/main/java/org/zanata/service/impl/TranslationStateCacheImpl.java
```diff
@@ -191,18 +191,23 @@ public Boolean textFlowTargetHasWarningOrError(Long targetId,
...
-        // update document status information
-        updateDocStatusCache(key, event.getTextFlowTargetId());
+        for(TextFlowTargetStateEvent.TextFlowTargetState state: event.getStates()) {
+            // invalidate target validation
+            targetValidationCache.remove(state.getTextFlowTargetId());
+        }
 
-        // invalidate target validation
-        targetValidationCache.remove(event.getTextFlowTargetId());
+        Long tftId = event.getStates().get(event.getStates().size() - 1)
```
seanf: Iterables.getLast from Guava would read more nicely (esp. with static import).

docs/release-notes.md
```diff
@@ -32,6 +32,7 @@
 * [ZNTA-905](https://zanata.atlassian.net/browse/ZNTA-905) - Remove 0% matching translation memory entry
 * [ZNTA-928](https://zanata.atlassian.net/browse/ZNTA-928) - Readonly project doesn't have "lock" icon in UI
 * [ZNTA-793](https://zanata.atlassian.net/browse/ZNTA-793) - Remove unused method in TranslationMemoryAction
+* [ZNTA-1017](https://zanata.atlassian.net/browse/ZNTA-1017) - Refactor translation update event
```
seanf: I don't think refactoring tasks like ZNTA-1017 or ZNTA-793 should appear in the release notes.

zanata-war/src/main/java/org/zanata/service/impl/TranslationServiceImpl.java
```diff
@@ -186,6 +192,12 @@
         validateReviewPermissionIfApplicable(translationRequests,
                 projectIteration, hLocale);
 
+        DocumentLocaleKey documentLocaleKey =
+            new DocumentLocaleKey(projectIteration.getId(),
+                sampleHTextFlow.getDocument().getId(), hLocale.getLocaleId());
```
seanf: This `documentLocaleKey` definition could be moved down to just before `new TextFlowTargetStateEvent`

zanata-war/src/main/java/org/zanata/events/DocumentLocaleKey.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.io.Serializable;
+
+/**
+ * @author Alex Eng<a href="mailto:aeng@redhat.com">aeng@redhat.com</a>
+ */
+@AllArgsConstructor
+@EqualsAndHashCode
+@Getter
+public class DocumentLocaleKey implements Serializable {
```
seanf: - [ ] We already have two other classes named `DocumentLocaleKey` - the only remaining difference between this and the other two is that this one includes the somewhat redundant field `projectIterationId`. Is there any way we can rationalise them all?

seanf:
:+1:

djansen-redhat:
After running copyTrans and seeing a couple of items updated in the logs, the version looks like this:
screenshot from 2016-04-20 14-19-48

djansen-redhat:
:white_check_mark: Looks fine to me

PR #1140, opened 2016-04-17
docs: improve English in translator-guide.md

fitojb:
In this trivial patch, I’ve added a synonym to avoid so many “click on…”, corrected prepositions and tenses, and fixed the phrasal verb “to log in”.


seanf:
Thanks for the contribution, @fitojb!

PR #1141, opened 2016-04-22
fix(ZNTA-470): contact admin validation and workflow

broccolinisoup:
contact admin text field validation has been added and the workflow of the sending process has been fixed.


seanf:
@zanata-jenkins ok to test

aeng:
@broccolinisoup 

run mvn:checkstyle:check to fix the styling issue.

zanata-war/src/main/webapp/WEB-INF/layout/admin/contact_admin_modal.xhtml
```diff
@@ -32,10 +34,26 @@
...
+    }
+
     function toggleSendButton() {
       var hasMessage = getContactAdminMessageField().val() !== '';
       getContactAdminSendButton().prop('disabled', !hasMessage);
     }
+
+    function completeSending() {
+      var errMsg = getContactAdminMessageErrorField().text();
+      if (errMsg != '') {
```
djansen-redhat: Is this a bit "dangerous" at all - e.g. can it be bypassed in some way by the user?

djansen-redhat:
This appears to fix both ZNTA-12 and ZNTA-470 - intentional?

aeng:
@djansen-redhat 
Yes. @broccolinisoup discovered the issue while fixing ZNTA-470

zanata-war/src/main/webapp/WEB-INF/layout/admin/contact_admin_modal.xhtml
broccolinisoup: I tried to break it by this way
getContactAdminForm().find('#contactAdminForm\\:messageField\\:input\\:contact-admin-error-message\\:contactAdminForm\\:messageField\\:input\\:contact-admin-message').**text('');**

It is broken now and validation error is disable. However, the user cannot submit a looong message in any way because sending process is happening in the back end side. 

Please correct me if I am wrong or try to break it. I am not a good hacker.

aeng: Instead of checking the text, can you only check if the error field is displayed or not?

broccolinisoup: I tried to control visibility of the span. 
element.is(":disabled")  element function doesn't work for a span. It returns "false" in both cases.
It seems that everything is fine with visibility check.

djansen-redhat:
:white_check_mark: Tested, waiting for review completion and Jenkins green light

zanata-war/src/main/webapp/WEB-INF/layout/admin/contact_admin_modal.xhtml
```diff
@@ -32,10 +34,26 @@
       return getContactAdminForm().find('#contact-admin-send-button');
     }
 
+    // Returns a span.
+    function getContactAdminMessageErrorField(){
+      return getContactAdminForm().find('#contactAdminForm\\:messageField\\:input\\:contact-admin-error-message\\:contactAdminForm\\:messageField\\:input\\:contact-admin-message');
```
carlosmunoz: This seems like a very long Id. Is there a better way to find it?

You could use Richfaces' `rich:componentId('contact-admin-message')` which would return the whole client side id for the jsf component. After that it could be passed to the `getContactAdminForm().find(...)` method.

broccolinisoup: We've tried different solutions for my js training with @carlosmunoz and changed a few things. Here how we went;

We removed the controlling of "send button enability". JSF "required=true" and "max length" validations are enough not to send invalid messages to the admin.

We changed the way of getting error message field from id to class to get rid of ridiculously long id. The function returns a span which cannot be controlled by its visibility. So, again we controlled its text. Carlos also adds that "Both of them can bypassed by someone. Because it is javascript." However, any bypasses cannot send invalid message because jsf validation is doing that on the java side. So no worries.

Finally, I added a label that shows the left characters. It was just an experimental in order to be bias-free to javascript . If you guys are not happy with that, I can remove it. 

Apparently, it works well.

aeng:
:+1:  from me once green light from jenkins.

broccolinisoup:
Rebased from master!

aeng:
Seems like functional test is broken in our master branch too after rebase.

broccolinisoup:
Shall we start to build again? @djansen-redhat @aeng Also, I don't know  if I can or how to do that.

djansen-redhat:

aeng:
:+1: All good!

definite:
Note that in order to make this fix appear in 3.9.0, this pull request should be backport to release branch

definite:
Backport completed.

PR #1143, opened 2016-04-27
fix(ZNTA-931): incorrect scopes

broccolinisoup:
I am inspired of TODO statement which was "// need to get username from DAO due to lazy loading of account in key". I got the username from AccountDAO instead of through HAccountResetPasswordKey entity.

### DIFF
--- a/zanata-war/src/main/java/org/zanata/action/PasswordResetAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/PasswordResetAction.java
@@ -1,29 +1,28 @@
 package org.zanata.action;
 
-import java.io.Serializable;
-
-import javax.persistence.EntityManager;
-import javax.validation.constraints.Size;
-
 import lombok.Getter;
 import lombok.Setter;
 import org.apache.deltaspike.core.api.scope.GroupedConversation;
 import org.apache.deltaspike.core.api.scope.GroupedConversationScoped;
 import org.apache.deltaspike.jpa.api.transaction.Transactional;
 import org.hibernate.validator.constraints.NotEmpty;
-import javax.inject.Inject;
-import javax.inject.Named;
-
+import org.zanata.ApplicationConfiguration;
+import org.zanata.dao.AccountDAO;
 import org.zanata.dao.AccountResetPasswordKeyDAO;
 import org.zanata.exception.AuthorizationException;
-import org.zanata.exception.NotLoggedInException;
-import org.zanata.ApplicationConfiguration;
 import org.zanata.exception.KeyNotFoundException;
+import org.zanata.exception.NotLoggedInException;
 import org.zanata.i18n.Messages;
 import org.zanata.model.HAccountResetPasswordKey;
 import org.zanata.seam.security.AbstractRunAsOperation;
 import org.zanata.seam.security.IdentityManager;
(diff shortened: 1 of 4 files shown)


djansen-redhat:
Did this get based off an incorrect branch?

carlosmunoz:
@djansen-redhat Doesn't look like it... what makes you think that?

broccolinisoup:
@djansen-redhat Yea at first I accidentally created it from ZNTA-470 but I revert it later and in this morning I caught up the origin master. @carlosmunoz  maybe from commits?

djansen-redhat:
@carlosmunoz I'm seeing commits for both contact admin fix (not yet merged) and reset password fix

broccolinisoup:
@carlosmunoz Could you review it? I cannot also add a label.

carlosmunoz:
So, I'm actually tempted to say that this was not necessary at all after looking at the code, and for the following reasons:

1. The purpose of the comment in the code was not to flag a TODO, but rather to explain why there is a need to query the data from the database instead of getting it from the `HAccountResetPasswordKey` object already in memory. The code is already performing such a query to fetch the `HAccount`'s details.

2. The query being performed in the current code is actually more efficient than the proposed one in the pull request; It is a 2-table join with a primary key filter. The proposed one is doing a correlated subquery (a query in the where clause) which is not nearly as efficient (although it has its uses).

seanf:
@carlosmunoz makes good points, but for next time, it's always good to give the pull request a meaningful name, eg "Znta 931" might be better as "ZNTA-931 get username from DAO".

@broccolinisoup also, please have a look at this guide to good commit messages (use the imperative!): http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

And https://github.com/zanata/zanata-server/blob/master/CONTRIBUTING.md#-commit-message-format

broccolinisoup:
@seanf Alex sent me these links a few days ago which I sent a commit with this format : ( fix(ZNTA-960): modify try catch block) I will follow both commit message and pull req format. Thank you.

@carlosmunoz I completely understood you point. Which one is the query being performed that you mentioned? I couldn't find a proper function -apparently I missed it - so that is the reason why I wrote a new one.

zanata-war/src/main/java/org/zanata/action/PasswordResetAction.java
```diff
@@ -108,9 +110,9 @@ public void validateActivationKey() {
 
     @Transactional
     public String changePassword() {
-        // need to get username from DAO due to lazy loading of account in key
         String username =
-            accountResetPasswordKeyDAO.getUsername(key.getKeyHash());
```
carlosmunoz: @broccolinisoup This is the query I was referring to. If you follow the `getUsername` method on this DAO you'll see the query.

broccolinisoup:
@carlosmunoz I just committed new changes that we did yesterday together. Could you take a quick look again?

carlosmunoz:
:+1: 
(let's see what Jenkins thinks)

broccolinisoup:
Seems that jenkins liked it.  The next step is testing? @djansen-redhat @carlosmunoz

broccolinisoup:
Move to https://github.com/zanata/zanata-server/pull/1159

PR #1146, opened 2016-04-28
fix(ZNTA-906): try catch for NullPointerException

broccolinisoup:
As it is written in jira ticket https://zanata.atlassian.net/browse/ZNTA-906, somehow  a NullPointerException is occurred when viewing an iteration of a project. It seems that it is caused by "getProject()" function returning null value at this case. 

There is also a possibility to be cause by the iteration which comes null somehow. 
Apparently, this is an "exception" which we may not encounter frequently. So I reckon putting try catch block is better than controlling the iteration and getProject() value are null or not.

Also, as I wrapped if-else statement with try catch, I initialized the local variable with **null** value which I am not enough happy with.

I am not completely sure about my solution. Is there any idea?

### DIFF
--- a/zanata-war/src/main/java/org/zanata/service/impl/LocaleServiceImpl.java
+++ b/zanata-war/src/main/java/org/zanata/service/impl/LocaleServiceImpl.java
@@ -22,6 +22,7 @@ package org.zanata.service.impl;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -89,17 +90,18 @@ public class LocaleServiceImpl implements LocaleService {
     }
 
     public static Map<LocaleId, String> getLocaleAliasesByIteration(HProjectIteration iteration) {
-        Map<LocaleId, String> localeAliases = null;
+        Map<LocaleId, String> localeAliases;
         try {
             if (iteration.isOverrideLocales()) {
                 localeAliases = iteration.getLocaleAliases();
             } else {
                 localeAliases = iteration.getProject().getLocaleAliases();
             }
+            return localeAliases;
         }catch (NullPointerException e) {
             e.printStackTrace();
         }
-        return localeAliases;
+        return Maps.newHashMap();
     }
 
     @Inject
--- a/zanata-war/src/main/java/org/zanata/service/impl/LocaleServiceImpl.java
+++ b/zanata-war/src/main/java/org/zanata/service/impl/LocaleServiceImpl.java
@@ -89,11 +89,15 @@ public class LocaleServiceImpl implements LocaleService {
     }
 
     public static Map<LocaleId, String> getLocaleAliasesByIteration(HProjectIteration iteration) {
-        Map<LocaleId, String> localeAliases;
-        if (iteration.isOverrideLocales()) {
-            localeAliases = iteration.getLocaleAliases();
-        } else {
-            localeAliases = iteration.getProject().getLocaleAliases();
+        Map<LocaleId, String> localeAliases = null;
+        try {
+            if (iteration.isOverrideLocales()) {
+                localeAliases = iteration.getLocaleAliases();
+            } else {
+                localeAliases = iteration.getProject().getLocaleAliases();
+            }
+        }catch (NullPointerException e) {
+            e.printStackTrace();
         }
         return localeAliases;
     }


broccolinisoup:
Could you review it please @aeng ?

aeng:
@zanata-jenkins ok to test

zanata-war/src/main/java/org/zanata/service/impl/LocaleServiceImpl.java
```diff
@@ -89,11 +89,15 @@ public LocaleServiceImpl(LocaleDAO localeDAO, ProjectDAO projectDAO,
     }
 
     public static Map<LocaleId, String> getLocaleAliasesByIteration(HProjectIteration iteration) {
-        Map<LocaleId, String> localeAliases;
-        if (iteration.isOverrideLocales()) {
-            localeAliases = iteration.getLocaleAliases();
-        } else {
-            localeAliases = iteration.getProject().getLocaleAliases();
+        Map<LocaleId, String> localeAliases = null;
```
aeng: Don't think this will solve the nullpointer exception. I think we can return `localeAliases` in the try statement. Then return `new HashMap<LocaleId, String>` at the end of the method to prevent the exception.

broccolinisoup:
I got your point @aeng. Actually,  I ignored that the returned value is used somewhere. Of course it cannot solve the NullPointerException problem. Shame on me!

zanata-war/src/main/java/org/zanata/service/impl/LocaleServiceImpl.java
```diff
@@ -89,17 +90,18 @@ public LocaleServiceImpl(LocaleDAO localeDAO, ProjectDAO projectDAO,
     }
 
     public static Map<LocaleId, String> getLocaleAliasesByIteration(HProjectIteration iteration) {
-        Map<LocaleId, String> localeAliases = null;
+        Map<LocaleId, String> localeAliases;
```
aeng: You don't need to declare this map at all. Just put the return statement in line 96 and 98.

zanata-war/src/main/java/org/zanata/service/impl/LocaleServiceImpl.java
```diff
@@ -90,12 +91,17 @@ public LocaleServiceImpl(LocaleDAO localeDAO, ProjectDAO projectDAO,
 
     public static Map<LocaleId, String> getLocaleAliasesByIteration(HProjectIteration iteration) {
         Map<LocaleId, String> localeAliases;
```
aeng: What i meant was you might no need this line at all. 
Just do 
`return iteration.getLocaleAliases();` in line 96 and 
`return iteration.getProject().getLocaleAliases();` in line 98. 

Also, your checkstyle might fail in line 101.

broccolinisoup: Hey @aeng I got your point but I was busy with the other ticket so I couldn't answer this. 
By the way, I was about to write this comment to you. 

We looked at this PR with @carlosmunoz First, tried to reproduce this exception but we couldn't and Carlos realized that this NullPointerException is caused by Seam. So he is considering closing the PR.

broccolinisoup:
Should we close this? @aeng @carlosmunoz

carlosmunoz:
Well, that was just a theory 😄 ... but the related bug doesn't have much details and we weren't able to reproduce it either.
I am personally hesitant of going forward with this PR as even the proposed solution doesn't really address the issue. My vote is to close.

aeng:
:+1: on what @carlosmunoz mentioned if we can't reproduce the issue.

zanata-war/src/main/java/org/zanata/service/impl/LocaleServiceImpl.java
```diff
@@ -90,12 +91,17 @@ public LocaleServiceImpl(LocaleDAO localeDAO, ProjectDAO projectDAO,
...
-            localeAliases = iteration.getProject().getLocaleAliases();
+        try {
+            if (iteration.isOverrideLocales()) {
+                localeAliases = iteration.getLocaleAliases();
+            } else {
+                localeAliases = iteration.getProject().getLocaleAliases();
+            }
+            return localeAliases;
+        }catch (NullPointerException e) {
+            e.printStackTrace();
```
seanf: Catching `NullPointerException` is hardly ever the best option (except _maybe_ for a throwaway local code change). Catching but not propagating the exception is generally going to lead to more problems. Besides, Zanata should already be logging the stack trace elsewhere, as long as the exception is propagated up the call stack.

Also, if you do want to log an exception's stack trace, it's better to use a Logger, eg `log.error("unexpected error", e) instead of dumping to `stderr`.

aeng:
@seanf we are closing this PR.

seanf:
@aeng I know, but I thought the point was worth making, for similar bugs in future.

seanf:
Everyone's talking about closing the PR, so we might as well close it.

PR #1151, opened 2016-05-03
fix(menu drop down): drop-down user menu does not obscure the Glossary button

aeng:
https://zanata.atlassian.net/browse/ZNTA-1065


frontend/src/main/web/lib/components/SystemGlossary.jsx
```diff
@@ -140,7 +140,7 @@ var SystemGlossary = React.createClass({
               onChange={this._onTranslationLocaleChange}/>
             {loader}
           </div>
-          <div className='dfx aic'>
+          <div className='dfx aic' style="z-index:0">
```
huangp: is there a class we can use? this seems to be a quick hack

huangp:
:+1:

definite:
:+1:

PR #1152, opened 2016-05-04
refactor: replace jndi config with system properties

carlosmunoz:
Use system properties instead of JNDI bindings for configuration. This should allow for set configuration in the standalone.xml file as well as overriding properties from the command line.

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1152)


docs/release-notes.md
```diff
@@ -1,3 +1,11 @@
+## 4.0
+
+<h5>Infrastructure Changes</h5>
+* Zanata has forgone all JNDI-based configuration and replaced it with system properties. Please see the following sections for how certain values are now configured:
```
seanf: I would suggest "eliminated" rather than "forgone".

docs/user-guide/system-admin/configuration/document-storage-directory.md
```diff
@@ -1,31 +1,27 @@
...
-    ...
-</subsystem>
+<property name="zanata.files.documentstorage.directory" value="/example/path"/>
 ```
 
+Alternatively, you can pass this and other system properties to JBoss when starting it (see JBoss documentation for details on how to do this).
+
 ## Running Liquibase from command line
-This section will not explain how to run liquibase from the command line, just how to add the required parameter. The parameter is ```document.storage.directory``` and should be added to the liquibase command in the form ```-Ddocument.storage.directory=/example/path```. The value passed to this parameter should exactly match the value bound to the above JNDI name.
+This section will not explain how to run liquibase from the command line, just how to add the required parameter. The parameter is ```document.storage.directory``` and should be added to the liquibase command in the form ```-Ddocument.storage.directory=/example/path```. The value passed to this parameter should exactly match the value bound to the system property above.
```
seanf: Does this mean it is now controlled by two different system properties: `zanata.files.documentstorage.directory` and `document.storage.directory`? One should be enough.

I think a better name might be `zanata.document.directory`, now that it doesn't have to fit into a JNDI naming scheme.

seanf: We need to be careful when changing the liquibase parameter. Liquibase may treat the changeset as modified, just because we stop passing a parameter to it, or because the parameter value changes. See https://github.com/zanata/zanata-server/blob/server-3.9.0-rc-1/zanata-war/src/main/resources/db/changelogs/db.changelog-3.1.xml#L265

I suppose that might be an argument for sticking to `document.storage.directory`.

carlosmunoz: @seanf What do you think of supporting the new property (so that we have a nice set of zanata-namespaced system properties), and internally adding some code to make sure that `document.storage.directory` gets populated with the value of the new property. Not sure if that's possible considering the order of execution of it all.

Alternatively, and since this will only be a problem when updating Zanata from a certain version X to this one, we could make it a documentation entry that states that when doing this massive upgrade, you might need to start the system with both properties.

Your thoughts?

seanf: Well, if we're okay with requiring admins to migrate from the old property
to the new one, we can fail startup if the new property is missing.

https://github.com/zanata/zanata-server/blob/server-3.9.0-rc-1/zanata-war/src/main/resources/db/changelogs/db.changelog-3.1.xml#L265

We can probably work around the checksum problem (if any) using
validCheckSum. The whole setup for that changeset is a bit broken, because
rearranging your filesystem at some point in the future shouldn't cause a
database migration error, long after the documents were migrated from DB to
file system.

Perhaps we should just eliminate the parameter basePathParam and have the
custom code look for the new system property internally (and failing if
it's missing). (This will require the use of validCheckSum.)

So I think we can get by with just the new property.

On 6 May 2016 at 10:13, Carlos Munoz <notifications@github.com> wrote:

> In
> docs/user-guide/system-admin/configuration/document-storage-directory.md
> <https://github.com/zanata/zanata-server/pull/1152#discussion_r62274950>:
>
> >  ## Running Liquibase from command line
> > -This section will not explain how to run liquibase from the command line, just how to add the required parameter. The parameter is ```document.storage.directory``` and should be added to the liquibase command in the form ```-Ddocument.storage.directory=/example/path```. The value passed to this parameter should exactly match the value bound to the above JNDI name.
> > +This section will not explain how to run liquibase from the command line, just how to add the required parameter. The parameter is ```document.storage.directory``` and should be added to the liquibase command in the form ```-Ddocument.storage.directory=/example/path```. The value passed to this parameter should exactly match the value bound to the system property above.
>
> @seanf <https://github.com/seanf> What do you think of supporting the new
> property (so that we have a nice set of zanata-namespaced system
> properties), and internally adding some code to make sure that
> document.storage.directory gets populated with the value of the new
> property. Not sure if that's possible considering the order of execution of
> it all.
>
> Alternatively, and since this will only be a problem when updating Zanata
> from a certain version X to this one, we could make it a documentation
> entry that states that when doing this massive upgrade, you might need to
> start the system with both properties.
>
> Your thoughts?
>
> —
> <https://github.com/zanata/zanata-server/pull/1152/files/322e50197bb4d13f502a91d47b8e0def0832a7c4#r62274950>
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

seanf:
Reviewed 10 of 10 files at r1.

---

*[docs/user-guide/system-admin/configuration/document-storage-directory.md, line 19 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1152#-KHcEcRYP8HVeiO6t45W-r1-19:-KHcFLeGyPcV_yrVme9x:1665111598) ([raw file](https://github.com/zanata/zanata-server/blob/322e50197bb4d13f502a91d47b8e0def0832a7c4/docs/user-guide/system-admin/configuration/document-storage-directory.md#L19)):*
> Well, if we're okay with requiring admins to migrate from the old property
> to the new one, we can fail startup if the new property is missing.
> 
> https://github.com/zanata/zanata-server/blob/server-3.9.0-rc-1/zanata-war/src/main/resources/db/changelogs/db.changelog-3.1.xml#L265
> 
> We can probably work around the checksum problem (if any) using
> validCheckSum. The whole setup for that changeset is a bit broken, because
> rearranging your filesystem at some point in the future shouldn't cause a
> database migration error, long after the documents were migrated from DB to
> file system.
> 
> Perhaps we should just eliminate the parameter basePathParam and have the
> custom code look for the new system property internally (and failing if
> it's missing). (This will require the use of validCheckSum.)
> 
> So I think we can get by with just the new property.
> 
> On 6 May 2016 at 10:13, Carlos Munoz <notifications@github.com> wrote:
> 
> > In
> > docs/user-guide/system-admin/configuration/document-storage-directory.md
> > <https://github.com/zanata/zanata-server/pull/1152#discussion_r62274950>:
> >
> > >  ## Running Liquibase from command line
> > > -This section will not explain how to run liquibase from the command line, just how to add the required parameter. The parameter is ```document.storage.directory``` and should be added to the liquibase command in the form ```-Ddocument.storage.directory=/example/path```. The value passed to this parameter should exactly match the value bound to the above JNDI name.
> > > +This section will not explain how to run liquibase from the command line, just how to add the required parameter. The parameter is ```document.storage.directory``` and should be added to the liquibase command in the form ```-Ddocument.storage.directory=/example/path```. The value passed to this parameter should exactly match the value bound to the system property above.
> >
> > @seanf <https://github.com/seanf> What do you think of supporting the new
> > property (so that we have a nice set of zanata-namespaced system
> > properties), and internally adding some code to make sure that
> > document.storage.directory gets populated with the value of the new
> > property. Not sure if that's possible considering the order of execution of
> > it all.
> >
> > Alternatively, and since this will only be a problem when updating Zanata
> > from a certain version X to this one, we could make it a documentation
> > entry that states that when doing this massive upgrade, you might need to
> > start the system with both properties.
> >
> > Your thoughts?
> >
> > —
> > You are receiving this because you were mentioned.
> > Reply to this email directly or view it on GitHub
> > <https://github.com/zanata/zanata-server/pull/1152/files/322e50197bb4d13f502a91d47b8e0def0832a7c4#r62274950>
> >
> 
> 
> 
> -- 
> Sean Flanigan
> 
> Principal Software Engineer
> Globalisation Tools Engineering
> Red Hat
> </details>
Just repeating what I said before, I think we should go with a single property, eg `zanata.document.directory`, eliminate the liquibase parameter `basePathParam` and have the custom code in `MigrateRawDocumentsToFileSystem` look for the new system property internally (and throw an exception if it's missing when the changeset is executed). This will require the use of `validCheckSum`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1152)*

carlosmunoz:
@seanf this is ready for another round of review

> #### refactor: replace jndi config with system properties
> 
> Use system properties instead of JNDI bindings for configuration. This should allow for set configuration in the standalone.xml file as well as overriding properties from the command line.
> 
> 
> </details>


---







*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1152#-:-KHwcn0XMuI6WexoH7kO:-1220490698)*

seanf:
Reviewed 11 of 11 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java, line 161 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1152#-KHwgPi_j0LQUmASQD6Y:-KHwgPi_j0LQUmASQD6Z:-2102431204) ([raw file](https://github.com/zanata/zanata-server/blob/2488f0e0c8d3d2b4cda2b50621f20fd6ad4073ac/zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java#L161)):*
> ```Java
>                 log.warn(
>                         "Attempted to configure an unrecognized authentication policy: " +
>                                 policyName);
> ```

Are you sure we shouldn't abort deployment/startup in this case?

---

*[zanata-war/src/main/java/org/zanata/liquibase/custom/MigrateRawDocumentsToFileSystem.java, line 115 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1152#-KHwgg8mF7pe1Lh8JuR7:-KHwgg8mF7pe1Lh8JuR8:1000673995) ([raw file](https://github.com/zanata/zanata-server/blob/2488f0e0c8d3d2b4cda2b50621f20fd6ad4073ac/zanata-war/src/main/java/org/zanata/liquibase/custom/MigrateRawDocumentsToFileSystem.java#L115)):*
> ```Java
> 
>     private String getConfiguredBasePath() throws SetupException {
>         return tryGetFallbackBasePath();
> ```

We should probably inline that method here (since there's no fallback any more). In fact, we could just put `@Setter` on the field.

---

*[zanata-war/src/main/java/org/zanata/liquibase/custom/MigrateRawDocumentsToFileSystem.java, line 125 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1152#-KHwgwsPQ9iJIxQ9wKY5:-KHwgwsPQ9iJIxQ9wKY6:-988610656) ([raw file](https://github.com/zanata/zanata-server/blob/2488f0e0c8d3d2b4cda2b50621f20fd6ad4073ac/zanata-war/src/main/java/org/zanata/liquibase/custom/MigrateRawDocumentsToFileSystem.java#L125)):*
> ```Java
>                             + "Fallback liquibase parameter \""
>                             + FILE_DIR_PROP_NAME + "\" not set. "
>                             + "Cannot migrate documents to file system.");
> ```

This error message needs to be updated. Also, we should move this check into the setUp method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1152)*

carlosmunoz:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java, line 161 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1152#-KHwgPi_j0LQUmASQD6Y:-KHwnS9djmMI5E4tsefZ:-1672109362) ([raw file](https://github.com/zanata/zanata-server/blob/2488f0e0c8d3d2b4cda2b50621f20fd6ad4073ac/zanata-war/src/main/java/org/zanata/ApplicationConfiguration.java#L161)):*
> Are you sure we shouldn't abort deployment/startup in this case?</details>
I'm sure. I designed it this way initially so we could have "inactive" authentication systems (by putting an invalid name). There's also a validation which completely fails startup if there are no authentication policies configured.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1152)*

carlosmunoz:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[docs/user-guide/system-admin/configuration/document-storage-directory.md, line 19 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1152#-KHcEcRYP8HVeiO6t45W-r1-19:-KHwo9tEKpQvzPBjXNpj:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/322e50197bb4d13f502a91d47b8e0def0832a7c4/docs/user-guide/system-admin/configuration/document-storage-directory.md#L19)):*
> Just repeating what I said before, I think we should go with a single property, eg `zanata.document.directory`, eliminate the liquibase parameter `basePathParam` and have the custom code in `MigrateRawDocumentsToFileSystem` look for the new system property internally (and throw an exception if it's missing when the changeset is executed). This will require the use of `validCheckSum`.</details>
Done.

---

*[zanata-war/src/main/java/org/zanata/liquibase/custom/MigrateRawDocumentsToFileSystem.java, line 115 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1152#-KHwgg8mF7pe1Lh8JuR7:-KHwo4m56VXcHgL5WsAr:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/2488f0e0c8d3d2b4cda2b50621f20fd6ad4073ac/zanata-war/src/main/java/org/zanata/liquibase/custom/MigrateRawDocumentsToFileSystem.java#L115)):*
> We should probably inline that method here (since there's no fallback any more). In fact, we could just put `@Setter` on the field.</details>
Done.

---

*[zanata-war/src/main/java/org/zanata/liquibase/custom/MigrateRawDocumentsToFileSystem.java, line 125 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1152#-KHwgwsPQ9iJIxQ9wKY5:-KHwo6_SII7XQphqSuom:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/2488f0e0c8d3d2b4cda2b50621f20fd6ad4073ac/zanata-war/src/main/java/org/zanata/liquibase/custom/MigrateRawDocumentsToFileSystem.java#L125)):*
> This error message needs to be updated. Also, we should move this check into the setUp method.</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1152)*

seanf:
:lgtm:

> @seanf this is ready for another round of review</details>


---

Reviewed 1 of 11 files at r2, 3 of 3 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1152#-:-KHxQOXv3uknSVwXjeC8:1417088721)*

carlosmunoz:
@djansen-redhat @anipeter @definite Please test this [build](http://zanata-build.lab.eng.bne.redhat.com/job/zanata-server-github-pull-requests/5880/)

I made the mistake of merging before this was QA'd and had to clean up. The PR still shows as merged but it isn't. Changes due to QA findings will require a new pull request.

PR #1149, opened 2016-05-03
Control CORS with system property

seanf:
https://zanata.atlassian.net/browse/ZNTA-1042

PR #1125, opened 2016-03-29
Redux app

aeng:
- User profile update
- Glossary update
- new side bar integration

https://zanata.atlassian.net/browse/ZNTA-858

Reviewed branch: https://github.com/zanata/zanata-server/tree/redux-app-reviewed
https://github.com/zanata/zanata-server/pull/1142

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1125)

PR #1156, opened 2016-05-05
ZNTA-1042 Control CORS with system property

seanf:
https://zanata.atlassian.net/browse/ZNTA-1042

Rebase of https://github.com/zanata/zanata-server/pull/1149

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1156)

### DIFF
--- a/docs/user-guide/system-admin/configuration/installation.md
+++ b/docs/user-guide/system-admin/configuration/installation.md
@@ -62,6 +62,21 @@ Zanata does not create an admin user by default. You need to register specific u
 
  This is the default email address that will appear as the sender on Zanata emails.
 
+## Enabling Cross-Origin Resource Sharing (advanced)
+
+If you have an alternative front-end for Zanata where the browser needs direct access to Zanata REST
+APIs (and authenticated user sessions), you will need to enable Cross-Origin Resource Sharing, by
+adding a system property like this to `<JBOSS>/standalone/configuration/standalone.xml`:
+
+    <property name="zanata.originWhitelist" value="http://localhost:8000" />
+
+You should adjust the protocol, hostname and port to suit your use case. Multiple origins should
+be separated via space characters.
+
+This setting will cause Zanata to output CORS headers for REST requests, including
+`Access-Control-Allow-Origin` (if Origin is in the whitelist above),
+`Access-Control-Allow-Credentials: true` and
+`Access-Control-Allow-Methods: PUT,POST,DELETE,GET,OPTIONS`.
 ### Email configuration (Zanata 3.6 or earlier)
 
 In Zanata 3.6 or earlier, email is configured by admin in the server settings screen.  By default, an SMTP server on localhost port 25 is expected.
--- a/zanata-war/src/main/java/org/zanata/rest/ZanataRestResponseInterceptor.java
+++ b/zanata-war/src/main/java/org/zanata/rest/ZanataRestResponseInterceptor.java
@@ -29,34 +29,85 @@ import javax.ws.rs.container.ContainerResponseFilter;
 import javax.ws.rs.core.MultivaluedMap;
 import javax.ws.rs.ext.Provider;
 
+import com.google.common.collect.ImmutableList;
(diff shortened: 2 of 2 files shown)


seanf:
@carlosmunoz This PR is the same as https://github.com/zanata/zanata-server/pull/1149, but could you please review the docs change? An entry will be added to the release notes (from JIRA) as part of the release process.

carlosmunoz:
Docs look good, though maybe we should put the CORS content in its own sub-section.

Also, does this mean we no longer need to add an entry in the release notes for bugs which are fixed? (because they will get added as part of the release process?)

seanf:
> Also, does this mean we no longer need to add an entry in the release notes for bugs which are fixed? (because they will get added as part of the release process?)

That's my understanding. Right @definite ?

seanf:
> Docs look good, though maybe we should put the CORS content in its own sub-section.

What did you have in mind? A different file, a different position in this file, or simply one `#` instead of two for the heading?

seanf:
Note: a version of my CORS changes (in https://github.com/zanata/zanata-server/pull/1156) was copied into https://github.com/zanata/zanata-server/commit/3b67e4dd5156dfb77804ef5898e30d1a58197af2 (which is part of https://github.com/zanata/zanata-server/pull/1125) so we will have to watch the conflict resolution when they are merged together.

djansen-redhat:
This looks fine to me, with one minor observation.  The docs state that, when CORS is enabled, the header value "Access-Control-Allow-Methods: PUT,POST,DELETE,GET,OPTIONS" will be in the response header, but it was always returned in the response regardless of setting.  The other options were omitted correctly.
@seanf @aeng I do not know if this is significant, in any way, but it can (likely) be fixed easily by moving the .add line into the conditional.  Comments?

aeng:
Sounds to me its just depends on how strict we want it to be. With current implementation (always return Access-Control-Allow-Methods: PUT,POST,DELETE,GET,OPTIONS), it just a way of saying, 
`if user can start their browser with disabling CORS check, then they can perform those action in our rest endpoint`

If we only returns 'Access-Control-Allow-Methods: PUT,POST,DELETE,GET,OPTIONS' when its a whitelisted requester, then we are preventing any actions to be perform even when user starts their browser with disabling CORS check.

seanf:
> Sounds to me its just depends on how strict we want it to be. With current implementation (always return Access-Control-Allow-Methods: PUT,POST,DELETE,GET,OPTIONS), it just a way of saying, 
"if user can start their browser with disabling CORS check, then they can perform those action in our rest endpoint"

Surely if the browser disable CORS checks, this header becomes completely irrelevant.

> If we only returns 'Access-Control-Allow-Methods: PUT,POST,DELETE,GET,OPTIONS' when its a whitelisted requester, then we are preventing any actions to be perform even when user starts their browser with disabling CORS check.

If CORS checks are disabled, they're disabled, aren't they?

aeng:
> If CORS checks are disabled, they're disabled, aren't they?
Not sure, need QA to test this scenario if it matters.

seanf:
> This looks fine to me, with one minor observation. The docs state that, when CORS is enabled, the header value "Access-Control-Allow-Methods: PUT,POST,DELETE,GET,OPTIONS" will be in the response header, but it was always returned in the response regardless of setting. The other options were omitted correctly.

> @seanf @aeng I do not know if this is significant, in any way, but it can (likely) be fixed easily by moving the .add line into the conditional. Comments?

I don't know if it's significant either, but docs and code really ought to match, and we might as well change the code to do what the docs say. It seems like the safer option anyway, since if you don't specify the whitelist system property, you won't get any CORS headers.

I'll make the change.

aeng:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1156)*

seanf:
@djansen-redhat Could you retest please?

djansen-redhat:
:white_check_mark: Works well, as far as I can determine

PR #1153, opened 2016-05-04
fix(ZNTA-1060): avoid NPE in DuplicateParamFilter

seanf:
https://zanata.atlassian.net/browse/ZNTA-1060

PR #1160, opened 2016-05-06
fix(ZNTA-1072):fix selectManyCheckbox in filter

aeng:
https://zanata.atlassian.net/browse/ZNTA-1072

### DIFF
--- a/zanata-war/src/main/java/org/zanata/servlet/DuplicateParamFilter.java
+++ b/zanata-war/src/main/java/org/zanata/servlet/DuplicateParamFilter.java
@@ -81,7 +81,7 @@ public class DuplicateParamFilter implements Filter {
         public String[] getParameterValues(String paramName) {
             // avoid constructing the entire map unless requested
             String[] value = getRequest().getParameterValues(paramName);
-            return value == null ? null : new String[] { value[0] };
+            return value == null ? null : value;
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


huangp:
@seanf I think this will defeat the purpose of your duplicate param filter. Maybe we need to iterate through both name and values and only remove them if both are the same.

aeng:
@huangp I was thinking the same. This filter actually prevents grouped component in form, e.g. grouped checkboxes, group radio button. Whats the use case for this filter?

seanf:
> @seanf I think this will defeat the purpose of your duplicate param filter. 

@huangp You're quite right. The job of this class is to eliminate duplicate params introduced by a bug in ocpsoft rewrite. I was hoping we wouldn't need duplicate params for ourselves, but apparently we do.

> Maybe we need to iterate through both name and values and only remove them if both are the same.

I had considered doing that, but wanted to keep the filter simple if possible. Are we sure that this would fix the current problem? Or might it _need_ two identical name/value combinations?

I don't suppose it would be possible to avoid the use of multiple param values?

The only complete fix would be a fix for https://github.com/ocpsoft/rewrite/issues/223

seanf:
Also note the existence of https://github.com/zanata/zanata-server/pull/1153 which touches the same class. We're going to have conflicts.

seanf:
I'll update #1153 to deduplicate identical values for the same parameter name.

aeng:
Fixed in https://github.com/zanata/zanata-server/pull/1153.

PR #1142, opened 2016-04-25
Redux app reviewed

aeng:
Changed on reviewed branch of https://github.com/zanata/zanata-server/pull/1125

PR #1162, opened 2016-05-08
Redux app reviewed

davidmason:
Re-created #1142 to target the correct branch


frontend/src/main/web/src/components/Row.js
```diff
@@ -25,7 +25,10 @@ const classes = {
     }
   }
 }
-
+/**
+ * A flexbox component that align all elements vertically in a row.
```
davidmason: - [ ] "vertically" indicates that everything will be stacked up on top of each other, is that right?
- [ ] is "elements" talking about the elements inside it (children), or talking about if I have multiple `<Row>` elements next to each other in a container?

PR #1169, opened 2016-05-12
fix(ZNTA-1073) : create custom exceptions + handlers

broccolinisoup:
Create ProjectNotFoundException and VersionNotFoundException and throw them when trying to write invalid project or iteration slug on project view or iteration view pages. Handle these exceptions and return warn page with the proper warning message.

Jira ticket : https://zanata.atlassian.net/browse/ZNTA-1073

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1169)

### DIFF
--- a/zanata-war/src/main/java/org/zanata/action/ProjectHome.java
+++ b/zanata-war/src/main/java/org/zanata/action/ProjectHome.java
@@ -61,6 +61,7 @@ import org.zanata.dao.AccountRoleDAO;
 import org.zanata.dao.LocaleDAO;
 import org.zanata.dao.PersonDAO;
 import org.zanata.dao.WebHookDAO;
+import org.zanata.exception.ProjectNotFoundException;
 import org.zanata.i18n.Messages;
 import org.zanata.model.HAccount;
 import org.zanata.model.HAccountRole;
@@ -688,7 +689,7 @@ public class ProjectHome extends SlugHome<HProject> implements
             log.warn(
                     "Project [id={}, slug={}], does not exist or is soft deleted: {}",
                     projectId, getSlug(), project);
-            throw new EntityNotFoundException();
+            throw new ProjectNotFoundException();
         }
     }
 
--- a/zanata-war/src/main/java/org/zanata/action/ProjectHome.java
+++ b/zanata-war/src/main/java/org/zanata/action/ProjectHome.java
@@ -689,7 +689,7 @@ public class ProjectHome extends SlugHome<HProject> implements
             log.warn(
                     "Project [id={}, slug={}], does not exist or is soft deleted: {}",
                     projectId, getSlug(), project);
-            throw new ProjectNotFoundException();
+            throw new ProjectNotFoundException(getSlug());
         }
     }
 
--- a/zanata-war/src/main/java/org/zanata/exception/ProjectNotFoundException.java
+++ b/zanata-war/src/main/java/org/zanata/exception/ProjectNotFoundException.java
@@ -1,0 +1,14 @@
+package org.zanata.exception;
+
+public class ProjectNotFoundException extends RuntimeException {
+    private static final long serialVersionUID = -7920694286L;
+
+    public ProjectNotFoundException() {
+        super();
+    }
+
+    public ProjectNotFoundException(String message) {
+        super(message);
+    }
+
+}
--- a/zanata-war/src/main/java/org/zanata/exception/ProjectNotFoundException.java
+++ b/zanata-war/src/main/java/org/zanata/exception/ProjectNotFoundException.java
@@ -1,3 +1,23 @@
+/*
+ * Copyright 2016, Red Hat, Inc. and individual contributors as indicated by the
+ * @author tags. See the copyright.txt file in the distribution for a full
+ * listing of individual contributors.
+ *
+ * This is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation; either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * This software is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this software; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA, or see the FSF
+ * site: http://www.fsf.org.
+ */
 package org.zanata.exception;
 
 public class ProjectNotFoundException extends RuntimeException {
--- a/zanata-war/src/main/java/org/zanata/exception/ProjectNotFoundException.java
+++ b/zanata-war/src/main/java/org/zanata/exception/ProjectNotFoundException.java
@@ -1,14 +1,14 @@
 package org.zanata.exception;
 
 public class ProjectNotFoundException extends RuntimeException {
-    private static final long serialVersionUID = -7920694286L;
+    private static final long serialVersionUID = 1L;
+    private final String projectSlug;
 
-    public ProjectNotFoundException() {
-        super();
+    public ProjectNotFoundException(String projectSlug) {
+        this.projectSlug = projectSlug;
     }
 
-    public ProjectNotFoundException(String message) {
-        super(message);
+    public String getProjectSlug() {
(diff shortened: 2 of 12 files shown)


broccolinisoup:
@aeng or @carlosmunoz could you review it, please?

zanata-war/src/main/java/org/zanata/action/ProjectHome.java
```diff
@@ -688,7 +689,7 @@ private void validateProjectState(HProject project) {
             log.warn(
                     "Project [id={}, slug={}], does not exist or is soft deleted: {}",
                     projectId, getSlug(), project);
-            throw new EntityNotFoundException();
+            throw new ProjectNotFoundException();
```
seanf: What's wrong with EntityNotFoundException for this sort of thing? If we really need to introduce ProjectNotFoundException and VersionNotFoundException, they should probably extend EntityNotFoundException, since they're just a specialisation of the same idea.

seanf:
The page warn.xhtml seems to be doing too much.

If we plan to redirect to it when there is a link to a non-existent project, that's more than a warning, that's really an error page. Also warn.xhtml includes strings like jsf.Error and jsf.ErrorTitle, which means it will look like an error page.

How are the project slug, version project slug and version slugs used? I don't think it makes sense to have these very specific slugs used on a generic error/warning page. Perhaps it would be better to split it into two pages, eg iteration/missing.xhtml and project/missing.xhtml.

zanata-war/src/main/java/org/zanata/servlet/UrlRewriteConfig.java
```diff
@@ -66,7 +66,7 @@ public Configuration getConfiguration(final ServletContext context) {
 
                 .addRule(Join.path("/error/").to("/error.xhtml"))
                 .addRule(Join.pathNonBinding("/error/{path}").to("/error/{path}.xhtml"))
-                .addRule(Join.path("/warn").to("/warn.xhtml"))
+//                .addRule(Join.path("/error/missing_entity").to("/error/missing_entity.xhtml"))
```
seanf: - [ ] If the missing entity error page is working without this line, we shouldn't need the comment.

zanata-war/src/main/java/org/zanata/exception/ProjectNotFoundException.java
```diff
@@ -0,0 +1,14 @@
+package org.zanata.exception;
```
seanf: - [ ] If you haven't, you should set up your IDE to add header comments (LGPL, Red Hat 2016, etc) for new classes.  (Also VersionNotFoundException.java).

zanata-war/src/main/java/org/zanata/exception/handler/ProjectNotFoundExceptionHandler.java
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright 2015, Red Hat, Inc. and individual contributors as indicated by the
```
seanf: - [ ] For new files, this should be 2016. (Also VersionNotFoundExceptionHandler.java).

seanf:
Looks good now; just need to clean up some of the comments in the code (see my other comments).

carlosmunoz:
@seanf I think this is ready for another round of review

seanf:
Reviewed 1 of 12 files at r1, 7 of 11 files at r2, 4 of 4 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/servlet/UrlRewriteConfig.java, line 69 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1169#-KI0VcjsMv9arQkcsF6i-r2-69:-KI70nhFGyZUlU4tT9Ot:754032608) ([raw file](https://github.com/zanata/zanata-server/blob/97380584947bb24626286cf87fac78970daacdcf/zanata-war/src/main/java/org/zanata/servlet/UrlRewriteConfig.java#L69)):*
> - [ ] If the missing entity error page is working without this line, we shouldn't need the comment.
</details>
Looks like this is not needed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1169)*

broccolinisoup:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/exception/ProjectNotFoundException.java, line 1 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1169#-KI0VcjsMv9arQkcsF6l-r2-1:-KIWCnaor6LRkxJEx8JZ:1734371038) ([raw file](https://github.com/zanata/zanata-server/blob/97380584947bb24626286cf87fac78970daacdcf/zanata-war/src/main/java/org/zanata/exception/ProjectNotFoundException.java#L1)):*
> - [x] If you haven't, you should set up your IDE to add header comments (LGPL, Red Hat 2016, etc) for new classes.  (Also VersionNotFoundException.java).
</details>
I set up my header comments

---

*[zanata-war/src/main/java/org/zanata/exception/handler/ProjectNotFoundExceptionHandler.java, line 2 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1169#-KI0VcjsMv9arQkcsF6n-r2-2:-KIWCi3vtfVyZox94rfs:1734371038) ([raw file](https://github.com/zanata/zanata-server/blob/97380584947bb24626286cf87fac78970daacdcf/zanata-war/src/main/java/org/zanata/exception/handler/ProjectNotFoundExceptionHandler.java#L2)):*
> - [x] For new files, this should be 2016. (Also VersionNotFoundExceptionHandler.java).
</details>
I set up my header comments

---

*[zanata-war/src/main/java/org/zanata/servlet/UrlRewriteConfig.java, line 69 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1169#-KI0VcjsMv9arQkcsF6i-r2-69:-KIWBuA1hwS-xrQADc6c:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/97380584947bb24626286cf87fac78970daacdcf/zanata-war/src/main/java/org/zanata/servlet/UrlRewriteConfig.java#L69)):*
> Looks like this is not needed.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1169)*

seanf:
Reviewed 1 of 1 files at r4.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1169)*

djansen-redhat:
*[zanata-war/src/main/resources/messages.properties, line 18 \[r4\]](https://reviewable.io:443/reviews/zanata/zanata-server/1169#-KJ935Kq6i31KJsi4QGD:-KJ935Kq6i31KJsi4QGE:-1816383313) ([raw file](https://github.com/zanata/zanata-server/blob/5761522a241c8cf516368b0156cc8084acfa2ec8/zanata-war/src/main/resources/messages.properties#L18)):*
> ```INI
> jsf.RecordNotFound=Record not found
> jsf.ProjectNotFound = Project "{0}" not found
> jsf.VersionNotFound = Project "{0}" or version "{1}" not found
> ```

Could this be better written as "Version "{1}" for project "{0}" not found"?
If the project is not found, then that is encountered before this so the 'or' is not accurate, I think. (Just my thought, when tested).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1169)*

broccolinisoup:
Review status: 9 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[zanata-war/src/main/resources/messages.properties, line 18 \[r4\]](https://reviewable.io:443/reviews/zanata/zanata-server/1169#-KJ935Kq6i31KJsi4QGD:-KJDUEwenmDv1Ie0iLRu:1706342512) ([raw file](https://github.com/zanata/zanata-server/blob/5761522a241c8cf516368b0156cc8084acfa2ec8/zanata-war/src/main/resources/messages.properties#L18)):*
<details><summary><i>Previously, djansen-redhat (Damian Jansen) wrote…</i></summary>
> Could this be better written as "Version "{1}" for project "{0}" not found"?
> If the project is not found, then that is encountered before this so the 'or' is not accurate, I think. (Just my thought, when tested).
</details>
Done!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1169)*

seanf:
Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1169)*

djansen-redhat:
:white_check_mark: verified

PR #1191, opened 2016-06-02
fix(po file): Fix po-revision-date header

aeng:
https://zanata.atlassian.net/browse/ZNTA-1145

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1191)


huangp:
:lgtm:

<details><summary><i>Previously, aeng (Alex Eng) wrote…</i></summary>
> #### fix\(po file\): Fix po\-revision\-date header
> 
> https://zanata.atlassian.net/browse/ZNTA-1145
> 
> 
> 
</details>


---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1191#-:-KJIYqJ5o-rVw-F2FP96:1417088721)*

seanf:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java, line 885 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1191#-KJIgPi4pbwDUqa7Ku3G:-KJIgPi4pbwDUqa7Ku3H:-124808480) ([raw file](https://github.com/zanata/zanata-server/blob/06c0a18e3b7d7923ec02b7fc557480887a25dc58/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java#L885)):*
> ```Java
>         HTextFlowTarget lastChanged = null;
> 
>         for (HTextFlowTarget tft : translations) {
> ```

This loop could be expensive. How often will we be doing this? Should we be caching the result somewhere?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1191)*

aeng:

---

*[zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java, line 885 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1191#-KJIgPi4pbwDUqa7Ku3G:-KJIh6hAkTRk0KSe4_n3:-353623721) ([raw file](https://github.com/zanata/zanata-server/blob/06c0a18e3b7d7923ec02b7fc557480887a25dc58/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java#L885)):*
> This loop could be expensive. How often will we be doing this? Should we be caching the result somewhere?
</details>
I don't think we should be caching anything in util class. This only happens once for each document.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1191)*

seanf:

---

*[zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java, line 885 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1191#-KJIgPi4pbwDUqa7Ku3G:-KJIiySL_D3eD619R_Nz:256855213) ([raw file](https://github.com/zanata/zanata-server/blob/06c0a18e3b7d7923ec02b7fc557480887a25dc58/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java#L885)):*
<details><summary><i>Previously, aeng (Alex Eng) wrote…</i></summary>
> I don't think we should be caching anything in util class. This only happens once for each document.
</details>
I didn't say where it should be cached. Maybe it would belong in HTranslatedDocument, for instance (which we haven't created yet) or HPoTargetHeader. 

Is that once for each document, each time it is pulled in PO form? Each time it is pulled in any form?  Either way, that will add up.

And will our etags still be accurate?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1191)*

aeng:

---

*[zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java, line 885 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1191#-KJIgPi4pbwDUqa7Ku3G:-KJImd25Os-iCxIAa9of:1339182373) ([raw file](https://github.com/zanata/zanata-server/blob/06c0a18e3b7d7923ec02b7fc557480887a25dc58/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java#L885)):*
> I didn't say where it should be cached. Maybe it would belong in HTranslatedDocument, for instance (which we haven't created yet) or HPoTargetHeader. 
> 
> Is that once for each document, each time it is pulled in PO form? Each time it is pulled in any form?  Either way, that will add up.
> 
> And will our etags still be accurate?
</details>
It happens once for each document in a locale, during zanata-pull, or download.

Our eTag generation uses 'lastChanged' in tft, which i didn't modify. What I updated was only the string used to generate the po file header.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1191)*

seanf:
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java, line 885 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1191#-KJIgPi4pbwDUqa7Ku3G:-KJIqRFU6sOy-puNwo8a:1572202454) ([raw file](https://github.com/zanata/zanata-server/blob/06c0a18e3b7d7923ec02b7fc557480887a25dc58/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java#L885)):*
<details><summary><i>Previously, aeng (Alex Eng) wrote…</i></summary>
> It happens once for each document in a locale, during zanata-pull, or download.
> 
> Our eTag generation uses 'lastChanged' in tft, which i didn't modify. What I updated was only the string used to generate the po file header.
> 
</details>
@aeng, so, the extra cost is incurred even for clients which don't care about PO headers, eg Properties files? 

If we're not going to store/cache this information somewhere, we should at least ensure that only PO users have to pay the costs of PO header generation. Large documents are already quite slow, without adding another loop through the data.

BTW, the cost of gathering this timestamp should be a lot lower (for large docs) if it were implemented in SQL/HQL, instead of iterating through all those Hibernate objects in memory. But I see that getLastTranslatedTarget is in the same boat, which means it must be bad for performance too. 

At any rate, if we're going to iterate through Hibernate objects in memory, we should only do it once, so it would make sense to combine the work of getLastTranslatedTarget with getLastChangedTarget into one iteration.

My point about the etag header is that we shouldn't change the returned data (including headers) without also changing the etag.  (Just like you should generally update .equals and .hashCode in Java together.) Fortunately, the etag includes an aggregation of all the TFT's lastChanged values, so if the timestamp changes, the etag will too. 

*Except* when this change is first deployed: the new PO header will appear, but the etag will not (unless the document has actually changed). Anyone who cares about this change may need to purge client-side caches to ensure they get the updated header value.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1191)*

seanf:
Reviewed 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java, line 895 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1191#-KJJyWDyhFfKh-O8uQKp:-KJJyWDyhFfKh-O8uQKq:1394070982) ([raw file](https://github.com/zanata/zanata-server/blob/c6f627928da2dc51532393219a340e249c467b50/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java#L895)):*
> ```Java
>                 if ((lastModifiedDate != null &&
>                     lastModifiedDate.after(headerRevisionDate) ||
>                     (lastModifiedDate == null))) {
> ```

There *must* be a way of simplifying that boolean mess. If you put the `lastModifiedDate == null` clause first, you won't need to check `lastModifiedDate != null` as well.

---

*[zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java, line 912 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1191#-KJJy2rTMolLZ6YWIeEQ:-KJJy2rTMolLZ6YWIeER:1955935386) ([raw file](https://github.com/zanata/zanata-server/blob/c6f627928da2dc51532393219a340e249c467b50/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java#L912)):*
> ```Java
>                  */
>                 lastTranslator = generateLastTranslator("Zanata",
>                     applicationConfiguration.getFromEmailAddr());
> ```

Crediting the Zanata server administrator (or mailing list) as a translator for a document which has copied translations from someone else could lead to trouble.  I really don't think we want to do that.

We should probably use an obviously artificial email address here, if we must have an email address at all. I don't know, maybe something like copied-by-zanata@<whatever zanata's hostname is>, with a name like "Copied by Zanata".

---

*[zanata-war/src/test/java/org/zanata/rest/service/raw/TranslationResourceRestITCase.java, line 657 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1191#-KJJzqF3reQUTcY5UUHe:-KJJzqF3reQUTcY5UUHf:792419183) ([raw file](https://github.com/zanata/zanata-server/blob/c6f627928da2dc51532393219a340e249c467b50/zanata-war/src/test/java/org/zanata/rest/service/raw/TranslationResourceRestITCase.java#L657)):*
> ```Java
> 
>         // Since it is a first push with no headers, the Last Translator and
>         // Last Revision Date header should be empty
> ```

This comment no longer matches the code.

---

*[zanata-war/src/test/java/org/zanata/rest/service/raw/TranslationResourceRestITCase.java, line 662 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1191#-KJJzsP4z4G-3MyNH8dK:-KJJzsP4z4G-3MyNH8dL:-1739658200) ([raw file](https://github.com/zanata/zanata-server/blob/c6f627928da2dc51532393219a340e249c467b50/zanata-war/src/test/java/org/zanata/rest/service/raw/TranslationResourceRestITCase.java#L662)):*
> ```Java
>             if (entry.getKey().equals(HeaderFields.KEY_LastTranslator)
>                     || entry.getKey().equals(HeaderFields.KEY_PoRevisionDate)) {
>                 assertThat(entry.getValue().trim(), not(isEmptyOrNullString()));
> ```

Not empty or null? So what is it? That's probably a bit too loose.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1191)*

aeng:
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java, line 895 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1191#-KJJyWDyhFfKh-O8uQKp:-KJXudl9bDcFvaIbFpfq:-1938606079) ([raw file](https://github.com/zanata/zanata-server/blob/c6f627928da2dc51532393219a340e249c467b50/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java#L895)):*
> There *must* be a way of simplifying that boolean mess. If you put the `lastModifiedDate == null` clause first, you won't need to check `lastModifiedDate != null` as well.
</details>
I'm not sure what you mean in here.  The condition should be 
if lastModifiedDate == null,  then generateLastTranslator OR
if lastModifiedDate != null, it need to be after  headerRevisionDate, then generateLastTranslator

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1191)*

djansen-redhat:
:white_check_mark: tested.

PR #1204, opened 2016-06-14
feat: add docker development images

carlosmunoz:
Add images, Dockerfile, scripts and documentation to be able to quickly run docker containers for Zanata development.

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1204)

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,3 +57,7 @@ useful for starting a Zanata instance with the aim of running functional tests
 from an IDE.
 
 The `-h` argument prints the script's help.
+
+#### Development using docker (experimental)
+
+For a quick Zanata development environment with Docker, please visit the [docker README](docker/README.md).
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,23 @@
+FROM jboss/wildfly:10.0.0.Final
+
+USER root
+RUN yum -y install mysql-connector-java wget && yum clean all
+
+RUN ln -sf /usr/share/java/mysql-connector-java.jar /opt/jboss/wildfly/standalone/deployments/mysql-connector-java.jar
+
+# Wildfly modules
+
+RUN wget --no-verbose https://dl.bintray.com/sflanigan/generic/wildfly-module-hibernate-main-4.2.20.Final.zip \
+    && unzip -q wildfly-module-hibernate-main-4.2.20.Final.zip -d /opt/jboss/wildfly
+
+ENV ZANATA_VERSION 3.9.0
+
+# Zanata war
+RUN wget --no-verbose -O /opt/jboss/wildfly/standalone/deployments/ROOT.war \
+    https://github.com/zanata/zanata-server/releases/download/server-${ZANATA_VERSION}/zanata-war-${ZANATA_VERSION}.war \
+    && chmod o+rw /opt/jboss/wildfly/standalone/deployments/ROOT.war
+
+# standalone.xml
+ADD conf/standalone.xml /opt/jboss/wildfly/standalone/configuration/
+
+USER jboss
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -10,14 +10,14 @@ RUN ln -sf /usr/share/java/mysql-connector-java.jar /opt/jboss/wildfly/standalon
 RUN wget --no-verbose https://dl.bintray.com/sflanigan/generic/wildfly-module-hibernate-main-4.2.20.Final.zip \
     && unzip -q wildfly-module-hibernate-main-4.2.20.Final.zip -d /opt/jboss/wildfly
 
-ENV ZANATA_VERSION 3.9.0
+# ENV ZANATA_VERSION 3.9.0
 
-# Zanata war
-RUN wget --no-verbose -O /opt/jboss/wildfly/standalone/deployments/ROOT.war \
-    https://github.com/zanata/zanata-server/releases/download/server-${ZANATA_VERSION}/zanata-war-${ZANATA_VERSION}.war \
-    && chmod o+rw /opt/jboss/wildfly/standalone/deployments/ROOT.war
-
-# standalone.xml
-ADD conf/standalone.xml /opt/jboss/wildfly/standalone/configuration/
+# Zanata war (this is not needed for dev)
+# RUN wget --no-verbose -O /opt/jboss/wildfly/standalone/deployments/ROOT.war \
+#     https://github.com/zanata/zanata-server/releases/download/server-${ZANATA_VERSION}/zanata-war-${ZANATA_VERSION}.war \
+#     && chmod o+rw /opt/jboss/wildfly/standalone/deployments/ROOT.war
 
 USER jboss
+
+# standalone.xml (not needed for dev)
+# ADD conf/standalone.xml /opt/jboss/wildfly/standalone/configuration/
--- a/docker/README.md
+++ b/docker/README.md
@@ -6,15 +6,21 @@ This image has been tested with Docker 1.11
 
 ## Building
 
-To build this docker image, simply type the following command:
+To build this docker image as "latest", simply type the following command:
 
 ```sh
-$ docker build -t zanata/server:<zanata version> .
+$ docker build -t zanata/server .
 ```
 
-The `-t` parameter indicates a tag for the image and is optional. The version is also optional, if not given, the build will be tagged as 'latest'.
+If you want to build with a different tag, eg 3.8.4, use a command like this:
 
-_Also note the '.' at the end of the command_
+```sh
+$ docker build -t zanata/server:3.8.4 .
+```
+
+Be careful about overwriting existing versions.
+
+The `-t` parameter indicates the name and/or tag for the image.
 
 ## Running a Zanata server with Docker
 
@@ -36,7 +42,13 @@ The database can be accessed via tcp via the `mysql` command or by using any dat
 mysql --protocol=tcp -h localhost --port=<PORT> -uzanata -pzanatapw zanata
 ```
 
-The username and password above are the default given by the script. If you wish to change them, edit the script file with your preferred credentials.
+Alternatively, a docker container can be used to connect to the database using the mysql client like this:
+
+```sh
+$ docker exec -i zanatadb mysql -uzanata -pzanatapw zanata
+```
+
+The username and password above are the default given by the `rundb.sh` script. If you wish to change them, edit the script file with your preferred credentials.
 
 _Note: The server container will be called `zanatadb`. Once stopped, docker will prevent the `rundb.sh` script from running again until the stopped container is removed. To do this, just type:_
 
@@ -58,7 +70,7 @@ To start the Zanata server run:
 $ ./rundev.sh
 ```
 
-This script will start another docker container with the Zanata server, and will log the server output. Contrary to the database container, the server container will run in daemon mode.
+This script will start another docker container with the Zanata server, and will log the server output. Unlike the database container, the server container will run in daemon mode.
 
 The server will connect to the db server which was started in the previous step. It will also take the zanata.war from any war file in the `zanata-war/target` directory. This means a war file must be built beforehand (See `etc/scripts/quickbuild.sh`). You must make sure there is only one `zanata-*.war` file in this directory, otherwise this step will fail.
 
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,92 @@
+This is a docker image for the Zanata server. This image runs a wildfly server with Zanata running on top of it. It does not run a database server for the Zanata application
+
+```
+This image has been tested with Docker 1.11
+```
+
+## Building
+
+To build this docker image, simply type the following command:
+
+```sh
+$ docker build -t zanata/server:<zanata version> .
+```
+
+The `-t` parameter indicates a tag for the image and is optional. The version is also optional, if not given, the build will be tagged as 'latest'.
+
+_Also note the '.' at the end of the command_
+
+## Running a Zanata server with Docker
+
+### Run a database server container
+
+Simply run
+
+```sh
+$ ./rundb.sh
+```
+
+This script will start a docker container with the database. You can inspect the script file to learn the exact docker command it's running.
+
+The container will map the mysql data directory to `$HOME/docker-volumes/zanata-mariadb`. This can be changed from the script file.
+
+The database can be accessed via tcp via the `mysql` command or by using any database administration tool. You need to get the actual mapped port on the host by typing `docker ps`. After you have the port, you can connect locally to the database with the following command:
+
+```sh
+mysql --protocol=tcp -h localhost --port=<PORT> -uzanata -pzanatapw zanata
+```
+
+The username and password above are the default given by the script. If you wish to change them, edit the script file with your preferred credentials.
+
+_Note: The server container will be called `zanatadb`. Once stopped, docker will prevent the `rundb.sh` script from running again until the stopped container is removed. To do this, just type:_
+
+```sh
+$ docker rm zanatadb
+```
+
+If you don't remove the container, and later want to simply restart this same instance, you can simply run
+
+```sh
+$ docker start zanatadb
+```
+
+## Run the Zanata development server container
+
+To start the Zanata server run:
+
+```sh
+$ ./rundev.sh
+```
+
+This script will start another docker container with the Zanata server, and will log the server output. Contrary to the database container, the server container will run in daemon mode.
+
+The server will connect to the db server which was started in the previous step. It will also take the zanata.war from any war file in the `zanata-war/target` directory. This means a war file must be built beforehand (See `etc/scripts/quickbuild.sh`). You must make sure there is only one `zanata-*.war` file in this directory, otherwise this step will fail.
+
+This container will have a mapped volume to your `$HOME/zanata` directory to store files, stats, caches, etc. This will allow for backups if you wish to switch between different versions of zanata for instance.
+
+## Create an admin user
+
+_This step is only required for empty databases. Since the `rundb.sh` script creates a mapped volume with all the mariadb data, it's not needed for subsequent runs._
+
+At this point, you have a running Zanata instance... with no users created (!)
+
+To create an admin user, you can connect to the database server and run the script at `/conf/admin-user-setup.sql` like this:
+
+```sh
+$ docker exec -i zanatadb  mysql -uzanata -pzanatapw zanata < conf/admin-user-setup.sql
+```
+
+_Note the -u and -p parameters to the mysql command must be followed by the database username and password as indicated in the rundb.sh file. The values above are the default values in the script._
+
+This will create a username `admin` with password `admin1234`
+
+## Access Zanata
+
+The docker host will forward port 8080 to the container's 8080 port where Wildfly is listening.
+
+Just open a browser and head to `http://IP_ADDRESS:8080`, where `IP_ADDRESS` is the docker host's address (again, this might localhost or some other IP address if running on OSX for example).
+
+## DockerHub
+
+The DockerHub site for Zanata images is here:
+https://hub.docker.com/r/zanata/server/
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,7 +1,7 @@
 This is a docker image for the Zanata server. This image runs a wildfly server with Zanata running on top of it. It does not run a database server for the Zanata application
 
 ```
-This image has been tested with Docker 1.11
+This image has been tested with Docker 1.9
 ```
 
 ## Building
@@ -36,7 +36,7 @@ This script will start a docker container with the database. You can inspect the
 
 The container will map the mysql data directory to `$HOME/docker-volumes/zanata-mariadb`. This can be changed from the script file.
 
-The database can be accessed via tcp via the `mysql` command or by using any database administration tool. You need to get the actual mapped port on the host by typing `docker ps`. After you have the port, you can connect locally to the database with the following command:
+The database can be accessed via tcp via the `mysql` command or by using any database administration tool. You need to get the actual mapped port on the host by typing `docker ps`. After you have the port, you can connect locally to the database. The following is an example to accomplish this using a locally installed mysql client:
 
 ```sh
 mysql --protocol=tcp -h localhost --port=<PORT> -uzanata -pzanatapw zanata
--- a/docker/conf/admin-user-setup.sql
+++ b/docker/conf/admin-user-setup.sql
@@ -1,0 +1,9 @@
+-- Username: admin
+-- Password: admin1234
+INSERT INTO HAccount (id, creationDate, lastChanged, versionNum, enabled, passwordHash, username)
+VALUES (1, now(), now(), 1, 1, 'UZMf4PIqtTBGAo9wWKuTpg==', 'admin');
+
+INSERT INTO HPerson (id, creationDate, lastChanged, versionNum, email, `name`, accountId)
+VALUES (1, now(), now(), 1, 'admin@noreply.org', 'Administrator', 1);
+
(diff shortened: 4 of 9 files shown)


seanf:
Reviewed 9 of 9 files at r1.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[docker/Dockerfile, line 18 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCfKMgvLC6vlic-cSL:-KKCfKMhkMJI5SlYmXaT:1770150831) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/Dockerfile#L18)):*
> ```Dockerfile
> RUN wget --no-verbose -O /opt/jboss/wildfly/standalone/deployments/ROOT.war \
>     https://github.com/zanata/zanata-server/releases/download/server-${ZANATA_VERSION}/zanata-war-${ZANATA_VERSION}.war \
>     && chmod o+rw /opt/jboss/wildfly/standalone/deployments/ROOT.war
> ```

Technically speaking, this bit is slow and useless for development purposes, because we will be replacing ROOT.war when launching the container. It makes sense for a production image though.

When will ZANATA_VERSION be updated? After all, this branch is for 4.0.0-SNAPSHOT.

How will this tie into the release process?

---

*[docker/init-db.sh, line 3 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCf69ilnnxkAJLFFP0:-KKCf69ilnnxkAJLFFP1:284703544) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/init-db.sh#L3)):*
> ```Shell
> docker run -it \
>     --link zanatadb:mysql --rm mariadb sh \
>     -c 'exec mysql --protocol=tcp -h"$MYSQL_PORT_3306_TCP_ADDR" -P"$MYSQL_PORT_3306_TCP_PORT" -uzanata -ppassword zanata > conf/admin-user-setup.sql'
> ```

Looks like this command is broken and obsolete.

---

*[docker/README.md, line 4 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCgzpBvqExmx9HoBSW:-KKCgzpBvqExmx9HoBSX:-1871668815) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L4)):*
> ```Markdown
> 
> ```
> This image has been tested with Docker 1.11
> ```

Is that the version for Fedora 23? 

Don't we want to stick to 1.8? I think RHEL 7 has Docker 1.8, as does Fedora 22.

---

*[docker/README.md, line 15 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCgXnchBxCKiWLK7sK:-KKCgXnchBxCKiWLK7sL:489730825) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L15)):*
> ```Markdown
> ```
> 
> The `-t` parameter indicates a tag for the image and is optional. The version is also optional, if not given, the build will be tagged as 'latest'.
> ```

I have different, hopefully better, text for the docker build and docker run commands, which apparently didn't make the trip over from the docker-images repo.

See https://github.com/zanata/docker-images/compare/f7d9e81ac41c31d66c2b1c87ac5edc059534f347...fbf93861079420677fec19429846089b95c44f7c

---

*[docker/README.md, line 36 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCi1alOzQfeWV1E7VM:-KKCi1alOzQfeWV1E7VN:-1526975336) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L36)):*
> ```Markdown
> 
> ```sh
> mysql --protocol=tcp -h localhost --port=<PORT> -uzanata -pzanatapw zanata
> ```

docker ps can be used to show the random port number which was chosen.

But I would recommend an approach based on docker exec zanatadb... since it will use the mysql binary/version from the database container. As used for admin-user-setup.sql below.

---

*[docker/README.md, line 39 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCiM-1hmt0QW9kFccK:-KKCiM-2tdcQ9o4RtxTu:-1583755743) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L39)):*
> ```Markdown
> ```
> 
> The username and password above are the default given by the script. If you wish to change them, edit the script file with your preferred credentials.
> ```

given by the rundb.sh script

---

*[docker/README.md, line 61 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCiRb2Lz4bJWTYODG3:-KKCiRb2Lz4bJWTYODG4:720744674) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L61)):*
> ```Markdown
> ```
> 
> This script will start another docker container with the Zanata server, and will log the server output. Contrary to the database container, the server container will run in daemon mode.
> ```

I would suggest "Unlike", not "Contrary".

---

*[docker/runapp.sh, line 3 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCiku8aosdEvS7-s8d:-KKCiku8aosdEvS7-s8e:1874290976) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/runapp.sh#L3)):*
> ```Shell
> #!/bin/sh
> docker run --name zanata --link zanatadb:db -p 8080:8080 -it \
>     -v $PWD/volumes/zanata:/opt/jboss/zanata \
> ```

This volume directory looks wrong, but we probably just want to delete this file for now.

---

*[docker/rundev.sh, line 2 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCj-iDg-bL0ddlQKHG:-KKCj-iDg-bL0ddlQKHH:744359038) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/rundev.sh#L2)):*
> ```Shell
> #!/bin/bash
> SOURCE="${BASH_SOURCE[0]}"
> ```

We should add a comment above, eg:

    # determine directory containing this script

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1204)*

carlosmunoz:
Review status: 5 of 7 files reviewed at latest revision, 9 unresolved discussions.

---

*[docker/Dockerfile, line 18 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCfKMgvLC6vlic-cSL:-KKL_7RQXFqoSSd487Du:2142947688) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/Dockerfile#L18)):*
> Technically speaking, this bit is slow and useless for development purposes, because we will be replacing ROOT.war when launching the container. It makes sense for a production image though.
> 
> When will ZANATA_VERSION be updated? After all, this branch is for 4.0.0-SNAPSHOT.
> 
> How will this tie into the release process?
</details>
I agree that this is not needed for development purposes. Having said that, do we want a separate dockerfile for the dev instance?

This would only need to be built rarely, but when it does it will download an unnecessary war file.

---

*[docker/README.md, line 4 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCgzpBvqExmx9HoBSW:-KKLdAPlk2id4F1Cg9g3:-2093652079) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L4)):*
> Is that the version for Fedora 23? 
> 
> Don't we want to stick to 1.8? I think RHEL 7 has Docker 1.8, as does Fedora 22.
</details>
Mmm.... I'll disagree here. There are ways to install the latest docker version for every system. I don't mind if we test and perhaps note if these images work with docker 1.8, but why tie them down if there are ways to run them with the latest?

The other point is the MariaDB image (for instance) is tested against 1.11, and they say support for lower versions is 'best-effort'. I say we follow this approach.

---

*[docker/README.md, line 15 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCgXnchBxCKiWLK7sK:-KKLee2MO4d7ci7rpUTo:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L15)):*
> I have different, hopefully better, text for the docker build and docker run commands, which apparently didn't make the trip over from the docker-images repo.
> 
> See https://github.com/zanata/docker-images/compare/f7d9e81ac41c31d66c2b1c87ac5edc059534f347...fbf93861079420677fec19429846089b95c44f7c
</details>
Done.

---

*[docker/README.md, line 36 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCi1alOzQfeWV1E7VM:-KKLf7u7Rj3EslqpBIHb:1909992389) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L36)):*
> docker ps can be used to show the random port number which was chosen.
> 
> But I would recommend an approach based on docker exec zanatadb... since it will use the mysql binary/version from the database container. As used for admin-user-setup.sql below.
</details>
My point here was to show how other clients can connect to the database, with things like DBeaver in mind. The mysql client is just an example which is easy to illustrate without a screenshot. I will add an alternative example showing how to do this.

---

*[docker/README.md, line 39 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCiM-1hmt0QW9kFccK:-KKLg-LOt_m-hOsWkjxn:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L39)):*
> given by the rundb.sh script
</details>
Done.

---

*[docker/README.md, line 61 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCiRb2Lz4bJWTYODG3:-KKLgDV8DPBerj2iXfdv:-1438446925) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L61)):*
> I would suggest "Unlike", not "Contrary".
</details>
Done. :+1:

---

*[docker/rundev.sh, line 2 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCj-iDg-bL0ddlQKHG:-KKLgZ3VRSdXLfiAqokm:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/rundev.sh#L2)):*
> We should add a comment above, eg:
> 
>     # determine directory containing this script
</details>
Done.

---

*[docker/init-db.sh, line 3 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCf69ilnnxkAJLFFP0:-KKLehXAPhNTh14s9Mb4:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/init-db.sh#L3)):*
> Looks like this command is broken and obsolete.
</details>
Done.

---

*[docker/runapp.sh, line 3 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCiku8aosdEvS7-s8d:-KKLgOgtBgK9DLmphOyn:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/runapp.sh#L3)):*
> This volume directory looks wrong, but we probably just want to delete this file for now.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1204)*

seanf:
Reviewed 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[docker/Dockerfile, line 18 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCfKMgvLC6vlic-cSL:-KKLnrze42mZWDbOGAzt:-1718828075) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/Dockerfile#L18)):*
> I agree that this is not needed for development purposes. Having said that, do we want a separate dockerfile for the dev instance?
> 
> This would only need to be built rarely, but when it does it will download an unnecessary war file.
</details>
If the intention is to use this for Dev only for now, I would suggest we optimise for that now, and sort out the prod version a bit later.

---

*[docker/Dockerfile, line 21 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKLxL5nDo8WHljcWRC3:-KKLxL5o_gfj2V1hZQel:-1364489710) ([raw file](https://github.com/zanata/zanata-server/blob/c855941428274f451906b83e19fe0bd505f9ab04/docker/Dockerfile#L21)):*
> ```Dockerfile
> 
> # standalone.xml
> ADD conf/standalone.xml /opt/jboss/wildfly/standalone/configuration/
> ```

For now, I suggest we comment out the above few instructions to focus on Dev, not Prod:

    ENV ZANATA_VERSION
    RUN wget
    ADD conf/standalone.xml

And modify rundev.sh to map the local conf/standalone.xml into the container.


@definite will try to modify the Dockerfile to work for both Dev and Prod after this PR goes in.  But I just realised that you can't use environment variables inside a docker build.

If it proves too difficult , maybe we should just keep it simple, and go with two Dockerfiles. The number of shared instructions won't be huge anyway. The most important thing is to share the standalone.xml, not the Dockerfile.

---

*[docker/README.md, line 4 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCgzpBvqExmx9HoBSW:-KKLkpOufFc6-vtU7iwV:-467936277) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L4)):*
> Mmm.... I'll disagree here. There are ways to install the latest docker version for every system. I don't mind if we test and perhaps note if these images work with docker 1.8, but why tie them down if there are ways to run them with the latest?
> 
> The other point is the MariaDB image (for instance) is tested against 1.11, and they say support for lower versions is 'best-effort'. I say we follow this approach.
</details>
It's often possible to install libraries from upstream, but it can sometimes be very bad for system stability (eg if the upstream yum repo has dependencies which conflict with the OS). Installing Fedora's (or RHEL's) copy of Docker is much safer for your system than grabbing the latest Docker from upstream, especially for privileged code like Docker.

Is there something from recent Docker that you're really keen to make use of?

Fedora 22 won't be around much longer (and 1.9 is in updates-testing), but Fedora 23/24 has 1.10, and RHEL 7 now has Docker 1.9. How about we make 1.9 the baseline?

---

*[docker/README.md, line 36 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCi1alOzQfeWV1E7VM:-KKLmpPmVmIHt80blZBU:-732850540) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L36)):*
> My point here was to show how other clients can connect to the database, with things like DBeaver in mind. The mysql client is just an example which is easy to illustrate without a screenshot. I will add an alternative example showing how to do this.
</details>
Okay. I suggest saying that in the text.

If someone wants to use the mysql command, I still think it's better to use docker exec to be sure of compatibility (and no need to install mysql on the host). So I think we should point them at docker exec for cases where that's all they need.

---

*[docker/rundev.sh, line 25 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKLzHViHbydRVSgQc5n:-KKLzHViHbydRVSgQc5o:-231149122) ([raw file](https://github.com/zanata/zanata-server/blob/c855941428274f451906b83e19fe0bd505f9ab04/docker/rundev.sh#L25)):*
> ```Shell
> docker run --rm --name zanata --link zanatadb:db -p 8080:8080 -it \
>     -v $ZANATA_WAR:/opt/jboss/wildfly/standalone/deployments/ROOT.war \
>     -v $ZANATA_DIR:/opt/jboss/zanata \
> ```

Don't forget to map the local (ie latest) conf/standalone.xml into the container.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1204)*

seanf:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[docker/Dockerfile, line 6 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKM1fIA7MXDFmt0yUT5:-KKM1fIBQYw3JKaKzCqy:875449703) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/Dockerfile#L6)):*
> ```Dockerfile
> RUN yum -y install mysql-connector-java wget && yum clean all
> 
> RUN ln -sf /usr/share/java/mysql-connector-java.jar /opt/jboss/wildfly/standalone/deployments/mysql-connector-java.jar
> ```

At some point, we should look at fetching a specific version of the mysql driver from Maven Central, instead of using whatever version came with whatever version of Fedora is being used by the wildfly image.

---

*[docker/rundev.sh, line 26 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKM0qyChOwNorh_dU4U:-KKM0qyChOwNorh_dU4V:2076530354) ([raw file](https://github.com/zanata/zanata-server/blob/c855941428274f451906b83e19fe0bd505f9ab04/docker/rundev.sh#L26)):*
> ```Shell
>     -v $ZANATA_WAR:/opt/jboss/wildfly/standalone/deployments/ROOT.war \
>     -v $ZANATA_DIR:/opt/jboss/zanata \
>     zanata/server
> ```

For Dev, instead of running `zanata/server` (probably fetched from docker hub), I think we should run `docker build` just before `docker run`, with a different image name like `zanata/server-dev`. The good news is, it should be highly cacheable.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1204)*

carlosmunoz:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[docker/README.md, line 4 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCgzpBvqExmx9HoBSW:-KKMZWmgGmJDO3xRHk_H:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L4)):*
> It's often possible to install libraries from upstream, but it can sometimes be very bad for system stability (eg if the upstream yum repo has dependencies which conflict with the OS). Installing Fedora's (or RHEL's) copy of Docker is much safer for your system than grabbing the latest Docker from upstream, especially for privileged code like Docker.
> 
> Is there something from recent Docker that you're really keen to make use of?
> 
> Fedora 22 won't be around much longer (and 1.9 is in updates-testing), but Fedora 23/24 has 1.10, and RHEL 7 now has Docker 1.9. How about we make 1.9 the baseline?
</details>
Done.

---

*[docker/README.md, line 36 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKCi1alOzQfeWV1E7VM:-KKMaSs4-8BjIXUebaBw:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/8efb328dbcb277c185546cb6d39d2a328d11bdba/docker/README.md#L36)):*
> Okay. I suggest saying that in the text.
> 
> If someone wants to use the mysql command, I still think it's better to use docker exec to be sure of compatibility (and no need to install mysql on the host). So I think we should point them at docker exec for cases where that's all they need.
</details>
Done.

---

*[docker/rundev.sh, line 25 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKLzHViHbydRVSgQc5n:-KKMaVpv3tSl_60SUmg6:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/c855941428274f451906b83e19fe0bd505f9ab04/docker/rundev.sh#L25)):*
> Don't forget to map the local (ie latest) conf/standalone.xml into the container.
</details>
Done.

---

*[docker/rundev.sh, line 26 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1204#-KKM0qyChOwNorh_dU4U:-KKMaaO2qgykUNDhblhv:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/c855941428274f451906b83e19fe0bd505f9ab04/docker/rundev.sh#L26)):*
> For Dev, instead of running `zanata/server` (probably fetched from docker hub), I think we should run `docker build` just before `docker run`, with a different image name like `zanata/server-dev`. The good news is, it should be highly cacheable.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1204)*

seanf:
Reviewed 3 of 3 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1204)*

seanf:
I suggest that we back-port this to `release`, since it could be very handy for maintenance, and the required configuration (standalone.xml) should be the same.

definite:
@seanf Perhaps we can use template like `Dockerfile.in`, like

```
ADD @MAVEN_URL@/mysql-connector-java-@MYSQL_CONNECTOR_VERSION@.jar mysql-connector-java.jar
```

 then ask the image building script to replace the place holder inside `Dockerfile.in`.

But that can be in other PR to do that.

seanf:
Remember, for automated builds on Docker Hub
<https://docs.docker.com/docker-hub/builds/>, the *prod* Dockerfile must
exist in the repo as is, not as Dockerfile.in. If we build it from a
template, we would have to be pushing the generated Dockerfiles to github
whenever the template changed, which is not good.

Now it may be possible to generate the dev Dockerfile from the prod
Dockerfile (using special markup comments or something) when rundev.sh is
executed, but frankly that would be over-engineering, because we're only
talking about a few lines of Dockerfile instructions. It will be even fewer
lines, once we eliminate wildfly-module-hibernate-main-4.2.20.Final.zip.

The important thing is to share the wildfly config. I think working on
reducing the number of standalone.xml files would be much more beneficial
than reducing two Dockerfiles into one. Finding a way to share with
zanata-overlay would be a good start.

On 16 June 2016 at 18:33, Ding-Yi Chen <notifications@github.com> wrote:

> @seanf <https://github.com/seanf> Perhaps we can use template like
> Dockerfile.in, like
>
> ADD @MAVEN_URL@/mysql-connector-java-@MYSQL_CONNECTOR_VERSION@.jar mysql-connector-java.jar
>
> then ask the image building script to replace the place holder inside
> Dockerfile.in.
>
> But that can be in other PR to do that.
>
> —
> Reply to this email directly, view it on GitHub
> <https://github.com/zanata/zanata-server/pull/1204#issuecomment-226422169>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AADTi1KWsN5WHrXMKhG7tbb6JijS0AKiks5qMQpfgaJpZM4I0491>
> .
>



-- 
Sean Flanigan

Principal Software Engineer
Red Hat

carlosmunoz:
Closing. Will rebase this on top of the release branch.

PR #1205, opened 2016-06-14
fix(po file): Fix po-revision-date header

aeng:
https://zanata.atlassian.net/browse/ZNTA-1145

Reviewed in https://github.com/zanata/zanata-server/pull/1191

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1205)

### DIFF
--- a/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java
+++ b/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java
@@ -124,6 +124,10 @@ public class ResourceUtils {
 
     private static Properties pluralForms;
 
+    public static final String COPIED_BY_ZANATA_NAME = "Copied by Zanata";
+    public static final String COPIED_BY_ZANATA_NAME_EMAIL =
+            "copied-by-zanata@zanata.org";
+
     @Inject
     private EntityManager entityManager;
 
@@ -704,8 +708,7 @@ public class ResourceUtils {
             final List<HTextFlowTarget> hTargets, final HLocale locale) {
         final Map<String, HeaderEntry> containedHeaders =
                 new LinkedHashMap<String, HeaderEntry>(headerEntries.size());
-        HTextFlowTarget lastTranslatedTarget =
-                this.getLastTranslatedTarget(hTargets);
+        HTextFlowTarget lastChangedTarget = getLastChangedTarget(hTargets);
 
         // Collect the existing header entries
         for (HeaderEntry entry : headerEntries) {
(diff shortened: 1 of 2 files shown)


seanf:
At this point, it's very difficult to work out exactly what was reviewed in #1191, because that branch got totally messed up.

@huangp I think you'd best give it the once over.

huangp:
Reviewed 2 of 2 files at r1.

---

*[zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java, line 894 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1205#-KKGYLTiNAtQ82Z63DQD:-KKGYLTiNAtQ82Z63DQE:1423384464) ([raw file](https://github.com/zanata/zanata-server/blob/b9eb1ffdf1c99d23f5f7f6e4a010867a7ac1f37a/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java#L894)):*
> ```Java
> 
>             if (lastTranslatedBy != null) {
>                 if (lastModifiedDate == null
> ```

I don't think lastChanged/lastModifiedDate can be null. Only the person can not the date. So no need to check here

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1205)*

seanf:

---

*[zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java, line 894 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1205#-KKGYLTiNAtQ82Z63DQD:-KKGZoSt0u801ql4mAmZ:577630560) ([raw file](https://github.com/zanata/zanata-server/blob/b9eb1ffdf1c99d23f5f7f6e4a010867a7ac1f37a/zanata-war/src/main/java/org/zanata/rest/service/ResourceUtils.java#L894)):*
<details><summary><i>Previously, huangp (Patrick Huang) wrote…</i></summary>
> I don't think lastChanged/lastModifiedDate can be null. Only the person can not the date. So no need to check here
</details>
It's true that lastModifiedDate can't be null, because `HTextFlowTarget.getLastChanged()` has `@Column(nullable = false)`. I had to check, because it's not obvious from this code.

However, the Java compiler doesn't know that, so I think it's fine to do a null check, to help out the null analysis. Perhaps `HTextFlowTarget.getLastChanged()` ought to be given a `NotNull` annotation (in `master`, not here).

For now, I recommend we keep this code as is, if we can. 

We already have too many merge conflicts because of this bug fix, because the messed up pull request https://github.com/zanata/zanata-server/pull/1191 had to be reverted on `release`, and then the **revert** had to be reverted on `master`, which had the effect of bringing back the bug fix on `master`. (After the first revert, I should have merged `release` to `master` manually with the 'ours' merge strategy, thus avoiding the second revert, but it's too late now.)

As long as this change is identical to the change which was originally merged, it should be easy to resolve the resulting "`release` to `master`" merge conflict.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1205)*

djansen-redhat:
Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1205)*

djansen-redhat:
:white_check_mark: tested. :shipit:

PR #1208, opened 2016-06-16
fix(ZNTA-1174): add OpenID credentials, transactions not committed

seanf:
https://zanata.atlassian.net/browse/ZNTA-1174

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1208)

### DIFF
--- a/zanata-war/src/main/java/org/zanata/action/AbstractProfileAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/AbstractProfileAction.java
@@ -5,7 +5,6 @@ import javax.validation.constraints.Size;
 import org.hibernate.validator.constraints.Email;
 import org.hibernate.validator.constraints.NotEmpty;
 import javax.inject.Inject;
-import org.zanata.seam.security.ZanataJpaIdentityStore;
 import org.zanata.dao.AccountDAO;
 import org.zanata.dao.PersonDAO;
 import org.zanata.model.HAccount;
--- a/zanata-war/src/main/java/org/zanata/action/AccountMergeAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/AccountMergeAction.java
@@ -30,8 +30,12 @@ import lombok.Setter;
 import javax.annotation.Nullable;
 import javax.enterprise.context.SessionScoped;
 import javax.annotation.PostConstruct;
+import javax.enterprise.inject.Model;
+import javax.faces.bean.ViewScoped;
 import javax.inject.Inject;
 import javax.inject.Named;
+
+import org.apache.deltaspike.jpa.api.transaction.Transactional;
 import org.zanata.dao.AccountDAO;
 import org.zanata.model.HAccount;
 import org.zanata.security.AuthenticationManager;
@@ -50,7 +54,9 @@ import org.zanata.util.Synchronized;
  *         href="mailto:camunoz@redhat.com">camunoz@redhat.com</a>
  */
 @Named("accountMergeAction")
-@javax.faces.bean.ViewScoped
+@ViewScoped
+@Model
+@Transactional
 public class AccountMergeAction implements Serializable {
 
     @SessionScoped
--- a/zanata-war/src/main/java/org/zanata/action/ActivateAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/ActivateAction.java
@@ -114,7 +114,7 @@ public class ActivateAction implements Serializable {
         facesMessages.addGlobal(FacesMessage.SEVERITY_INFO,
                 "Your account was successfully activated. You can now sign in.");
 
-        urlUtil.redirectTo(urlUtil.signInPage());
+        urlUtil.redirectToInternal(urlUtil.signInPage());
         conversation.close();
     }
 }
--- a/zanata-war/src/main/java/org/zanata/action/ActivateAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/ActivateAction.java
@@ -23,6 +23,7 @@ package org.zanata.action;
 import java.io.Serializable;
 import java.util.Date;
 
+import javax.enterprise.inject.Model;
 import javax.faces.application.FacesMessage;
 
 import org.apache.commons.lang.time.DateUtils;
@@ -46,6 +47,8 @@ import lombok.Setter;
 
 @Named("activate")
 @GroupedConversationScoped
+@Model
+@Transactional
 public class ActivateAction implements Serializable {
 
     private static final long serialVersionUID = -8079131168179421345L;
--- a/zanata-war/src/main/java/org/zanata/action/ActivityAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/ActivityAction.java
@@ -23,9 +23,12 @@ package org.zanata.action;
 import java.io.Serializable;
 import java.util.Collections;
 import java.util.List;
+import javax.enterprise.inject.Model;
+import javax.faces.bean.ViewScoped;
 import javax.inject.Inject;
 import javax.inject.Named;
 
+import org.apache.deltaspike.jpa.api.transaction.Transactional;
 import org.zanata.model.Activity;
 import org.zanata.model.HAccount;
 import org.zanata.security.annotations.Authenticated;
@@ -36,9 +39,10 @@ import org.zanata.service.ActivityService;
  * @author Alex Eng <a href="mailto:aeng@redhat.com">aeng@redhat.com</a>
  */
 @Named("activityAction")
-@javax.faces.bean.ViewScoped
-
+@ViewScoped
 @CheckLoggedIn
+@Model
+@Transactional
 public class ActivityAction implements Serializable {
     private static final long serialVersionUID = 1L;
 
--- a/zanata-war/src/main/java/org/zanata/action/AdminStatsAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/AdminStatsAction.java
@@ -24,8 +24,13 @@ import java.io.Serializable;
 
 import lombok.Getter;
 import lombok.Setter;
+
+import javax.enterprise.inject.Model;
+import javax.faces.bean.ViewScoped;
 import javax.inject.Inject;
 import javax.inject.Named;
+
+import org.apache.deltaspike.jpa.api.transaction.Transactional;
 import org.zanata.dao.DocumentDAO;
 import org.zanata.dao.PersonDAO;
 import org.zanata.dao.ProjectDAO;
@@ -39,7 +44,9 @@ import org.zanata.dao.TextFlowTargetDAO;
  *
  **/
 @Named("adminStatsAction")
-@javax.faces.bean.ViewScoped
+@ViewScoped
+@Model
+@Transactional
 public class AdminStatsAction implements Serializable {
     /**
     *
--- a/zanata-war/src/main/java/org/zanata/action/ApplicationConfigurationAction.java
+++ b/zanata-war/src/main/java/org/zanata/action/ApplicationConfigurationAction.java
@@ -21,8 +21,13 @@
 package org.zanata.action;
 
 import org.apache.commons.lang.StringUtils;
+
+import javax.enterprise.context.Dependent;
+import javax.enterprise.inject.Model;
 import javax.inject.Inject;
 import javax.inject.Named;
+
+import org.apache.deltaspike.jpa.api.transaction.Transactional;
 import org.zanata.ApplicationConfiguration;
 
 /**
@@ -34,7 +39,9 @@ import org.zanata.ApplicationConfiguration;
  *         href="mailto:camunoz@redhat.com">camunoz@redhat.com</a>
  */
 @Named("applicationConfigurationAction")
-@javax.enterprise.context.Dependent
+@Dependent
+@Model
+@Transactional
 public class ApplicationConfigurationAction {
     @Inject
     private ApplicationConfiguration applicationConfiguration;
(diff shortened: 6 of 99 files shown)


carlosmunoz:
:lgtm:

Never have done the static producer method, but makes sense for it to work. -(Awaiting review) +Reviewed 

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-:-KKMUmmds4VcjG7_5WID:-171888955)*

seanf:
The fix for this has exposed wider problems (in various services) with transactions not being committed, ie messages like "your changes have been saved" but in fact the changes get rolled back.

seanf:
@carlosmunoz Could you please take another look to review the transaction changes?

Even in cases where we don't modify the database, we should ideally use transactions to ensure a consistent view during the request. 

I have added `@Transactional` to various REST services, and added a JSF `PhaseListener` for Seam-style transactions, but I also added `@Transactional` to various services (and `ZanataOpenId`) where I can see DB access. (Except a few cases where the transactions might be too long, eg TMX download.)

However, I still can't work out how some GWT services are able to save their data without any visible transaction management. Many of them call other beans which have `@Transactional` around their DB changes, but then there are cases like `SaveOptionHandler` and `AddReviewCommentHandler` found by @huangp which don't, and yet their changes do seem to be persistent in my tests.

I don't think we should simply put `@Transactional` on `SeamDispatch`/`GwtDispatchService`, because actions like `TransMemoryMerge` may be too big for one transaction.

Irksome though it is, I'm leaving the GWT services in the too-hard-to-verify basket for now, since at least they seem to be working.

seanf:
@huangp I think you should take a look too, please. We want to get transactions right.

Note that this PR might be easier to review in GitHub, not Reviewable. Reviewable doesn't perform well with lots of small changes (eg adding the same annotation to scores of files).

seanf:

carlosmunoz:
Reviewed 5 of 6 files at r2, 96 of 96 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[zanata-war/src/main/java/org/zanata/action/CopyTransManager.java, line 53 \[r3\]](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-KLEQV26s_CKiRGwGJjc:-KLEQV26s_CKiRGwGJjd:2096686559) ([raw file](https://github.com/zanata/zanata-server/blob/c5c99347877076a2cd72a04086647beee16e6d3e/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java#L53)):*
> ```Java
>  */
> 
> @Named("copyTransManager")
> ```

Just want to make sure we are not using this bean by name somewhere.

---

*[zanata-war/src/main/java/org/zanata/ui/ModelAction.java, line 48 \[r3\]](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-KLETk1E8h9Xie2WZu4a:-KLETk1E8h9Xie2WZu4b:876124583) ([raw file](https://github.com/zanata/zanata-server/blob/b2669b3b65b2e1bf32080326aeaaf7d52bc7d32f/zanata-war/src/main/java/org/zanata/ui/ModelAction.java#L48)):*
> ```Java
> @Target( { TYPE, METHOD, FIELD })
> @Retention(RUNTIME)
> public @interface ModelAction {
> ```

Now that I read your description and look at the code, perhaps we should've gone with a name like `TransactionalModel`

---

*[zanata-war/src/main/java/org/zanata/util/UrlUtil.java, line 257 \[r3\]](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-KLEVW95en5fUfAi_Cwt:-KLEVW96BvR-vSVU_ySq:591941703) ([raw file](https://github.com/zanata/zanata-server/blob/b2669b3b65b2e1bf32080326aeaaf7d52bc7d32f/zanata-war/src/main/java/org/zanata/util/UrlUtil.java#L257)):*
> ```Java
>      * Redirect to a Zanata url, adding dswid parameter if missing. Do not use for external URLs!
>      */
>     public void redirectTo(String url) {
> ```

Should we perhaps add something to the method name to specify this is an internal redirection mechanism? .. like `redirectToInternal`

---

*[zanata-war/src/main/webapp-jboss/WEB-INF/classes/META-INF/persistence.xml, line 68 \[r3\]](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-KLEVjlnH-DPYZJBuOa2:-KLEVjlnH-DPYZJBuOa3:1070219665) ([raw file](https://github.com/zanata/zanata-server/blob/b2669b3b65b2e1bf32080326aeaaf7d52bc7d32f/zanata-war/src/main/webapp-jboss/WEB-INF/classes/META-INF/persistence.xml#L68)):*
> ```XML
>         value="java:/zanataEntityManagerFactory" />
> 
>       <property name="hibernate.connection.autocommit" value="true" />
> ```

Does this work? if not, it might be misleading to leave it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1208)*

seanf:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[zanata-war/src/main/java/org/zanata/action/CopyTransManager.java, line 53 \[r3\]](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-KLEQV26s_CKiRGwGJjc:-KLEjotw8kT03-gzE9La:-1180200492) ([raw file](https://github.com/zanata/zanata-server/blob/c5c99347877076a2cd72a04086647beee16e6d3e/zanata-war/src/main/java/org/zanata/action/CopyTransManager.java#L53)):*
> Just want to make sure we are not using this bean by name somewhere.
</details>
No, it's injected into various Action classes, not used in xhtml.

---

*[zanata-war/src/main/java/org/zanata/ui/ModelAction.java, line 48 \[r3\]](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-KLETk1E8h9Xie2WZu4a:-KLEoYTYAmzzGCFb5C5L:485653824) ([raw file](https://github.com/zanata/zanata-server/blob/b2669b3b65b2e1bf32080326aeaaf7d52bc7d32f/zanata-war/src/main/java/org/zanata/ui/ModelAction.java#L48)):*
> Now that I read your description and look at the code, perhaps we should've gone with a name like `TransactionalModel`
</details>
I chose Action to be part of the stereotype name because most of the candidate beans are already named somethingAction. Although  whoever created `javax.enterprise.inject.Model` (Gavin King) would probably consider such classes models, not actions.

If we start putting all the attributes of the stereotype into its name, there won't be much left. `@TransactionalModel` will be basically the same as `@Transactional @Model`. 

It's also affected by the fact that we now have the `TransactionPhaseListener` to provide Seam-style transactions for JSF requests. I think `@Transactional` should interact safely with `TransactionPhaseListener` (by joining the phase listener's transactions), so perhaps it's still worth having the annotation, if only for documentation purposes. 

Perhaps I should get rid of the stereotype and mark all the Actions with `@Transactional @Model` instead.  Or I could get rid of all the  annotation changes, and let the phase listener do its job. What do you think?

---

*[zanata-war/src/main/java/org/zanata/util/UrlUtil.java, line 257 \[r3\]](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-KLEVW95en5fUfAi_Cwt:-KLEkr37nj2ut3exNsHO:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/b2669b3b65b2e1bf32080326aeaaf7d52bc7d32f/zanata-war/src/main/java/org/zanata/util/UrlUtil.java#L257)):*
> Should we perhaps add something to the method name to specify this is an internal redirection mechanism? .. like `redirectToInternal`
</details>
Done.

---

*[zanata-war/src/main/webapp-jboss/WEB-INF/classes/META-INF/persistence.xml, line 68 \[r3\]](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-KLEVjlnH-DPYZJBuOa2:-KLEnBZFBmufFGt0fEV8:1517192070) ([raw file](https://github.com/zanata/zanata-server/blob/b2669b3b65b2e1bf32080326aeaaf7d52bc7d32f/zanata-war/src/main/webapp-jboss/WEB-INF/classes/META-INF/persistence.xml#L68)):*
> Does this work? if not, it might be misleading to leave it.
</details>
Reverted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1208)*

carlosmunoz:
Reviewed 10 of 12 files at r4.

---

*[zanata-war/src/main/java/org/zanata/ui/ModelAction.java, line 48 \[r3\]](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-KLETk1E8h9Xie2WZu4a:-KLEv95Oi90YI7WpLtdf:-1127263745) ([raw file](https://github.com/zanata/zanata-server/blob/b2669b3b65b2e1bf32080326aeaaf7d52bc7d32f/zanata-war/src/main/java/org/zanata/ui/ModelAction.java#L48)):*
> I chose Action to be part of the stereotype name because most of the candidate beans are already named somethingAction. Although  whoever created `javax.enterprise.inject.Model` (Gavin King) would probably consider such classes models, not actions.
> 
> If we start putting all the attributes of the stereotype into its name, there won't be much left. `@TransactionalModel` will be basically the same as `@Transactional @Model`. 
> 
> It's also affected by the fact that we now have the `TransactionPhaseListener` to provide Seam-style transactions for JSF requests. I think `@Transactional` should interact safely with `TransactionPhaseListener` (by joining the phase listener's transactions), so perhaps it's still worth having the annotation, if only for documentation purposes. 
> 
> Perhaps I should get rid of the stereotype and mark all the Actions with `@Transactional @Model` instead.  Or I could get rid of all the  annotation changes, and let the phase listener do its job. What do you think?
</details>
I like being specific, but maybe I'd go with `@Transactional @Model` instead of having an extra annotation just for that. 

---

*[zanata-war/src/main/webapp-jboss/WEB-INF/classes/META-INF/persistence.xml, line 68 \[r3\]](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-KLEVjlnH-DPYZJBuOa2:-KLEuaPJO5ffgAkYHmFX:-1874550970) ([raw file](https://github.com/zanata/zanata-server/blob/b2669b3b65b2e1bf32080326aeaaf7d52bc7d32f/zanata-war/src/main/webapp-jboss/WEB-INF/classes/META-INF/persistence.xml#L68)):*
> Reverted.
</details>
All good. I think there are a couple more instances of this line in the other persistence.xml's 

After that, this looks good pending green tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1208)*

seanf:
Review status: 50 of 99 files reviewed at latest revision, 2 unresolved discussions.

---

*[zanata-war/src/main/java/org/zanata/ui/ModelAction.java, line 48 \[r3\]](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-KLETk1E8h9Xie2WZu4a:-KLFPpz4X6Aui9RYSgwH:-499146009) ([raw file](https://github.com/zanata/zanata-server/blob/b2669b3b65b2e1bf32080326aeaaf7d52bc7d32f/zanata-war/src/main/java/org/zanata/ui/ModelAction.java#L48)):*
> I like being specific, but maybe I'd go with `@Transactional @Model` instead of having an extra annotation just for that. 
</details>
Done.

---

*[zanata-war/src/main/webapp-jboss/WEB-INF/classes/META-INF/persistence.xml, line 68 \[r3\]](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-KLEVjlnH-DPYZJBuOa2:-KLElAMzlTnq1TkhZsjb:681067958) ([raw file](https://github.com/zanata/zanata-server/blob/b2669b3b65b2e1bf32080326aeaaf7d52bc7d32f/zanata-war/src/main/webapp-jboss/WEB-INF/classes/META-INF/persistence.xml#L68)):*
> All good. I think there are a couple more instances of this line in the other persistence.xml's

No problem: I reverted that whole commit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1208)*

carlosmunoz:
:lgtm:
+Reviewed -(Awaiting review) 

---

Reviewed 50 of 50 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1208#-:-KLFRHZq6gpmJ9yrw0PM:-1199085999)*

anipeter:
Few observations listed below. All are known, but writing them down for records.
* Successfully logged in using Fedora, Yahoo & livejournal OpenD.
* The new user page & Dashboard (after login) is displayed with "Login Failed" notification.
* The behaviour in the new user page is not consistent - Fedora login has the user information prefilled whereas other openid methods showed empty "new user" page
* The "email is already taken" notification appears twice as shown in screenshot
twice email taken notification-fedora openid tesing

anipeter:
Logged in successfully, so good to go :+1:

seanf:
Thanks @anipeter. So, those are all previously known issues? Is that known in Zanata 3.8, or in 3.9?

anipeter:
@seanf yes these are known issues from 3.9 caused as regression of CDI. 
Bugs are already filed for "Login failed" notifications. Need to check if we have bugs filed for notifications like "The email is already taken" and "The username is already taken" appearing twice.
Hope to get the inconsistencies with "new user" page to be fixed with OAuth.

seanf:
For reference, the "Login failed" regression is https://zanata.atlassian.net/browse/ZNTA-1112

anipeter:
Bug filed for "The email already taken" & "The username already taken" notifications appearing twice: https://zanata.atlassian.net/browse/ZNTA-1208

PR #1218, opened 2016-06-23
refactor(profile): migrate profile page to redux

aeng:
https://zanata.atlassian.net/browse/ZNTA-1163

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1218)


huangp:
Reviewed 22 of 22 files at r1.

---

*[frontend/src/main/web/src/actions/profile.js, line 41 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1218#-KKueBJBNgv8BSV_tpZR:-KKueBJCpu0wtL5r_meU:-1994606518) ([raw file](https://github.com/zanata/zanata-server/blob/be90a076ef49b24c74fead82f3dd8690d02edf56/frontend/src/main/web/src/actions/profile.js#L41)):*
> ```JavaScript
>         const contentType = res.headers.get('Content-Type')
>         if (contentType && includes(contentType, 'json')) {
>           return res.json().then((json) => {
> ```

I thought this will be done automatically by redux-api-middleware? At least I don't need to do this in Zanata Sync (but I always consume and produce json from the server)

---

*[frontend/src/main/web/src/containers/UserProfile/ContentStateFilter.jsx, line 76 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1218#-KKufh9XqrAzupda93-Z:-KKufh9YQME2mH95nTYi:1080304315) ([raw file](https://github.com/zanata/zanata-server/blob/be90a076ef49b24c74fead82f3dd8690d02edf56/frontend/src/main/web/src/containers/UserProfile/ContentStateFilter.jsx#L76)):*
> ```JSX
>   handleFilterChanged,
>   ...props
> }) => {
> ```

Just curious. What will this function syntax do? Will it apply the pureRenderMixin automatically?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1218)*

aeng:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[frontend/src/main/web/src/actions/profile.js, line 41 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1218#-KKueBJBNgv8BSV_tpZR:-KKv4qbXZygMbqTj-Lbf:1114818334) ([raw file](https://github.com/zanata/zanata-server/blob/be90a076ef49b24c74fead82f3dd8690d02edf56/frontend/src/main/web/src/actions/profile.js#L41)):*
<details><summary><i>Previously, huangp (Patrick Huang) wrote…</i></summary>
> I thought this will be done automatically by redux-api-middleware? At least I don't need to do this in Zanata Sync (but I always consume and produce json from the server)
</details>
Not sure myself. Been doing that in other api call.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1218)*

aeng:
@huangp 

> Just curious. What will this function syntax do? Will it apply the pureRenderMixin automatically?

It used for stateless component. And for redux, which is single state, there's no need of pureRenderMixin

huangp:
I think they are two different concepts. pureRenderMixin will skip re-render if props don't change (it basically implements a shouldComponentUpdate() method). stateless component is a candidate for it. So now this will re-render each time the parent component re-renders. I don't think it will kill the performance but it's nice to have. 
see https://github.com/facebook/react/issues/5677

---

Review status: 20 of 22 files reviewed at latest revision, 2 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/zanata/zanata-server/1218#-KKvCjwWptjpSWu1vGHP:-KKvCjwWptjpSWu1vGHQ:7167892) (no related file):*
:lgtm: -(Awaiting review) +Reviewed +(On QA) 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1218#-:-KKvBII7KREYrjl4-t-7:-404764260)*

aeng:
@huangp You're right. But just found out

> Unfortunately ES6 launched without any mixin support. Therefore, there is no support for mixins when you use React with ES6 classes. Instead, we're working on making it easier to support such use cases without resorting to mixins.

But in this case, there shouldn't be any performance issue as the there's only minimal state changes in profile page.

djansen-redhat:
:white_check_mark: tested

PR #1222, opened 2016-06-27
Remove Hibernate module for WildFly

seanf:
https://zanata.atlassian.net/browse/ZNTA-1203

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1222)

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -3,9 +3,15 @@
 ##### Bug Fixes
 * [ZNTA-846](https://zanata.atlassian.net/browse/ZNTA-846) - Group "request add project version" needs field limit
 * [ZNTA-872](https://zanata.atlassian.net/browse/ZNTA-872) - Redirect to home from cancel button in create user page
+
 ##### Infrastructure Changes
 
 * Zanata's cache configuration (Infinispan) is now controlled by `zanata-infinispan.xml`, not `standalone.xml`. Please see the [Infinispan](user-guide/system-admin/configuration/infinispan) section for more information.
+
+The WildFly module for Hibernate 4.2 has been removed. Zanata will now use the version of Hibernate provided by
+WildFly. The Zanata installer no longer installs any modules for WildFly. If a previous version of the Zanata
+installer has put a Hibernate 4.2 module into your WildFly installation (in modules/org/hibernate/main) you should
+remove it.
 
 ## 3.9
 
--- a/functional-test/pom.xml
+++ b/functional-test/pom.xml
@@ -423,19 +423,6 @@
         <jndi-remote-client.artifactId>wildfly-controller-client</jndi-remote-client.artifactId>
         <jndi-remote-client.version>${wildfly.client.version}</jndi-remote-client.version>
       </properties>
-      <build>
-        <plugins>
-          <plugin>
-            <artifactId>maven-antrun-plugin</artifactId>
-            <executions>
-              <execution>
-                <id>install-wildfly-modules</id>
-                <phase>package</phase>
-              </execution>
-            </executions>
-          </plugin>
-        </plugins>
-      </build>
       <dependencyManagement>
         <dependencies>
           <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -1925,34 +1925,6 @@
         <hibernate.module.url>https://bintray.com/artifact/download/sflanigan/generic/${hibernate.module.filename}</hibernate.module.url>
       </properties>
 
-      <build>
-        <plugins>
-          <!--
-          Download wildfly modules from sf.net and extract under appserver.home
-          -->
-          <plugin>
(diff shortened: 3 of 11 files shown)


carlosmunoz:
Reviewed 13 of 13 files at r1.

---

*[zanata-overlay/build.gradle, line 68 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1222#-KLKQvbqOjLx0j1jYti0:-KLKQvbrdvkUbkGgiWHc:2056541716) ([raw file](https://github.com/zanata/zanata-server/blob/05a0b4f1dd5e9d6abacca63bad8490112223dcb9/zanata-overlay/build.gradle#L68)):*
> ```Gradle
>         }
>         // Download remote dependencies
>         downloadRemoteDeps(distro, distroDir)
> ```

We might want to keep this, in case there are ever other remote dependencies which need to be added. It would just not do much at the moment.

---

*[zanata-overlay/build.gradle, line 109 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1222#-KLK_YmmYf7eZpO0aTMA:-KLK_YmmYf7eZpO0aTMB:721328965) ([raw file](https://github.com/zanata/zanata-server/blob/05a0b4f1dd5e9d6abacca63bad8490112223dcb9/zanata-overlay/build.gradle#L109)):*
> ```Gradle
> }
> 
> def downloadRemoteDeps(String distro, String distroDir) {
> ```

Same comment as above

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1222)*

seanf:

---

*[zanata-overlay/build.gradle, line 68 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1222#-KLKQvbqOjLx0j1jYti0:-KLOVHXvFN-8P3vtkjpm:-883243575) ([raw file](https://github.com/zanata/zanata-server/blob/05a0b4f1dd5e9d6abacca63bad8490112223dcb9/zanata-overlay/build.gradle#L68)):*
> We might want to keep this, in case there are ever other remote dependencies which need to be added. It would just not do much at the moment.
</details>
Well... YAGNI. And KISS.

I'll see if I can bring back enough of it to make it (somewhat) clear how to use this feature, without having it actually download anything.

But do we think we'll use modules again anytime soon, or will we just be expending effort maintaining a function which won't be used for a couple of years, if ever?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1222)*

carlosmunoz:
:lgtm:
+Reviewed

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[zanata-overlay/build.gradle, line 68 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1222#-KLKQvbqOjLx0j1jYti0:-KLOcI4T4SvTN-HTBvUy:-1431683331) ([raw file](https://github.com/zanata/zanata-server/blob/05a0b4f1dd5e9d6abacca63bad8490112223dcb9/zanata-overlay/build.gradle#L68)):*
> Well... YAGNI. And KISS.
> 
> I'll see if I can bring back enough of it to make it (somewhat) clear how to use this feature, without having it actually download anything.
> 
> But do we think we'll use modules again anytime soon, or will we just be expending effort maintaining a function which won't be used for a couple of years, if ever?
</details>
Well, it's already there, so it's not like we're going to waste time developing it. And it's also in history. So you make the call. I might have a feeling this will evolve even further perhaps with docker or swarm.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1222#-:-KLOd0N4rRzxTfIE5bhn:-489062391)*

seanf:
@carlosmunoz I've made a few documentation/comment changes. I've also brought back the wildfly module download (in overlay) as an example in comments. 

But if we want to download wildfly modules for Maven tests in future, we'll just have to refer to the git history for pom.xml. I really don't want to maintain those bits unless we need them.

carlosmunoz:
:lgtm:
-(Awaiting review) +Reviewed 

---

Reviewed 4 of 7 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1222#-:-KLPBM1gDBWcQFp-wX4R:-1149335731)*

seanf:
Notes for QA:

Functional tests and production builds no longer have a JDBC wrapper active: a debugging wrapper which does (and did) its very best to be inert when mysql is in use.

Also, as described in the release notes, the hibernate module for wildfly is no longer needed, nor is it installed by zanata-overlay.

djansen-redhat:
:white_check_mark: This seems to be working as intended

seanf:
Note: I had to revert these changes for now, because they break Arquillian tests (and thus the build) on WildFly.

PR #1225, opened 2016-06-29
feat(webhook): Test event for new webhook

aeng:
https://zanata.atlassian.net/browse/ZNTA-1160

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1225)

### DIFF
--- a/functional-test/src/main/java/org/zanata/page/projects/projectsettings/ProjectWebHooksTab.java
+++ b/functional-test/src/main/java/org/zanata/page/projects/projectsettings/ProjectWebHooksTab.java
@@ -24,11 +24,9 @@ import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.openqa.selenium.By;
-import org.openqa.selenium.Keys;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebElement;
 import org.zanata.page.projects.ProjectBasePage;
-import org.zanata.util.WebElementUtil;
 
 import java.util.List;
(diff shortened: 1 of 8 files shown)


huangp:
Reviewed 4 of 4 files at r1.

---

*[zanata-war/src/main/java/org/zanata/action/ProjectHome.java, line 1120 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1225#-KLTSf_CpO_H9a0XMBZY:-KLTSf_CpO_H9a0XMBZZ:-776581177) ([raw file](https://github.com/zanata/zanata-server/blob/5c32141647532a6831e62c69ead89f1bfea460b7/zanata-war/src/main/java/org/zanata/action/ProjectHome.java#L1120)):*
> ```Java
>         TestEvent event =
>                 new TestEvent(identity.getAccountUsername(), getSlug());
>         WebHooksPublisher.publish(url, event, Optional.fromNullable(secret));
> ```

We should try to use java.util.Optional. Guava may drop it one day since java 8 has it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1225)*

djansen-redhat:
- [ ] If the user enters an invalid url and presses add, the url field is cleared and the buttons won't work until the page is refreshed
- [ ] Buttons don't re-enable after adding a wehbook
- [ ] Test webhook should validate the url

djansen-redhat:
- [ ] Pressing Test doesn't return, buttons can't be used
- [ ] Press add on an invalid url clears the url field

djansen-redhat:
:white_check_mark: tested

PR #1230, opened 2016-07-04
fix(docker): build with mac os and staff_selinux

definite:
Fix the docker build for [staff_selinux](http://linux.die.net/man/8/staff_selinux) and mac os

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1230)

### DIFF
--- a/docker/function.sh
+++ b/docker/function.sh
@@ -1,0 +1,37 @@
+#!/bin/bash -xeu
+
+function chcon_exec(){
+    case `uname` in
+	Darwin )
+	    # mac OS has neither SELinux nor chcon
+	    return
+	    ;;
+	Linux )
+	    # non root SELinux confined user need sudo
+	    if [ `id -u` -ne 0 ];then
+		if ! (id -Z | grep "unconfined_t");then
+		    sudo chcon $@
+		    return
+		fi
+	    fi
+	    chcon $@
+	    return
+	    ;;
+	* )
+	    return
+	    ;;
+    esac
+}
+
+function dir_setup_for_docker(){
+    local dir="$1"
+    mkdir -p "$dir"
+    chcon_exec -Rt svirt_sandbox_file_t "$dir"
+}
+
+function file_setup_for_docker(){
+    local f="$1"
+    local dir=$(dirname "$f")
+    dir_setup_for_docker "$dir"
+    chcon_exec -Rt svirt_sandbox_file_t "$f"
+}
--- a/docker/rundb.sh
+++ b/docker/rundb.sh
@@ -8,23 +8,18 @@ DB_ROOT_PASSWORD=rootpw
 
 VOLUME_DIR=$HOME/docker-volumes/zanata-mariadb
 
-mkdir -p $VOLUME_DIR
-case `uname` in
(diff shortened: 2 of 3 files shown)


definite:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[docker/rundev.sh, line 19 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1230#-KLnfdXqha3_Sz6QoK4d:-KLnmsA94vV52ijv_tKk:608688714) ([raw file](https://github.com/zanata/zanata-server/blob/a35e08179a8cd0828882bb42a7d944489f54c70c/docker/rundev.sh#L19)):*
> I think the intention was to run chcon only when the directory  was actually created, but I guess that wasn't true with `-p`.
> 
> Perhaps we ought to check for the existence of $ZANATA_DIR before mkdir and chcon.
</details>
`mkdir -p` should create the dir when missing, and ignore if the target is already created.
If it does encounter the permission deny, the script should be fail because `bash -e`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1230)*

definite:
Reviewed 2 of 2 files at r1, 1 of 3 files at r2.
Review status: 1 of 3 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1230)*

seanf:
Reviewed 2 of 3 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[docker/function.sh, line 33 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1230#-KLoC0oXcWYk7alO8T16:-KLoC0oXcWYk7alO8T17:65423878) ([raw file](https://github.com/zanata/zanata-server/blob/c217788f715efe3e8bdfd22e6452ea554cb6a9fd/docker/function.sh#L33)):*
> ```Shell
> 
> function file_setup_for_docker(){
>     local f="$1"
> ```

Can you have these two functions do nothing (ie no chcon) if the file/directory already exists?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1230)*

definite:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[docker/function.sh, line 33 \[r2\]](https://reviewable.io:443/reviews/zanata/zanata-server/1230#-KLoC0oXcWYk7alO8T16:-KLoFhyeM4V9hBry354U:1655881504) ([raw file](https://github.com/zanata/zanata-server/blob/c217788f715efe3e8bdfd22e6452ea554cb6a9fd/docker/function.sh#L33)):*
> Can you have these two functions do nothing (ie no chcon) if the file/directory already exists?
</details>
It is useful when the file/directory exists, but incorrect permission.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1230)*

definite:

definite:
This PR is no longer relevant, as we are now in flavour of data volume.

PR #1234, opened 2016-07-11
fix(arq): Allow Java 8 to auto allocate heap size

definite:
OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0

---

This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1234)

### DIFF
--- a/zanata-war/src/test/resources/arquillian.xml
+++ b/zanata-war/src/test/resources/arquillian.xml
@@ -31,7 +31,7 @@
   <container qualifier="wildfly81">
     <configuration>
       <!--<property name="jbossHome">${wildfly.home}</property>-->
-      <property name="javaVmArguments">-Dorg.jboss.as.logging.per-deployment=false</property>
+      <!-- <property name="javaVmArguments">-Dorg.jboss.as.logging.per-deployment=false</property> -->
       <!-- NB: don't put whitespace around the number -->
       <property name="managementPort">${jboss.management.http.port,env.JBOSS_MANAGEMENT_HTTP_PORT:10090}</property>
     </configuration>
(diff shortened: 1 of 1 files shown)


zanata-war/src/test/resources/arquillian.xml
```diff
@@ -18,10 +18,6 @@
       <!--
                   <property name="javaVmArguments">-Xms1536m -Xmx1536m -XX:MaxPermSize=256m -Dorg.jboss.as.logging.per-deployment=false -Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=8787</property>
       -->
-      <property
-        name="javaVmArguments">-Xms1536m -Xmx1536m -XX:MaxPermSize=256m -Dorg.jboss.as.logging.per-deployment=false
```
seanf: I think it would be best to keep the `-Dorg.jboss.as.logging.per-deployment=false` (here and in the comment below) to make sure logging doesn't get switched off: http://stackoverflow.com/a/14575830/14379

zanata-war/src/test/resources/arquillian.xml
```diff
@@ -34,10 +31,7 @@
   <container qualifier="wildfly81">
     <configuration>
       <!--<property name="jbossHome">${wildfly.home}</property>-->
-      <!--<property-->
-        <!--name="javaVmArguments">-Xms1536m -Xmx1536m -XX:MaxPermSize=256m -Dorg.jboss.as.logging.per-deployment=false-->
-      <!--</property>-->
-
+      <property name="javaVmArguments">-Dorg.jboss.as.logging.per-deployment=false</property>
```
seanf: It think that system property has actually been deprecated in WildFly - https://issues.jboss.org/browse/WFLY-2600, but I'm not quite sure what it's meant to be replaced by. See https://docs.jboss.org/author/display/WFLY10/Logging+Configuration#LoggingConfiguration-Attributes

For now, let's just keep it as a comment, as before.

seanf:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1234#-:-KMSWNHtF_LQI65F4O1N:1417088721)*

seanf:
@definite This was never merged to `release` or `master`.

PR #1183, opened 2016-05-24
ZNTA-1156 Embed Infinispan

seanf:
Ideally we would embed Infinispan 8.2 rather than 7.2, but we would have to resolve a number of conflicts in transitive dependency versions (Arquillian-related).

This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1183)

PR #1235, opened 2016-07-11
Mostly revert "ZNTA-1156 Embed Infinispan (#1183)"

seanf:
This reverts commit 35f111c6610db5cf0aac3c430b5a305e798e5bd2.

Infinispan is now accessed via jboss module again. This allows Zanata to
take advantage of configuration via standalone.xml, including cluster
configuration. Using a private module is not ideal, but this way we won't
have to copy JGroups config from standalone.xml into zanata-infinispan.xml.

Provided Infinispan version now matches JBoss EAP 6.4.8.

CacheManagerProducer still produces EmbeddedCacheManagers, not
CacheContainers.

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1235)


carlosmunoz:
When you say "mostly", which parts remained of the embedded infinispan remained?

---

Reviewed 16 of 16 files at r1.

---

*[pom.xml, line 2044 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1235#-KMXlQUHXLWwbX3UEK52:-KMXlQUHXLWwbX3UEK53:331704379) ([raw file](https://github.com/zanata/zanata-server/blob/39a2da01b631e2c7184c6b3df0e555930dac028b/pom.xml#L2044)):*
> ```XML
>             <groupId>org.jboss.marshalling</groupId>
>             <artifactId>jboss-marshalling</artifactId>
>             <version>1.4.6.Final</version>
> ```

I think this dependency is declared twice

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1235#-:-KMXmLWXBBJjD_TsjDKN:-1577165546)*

seanf:
Infinispan is no longer embedded at all, but some of the other changes in that PR have not been reverted exactly as they once were, eg CacheManagerProducer which used to be CacheContainerProducer.

---

---

*[pom.xml, line 2044 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1235#-KMXlQUHXLWwbX3UEK52:-KMXsxPdlzR8ShMyjYy1:-1409961014) ([raw file](https://github.com/zanata/zanata-server/blob/39a2da01b631e2c7184c6b3df0e555930dac028b/pom.xml#L2044)):*
> I think this dependency is declared twice
</details>
This one is actually inside the 'wildfly8' profile. I guess the versions used to be different here. I'll get rid of it. This was another one of those things in the original PR which should not have been reverted.

I'd really like to get rid of dependencyManagement from the 'wildfly8' profile. 

IIRC, it's needed because jboss-as-arquillian-container-managed (native management protocol) has one set of dependencies, and wildfly-arquillian-container-managed (native with HTTP upgrade) has another set of newer dependencies (newer than EAP 6). 

With a bit of work, I think we could enable a single management protocol (native) for both appservers, and thus use only jboss-as-arquillian-container-managed, and a single set of dependencies. We might just need to copy the management-native socket-binding from standalone-arquillian.xml to standalone-arquillian-wildfly.xml.

Once we upgrade to EAP 7, we might be able to use wildfly-arquillian-container-managed in both cases.

Or perhaps we can just use https://github.com/arquillian/arquillian-container-chameleon , which I just discovered, which claims to support EAP 6, 7, WF 8, 9, 10 and others.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1235#-:-KMXrvzCMwn3iMsGOxSp:-102957732)*

carlosmunoz:
:lgtm:
-(Awaiting review) +Reviewed 

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1235#-:-KMb4A2qY4oPSoVRMRJS:-1149335731)*

seanf:

seanf:

djansen-redhat:
:white_check_mark: this doesn't seem to break anything

PR #1022, opened 2015-10-23
refactor: replace Joda-Time/PrettyTime with java.time and ICU4J

seanf:

PR #1246, opened 2016-07-21
Remove joda time

seanf:
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1246)


seanf:
This should be re-done, taking advantage of org.hibernate:hibernate-java8.

See also https://github.com/zanata/zanata-server/pull/1022 for another branch of this code.

PR #1252, opened 2016-07-25
fix(openid): fix encoding issue with openid login

aeng:
https://zanata.atlassian.net/browse/ZNTA-1275

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1252)

### DIFF
--- a/zanata-war/src/main/java/org/zanata/security/ZanataOpenId.java
+++ b/zanata-war/src/main/java/org/zanata/security/ZanataOpenId.java
@@ -189,10 +189,19 @@ public class ZanataOpenId implements OpenIdAuthCallback, Serializable {
             // clear any previous messages which might have been generated
             // before the openId authentication took place
             facesMessages.clear();
-            // extract the parameters from the authentication response
-            // (which comes in as a HTTP request from the OpenID providerType)
+            /**
+             * extract the parameters from the authentication response query string
+             * (which comes in as a HTTP request from the OpenID providerType)
+             * instead of from httpReq.getParameterMap() to make sure params
+             * is encoded correctly bypassing default servlets encoding.
+             *
+             * httpReq.getParameterMap() failed check in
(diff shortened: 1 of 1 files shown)


zanata-war/src/main/java/org/zanata/security/ZanataOpenId.java
```diff
@@ -189,10 +189,19 @@ public String verifyResponse(HttpServletRequest httpReq) {
...
+            /**
+             * extract the parameters from the authentication response query string
+             * (which comes in as a HTTP request from the OpenID providerType)
+             * instead of from httpReq.getParameterMap() to make sure params
+             * is encoded correctly bypassing default servlets encoding.
+             *
+             * httpReq.getParameterMap() failed check in
+             * {@link org.openid4java.consumer.ConsumerManager#verifySignature}
+             * due to the unicode encoding in URI is different from signature.
+             * Reported issue: https://zanata.atlassian.net/browse/ZNTA-1275
```
seanf: When I said to put the "reported issue" into the comment, I meant the issue which someone reported against openid4java (or forum link). The link we were looking at yesterday. 

@aeng Could you please find the link we were looking at yesterday, and link to it from the Zanata issue?

seanf:
:+1:

seanf:
@aeng I think this is the issue we were looking at: https://github.com/jbufu/openid4java/issues/141

djansen-redhat:
:white_check_mark: Tested - @aeng @seanf assuming the error was in the "Full Name" field (it wasn't entirely clear), this logs in correctly with Chinese characters in the field.

aeng:
@djansen-redhat The encoding issue is on all the fields requested by Zanata. 
But the easiest to test is the full name field as it supports unicode. Another way to test is the email but I'm not sure if email address supports unicode.

PR #1269, opened 2016-08-15
Validate runtime schemas against Hibernate

seanf:
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1269)


seanf:
Unfortunately, EntityManagerFactory currently gets initialised (and tries to validate tables) before LiquibaseServletListener can create the database tables.

Possible workaround: an Undertow ServletExtension should be able to invoke Liquibase before EntityManagerFactory is initialised.

seanf:
For the status of this pull request, please see https://github.com/zanata/zanata-platform/pull/19

PR #1270, opened 2016-08-17
fix(encoding): force utf8 encoding for /account/ request

aeng:
https://zanata.atlassian.net/browse/ZNTA-1323

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1270)

### DIFF
--- a/zanata-war/src/main/webapp-jboss/WEB-INF/web.xml
+++ b/zanata-war/src/main/webapp-jboss/WEB-INF/web.xml
@@ -156,6 +156,12 @@
     <!--<listener-class>net.bull.javamelody.SessionListener</listener-class>-->
   <!--</listener>-->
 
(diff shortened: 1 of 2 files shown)


aeng:

zanata-war/src/main/webapp-jboss/WEB-INF/web.xml
```diff
@@ -157,6 +157,11 @@
   <!--</listener>-->
 
   <filter-mapping>
+    <filter-name>CharacterEncodingFilter</filter-name>
```
seanf: It would be good to have a comment, explaining why this filter needs to be first: otherwise the character set might accidentally be locked to ISO-8859-1.

seanf:
:+1:

djansen-redhat:
:white_check_mark: looks good to me
:cookie: no rogue commits

PR #1272, opened 2016-08-17
chore(editor): avoid script execution to make npm install more secure

davidmason:
**Migrated to  https://github.com/zanata/zanata-platform/pull/27**

This aims to prevent running npm scripts for dependencies, because that is insecure, but allow our own npm scripts to run because they are needed at the moment to run our build and development tasks.

Our scripts are specified in the `"scripts"` section of package.json.

We can turn off npm script execution completely if we replace all our npm scripts with direct invocations of the relevant binaries (generally in the `node_modules/.bin` directory).

---

This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1272)

### DIFF
--- a/zanata-frontend/pom.xml
+++ b/zanata-frontend/pom.xml
@@ -144,6 +144,7 @@
                 <argument>install</argument>
                 <argument>--cache-min</argument>
                 <argument>${npm.cache.min}</argument>
+                <argument>--ignore-scripts</argument>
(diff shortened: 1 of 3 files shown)


davidmason:
Closing since this has been migrated to the zanata-platform repository: https://github.com/zanata/zanata-platform/pull/27

PR #1278, opened 2016-08-28
fix(statistics): Update profile page statistics query to match with c…

aeng:
Update profile page statistics query to match with contribution query.

https://zanata.atlassian.net/browse/ZNTA-59

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1278)


aeng:
Close this PR. New direction for the statisticws view.

PR #1279, opened 2016-08-29
Increase Docker image memory for Zanata server

djansen-redhat:
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1279)

### DIFF
--- a/docker/rundev.sh
+++ b/docker/rundev.sh
@@ -23,7 +23,9 @@ chcon -Rt svirt_sandbox_file_t "$ZANATA_WAR"
 docker build -t zanata/server-dev docker/
 
 # runs zanata/server-dev:latest docker image
-docker run --rm --name zanata --link zanatadb:db \
+docker run --rm \
+    -e JAVA_OPTS='-Xmx2g' \
+    --name zanata --link zanatadb:db \
     -p 8080:8080 -p 8787:8787 -it \
     -v $ZANATA_WAR:/opt/jboss/wildfly/standalone/deployments/ROOT.war \
     -v $ZANATA_DIR:/opt/jboss/zanata \


djansen-redhat:
Considerations:
- [ ] Do we need to do something similar for a production image? What do we need there?
- [ ] Does it need to be 2GB?
- [ ] Should it be a script argument?

aeng:
>Do we need to do something similar for a production image? What do we need there?
Which production image? 

> Does it need to be 2GB?
Can we set minimum?

>Should it be a script argument?
We can accept max, but I still think we should set the minimum.

seanf:
@djansen-redhat the default for EAP 6 and 7 (in the file `bin/standalone.conf`) is `-Xms1303m -Xmx1303m`. I think that's what we use in production too.

However, that's probably overkill for development servers in most cases. What sort of testing were you doing where 512m was insufficient?

seanf:
>> Does it need to be 2GB?

> Can we set minimum?

Not as such. There's `-Xms`, but it's initial heap size, not minimum.

On EAP, the initial size is 1303m, but on WildFly it's 64m. We could increase that, but to do so is just a speed optimisation, like pre-allocating large ArrayLists. For development, a good initial size (for faster booting) might be, say, 50-100% more than the heap size used while booting Zanata.

But it's probably not worth worrying about.

>> Should it be a script argument?

An environment variable might be easier.

djansen-redhat:
@seanf just doing normal PR testing, and then leaving it for a while (overnight) and coming back to it. Perhaps this is indicative of a severe memory leak, rather than a lack of sufficient memory?

seanf:
@djansen-redhat Yes, it could be a memory leak. If it happens again, please try to get the OOM exception trace (just in case), plus it would be helpful to grab a memory histogram in Javamelody. Maybe even a heap dump.

seanf:
At this stage, I'm not convinced increasing the dev heap size is justified, since it might merely mask a memory leak.

djansen-redhat:
@seanf I agree. Abandoning this change.

PR #1281, opened 2016-08-30
Enable heap dump on OOM

seanf:
See also https://github.com/zanata/zanata-server/pull/1279

---
This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1281)


seanf:
@djansen-redhat Do you want to do anything here?

djansen-redhat:
@seanf not particularly.

PR #1284, opened 2016-09-06
fix(ZNTA-1363): handle stats being disabled

seanf:
https://zanata.atlassian.net/browse/ZNTA-1363

See also https://zanata.atlassian.net/browse/ZNTA-1364, https://issues.jboss.org/browse/WFLY-3377 and https://issues.jboss.org/browse/ISPN-5421

This change is [Reviewable](https://reviewable.io/reviews/zanata/zanata-server/1284)

### DIFF
--- a/docker/conf/standalone.xml
+++ b/docker/conf/standalone.xml
@@ -615,9 +615,9 @@
 
     <socket-binding-group name="standard-sockets" default-interface="public"
       port-offset="${jboss.socket.binding.port-offset:0}">
-        <socket-binding name="management-http" interface="management"
+        <socket-binding name="management-http" interface="public"
           port="${jboss.management.http.port:9990}" />
-        <socket-binding name="management-https" interface="management"
+        <socket-binding name="management-https" interface="public"
           port="${jboss.management.https.port:9993}" />
         <socket-binding name="ajp"
           port="${jboss.ajp.port:8009}" />
--- a/docker/conf/standalone.xml
+++ b/docker/conf/standalone.xml
@@ -377,8 +377,8 @@
             -->
             <cache-container name="zanata"
               module="org.jboss.as.clustering.web.infinispan" default-cache="default"
-              jndi-name="java:jboss/infinispan/container/zanata">
-                <local-cache name="default">
+              jndi-name="java:jboss/infinispan/container/zanata" statistics-enabled="true">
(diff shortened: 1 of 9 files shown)


aeng:
@seanf I will need to sit down with you to understand more on this PR.

seanf:
@aeng sure. But basically, if stats aren't enabled for a given cache (for whatever reason), AdvancedCache.getStats() throws an exception, so I substitute a dummy Stats object. It's possible AdvancedCache.getStats() has changed, but I think it's simply become more difficult to enable stats on caches (see https://issues.jboss.org/browse/WFLY-3377 and https://issues.jboss.org/browse/ISPN-5421)

I'm sure there's a cleaner way of dealing with this, but if we're going to spend more than minimal time on it, I'm thinking we should use generic JMX access instead of depending on Infinispan classes (and the private jboss module for infinispan).

https://docs.jboss.org/infinispan/8.1/apidocs/jmxComponents.html

aeng:
Reviewed 8 of 9 files at r1.
Review status: 8 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[zanata-war/src/main/java/org/zanata/action/CacheAction.java, line 77 \[r1\]](https://reviewable.io:443/reviews/zanata/zanata-server/1284#-KR11vSvY8i1phJ-zLPh:-KR11vSvY8i1phJ-zLPi:b-6gpwa) ([raw file](https://github.com/zanata/zanata-server/blob/60f99401469d50fac5f35c48f21cc18274d5fb96/zanata-war/src/main/java/org/zanata/action/CacheAction.java#L77)):*
> ```Java
>             } else {
>                 log.debug("Statistics are not enabled for cache '{}'", cacheName);
>                 return new UnavailableStats(-1);
> ```

Perhaps a comment of what -1 and -2 means?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1284)*

aeng:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/zanata/zanata-server/1284#-:-KRaZ_apG3P9DLjOMWcW:bnfp4nl)*

djansen-redhat:
:white_check_mark: I think this is ok

